PR #110, opened 2015-03-09
Try SHGetKnownFolderPath. Needs testing.

jdolan:
@jar3817 Can you take a look at this and massage it until it works? Once we get it working, I'll port it to GtkRadiant. NetRadiant already does this.

### DIFF
--- a/src/sys.c
+++ b/src/sys.c
@@ -27,6 +27,7 @@
 
 #if defined(_WIN32)
 #include <shlobj.h>
+
 #define RTLD_NOW 0
 #define dlopen(file_name, mode) LoadLibrary(file_name)
 #define dlerror() "Windows.. go figure."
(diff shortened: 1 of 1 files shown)


jar3817:
Works! Tested on 7 Pro x64 (detects as ~/Documents/My Games/Quetoo) and on XP Pro SP3 i686 (detects as ~/My Documents/My Games/Quetoo).

You should probably change the Vista+ path to ~/Saved Games/Quetoo (wasn't that the path that something used that we had to mimic?)

Either way seems to work.

jdolan:
Oh. That's weird. The GUID I provide should be the same as FOLDERID_SavedGames. Hm. Will have to look a bit more.

jdolan:
I think my call to SHGetKnownFolderPath must be failing, if you're getting anything with "My Games." I'll add some debug statements so we can see what's going on.

PR #115, opened 2015-11-09
Jdolan/quetoo dedicated

jdolan:
Please don't merge this PR yet. I'd like to use it for code review purposes as we stabilize the new console / new build system. There are still major breakages (tho no obvious crashes!). The chat and server consoles wonky, and the new loading screens are not yet implemented.


configure.ac
```diff
@@ -312,98 +239,6 @@ AC_CHECK_HEADER(execinfo.h,
...
-	fi
-
-	CPPFLAGS="$save_CPPFLAGS"
-	LDFLAGS="$save_LDFLAGS"
-fi
-
-AC_SUBST(CURSES_CFLAGS)
-AC_SUBST(CURSES_LDFLAGS)
-AC_SUBST(CURSES_LIBS)
-
```
jdolan: The goal here was to streamline our build configuration, such that a build environment will be required to have all of our "core" dependencies (libCURL, libjpeg, libSDL, libxml). Because we are now able to easily produce a dedicated server binary alongside the client binary, we can provide and distribute that ourselves, rather than asking server admins to compile their own executable. So making these dependencies a requirement is not imposing on anyone but us.

configure.ac
```diff
@@ -453,9 +289,31 @@ AC_ARG_WITH(games,
...
+	)
+)
+
+if test "x${with_tools}" = x; then
+	TOOLS="${with_tools}"
+fi
+
+AC_SUBST(TOOLS)
+AC_MSG_RESULT($TOOLS)
+
```
jdolan: The map compiler is now built by default. Pass `--tools=none` to disable it.

Issue #98, opened 2015-02-22
Player does not clip correctly when dead


jdolan:
Client entities are immediately made `SOLID_NOT` upon dying. They also immediately spawn a corpse that is `SOLID_DEAD`. This is almost correct, but not quite. The effect is that, until they respawn, dead clients do not correctly clip to the world and to projectiles.

What should happen instead, is that the client entity should run the death animations and become `SOLID_DEAD` themselves. Only when the player initiates the respawn should we drop a corpse in their place.

jdolan:
See `g_client.c`, `G_ClientDie` and friends.

jdolan:
Done and done.

PR #122, opened 2015-12-31
#98: First pass at making clients clip correctly when dead. This re-i…

jdolan:
…mplements the corpse logic such that the client _is_ the corpse, and even _is_ a giblet, until they respawn. When they respawn, only then do we drop a corpse or giblet back in their stead. A fun side effect of this change is that you can gib your buddy, and then use him as a soccer ball ad nausea until he decides to respawn.

### DIFF
--- a/src/cgame/default/cg_entity.c
+++ b/src/cgame/default/cg_entity.c
@@ -26,6 +26,9 @@
  */
 _Bool Cg_IsSelf(const cl_entity_t *ent) {
 
+	if (ent->current.number == cgi.client->client_num + 1)
+		return true;
+
 	if (ent->current.model1 == MODEL_CLIENT || (ent->current.effects & EF_BEAM)) {
 
 		if (ent->current.client == cgi.client->client_num)
@@ -56,12 +59,16 @@ static void Cg_AddClientEntity(cl_entity_t *ent, r_entity_t *e) {
 	e->effects |= EF_CLIENT;
 
 	// don't draw ourselves unless third person is set
-	if (Cg_IsSelf(ent) && !(e->effects & EF_CORPSE) && !cg_third_person->value) {
-		e->effects |= EF_NO_DRAW;
+	if (Cg_IsSelf(ent) && !cg_third_person->value) {
 
-		// keep our shadow underneath us using the predicted origin
-		e->origin[0] = cgi.view->origin[0];
-		e->origin[1] = cgi.view->origin[1];
+		// corpses pass the "self" test (of course)
+		if (!(e->effects & EF_CORPSE)) {
+			e->effects |= EF_NO_DRAW;
+
+			// keep our shadow underneath us using the predicted origin
+			e->origin[0] = cgi.view->origin[0];
+			e->origin[1] = cgi.view->origin[1];
+		}
 	}
 
 	r_entity_t head, upper, lower;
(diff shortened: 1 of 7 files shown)


src/cgame/default/cg_entity.c
```diff
@@ -26,6 +26,9 @@
  */
 _Bool Cg_IsSelf(const cl_entity_t *ent) {
 
+	if (ent->current.number == cgi.client->client_num + 1)
```
jdolan: This is a more direct way to check this, anyway, and it works no matter what model (e.g. giblet) our client has.

src/game/default/bg_pmove.c
```diff
@@ -637,43 +640,44 @@ static void Pm_CheckDuck(void) {
...
-		if (pml.view_offset[2] < target) // go up
-			pml.view_offset[2] += pml.time * PM_SPEED_DUCK_STAND;
+			if (pml.view_offset[2] < target) // go up
+				pml.view_offset[2] += pml.time * PM_SPEED_DUCK_STAND;
 
-		if (pml.view_offset[2] > target)
-			pml.view_offset[2] = target;
+			if (pml.view_offset[2] > target)
+				pml.view_offset[2] = target;
+		}
```
jdolan: Simplified the ducking logic such that PM_DEAD is it's own separate case, not mixed in with PM_NORMAL.

src/game/default/bg_pmove.c
```diff
@@ -1176,8 +1180,14 @@ static void Pm_SpectatorMove() {
...
-	VectorScale(PM_MINS, PM_SCALE, pm->mins);
-	VectorScale(PM_MAXS, PM_SCALE, pm->maxs);
+	// set the default bounding box
+	if (pm->s.flags & PMF_GIBLET) {
+		VectorCopy(PM_GIBLET_MINS, pm->mins);
+		VectorCopy(PM_GIBLET_MAXS, pm->maxs);
+	} else {
+		VectorScale(PM_MINS, PM_SCALE, pm->mins);
+		VectorScale(PM_MAXS, PM_SCALE, pm->maxs);
+	}
```
jdolan: PMF_GIBLET constitutes its own bounding box. I suppose PM_DEAD could be incorporated here, actually.

src/game/default/g_client.c
```diff
@@ -346,7 +324,7 @@ static void G_ClientCorpse_Die(g_entity_t *self, g_entity_t *attacker __attribut
 
 		ent->locals.clip_mask = MASK_CLIP_CORPSE;
 		ent->locals.dead = true;
-		ent->locals.mass = (i % NUM_GIB_MODELS) * 20.0;
+		ent->locals.mass = ((i % NUM_GIB_MODELS) + 1) * 20.0;
```
jdolan: Old bug: the smallest giblet had `0` mass.

src/game/default/g_combat.c
```diff
@@ -317,7 +326,7 @@ void G_Damage(g_entity_t *target, g_entity_t *inflictor, g_entity_t *attacker, c
 		G_ClientDamageKick(target, dir, kick);
 
 		if (attacker->client && attacker->client != client) {
-			attacker->client->locals.damage_inflicted += damage_health;
+			attacker->client->locals.damage_inflicted += damage_health + damage_armor;
 		}
```
jdolan: Fix another bug where a hit sound is not heard if the target had armor.

src/game/default/g_combat.c
```diff
@@ -299,6 +304,10 @@ void G_Damage(g_entity_t *target, g_entity_t *inflictor, g_entity_t *attacker, c
...
 	}
 
+	// if the target was already dead, we're done
+	if (was_dead)
+		return;
+
```
jdolan: Don't give credit for hurting dead stuff.

PR #129, opened 2016-01-05
Claire/123 armor

jdolan:
Opening a PR to discuss changes

### DIFF
--- a/src/game/default/g_ballistics.c
+++ b/src/game/default/g_ballistics.c
@@ -159,7 +159,7 @@ static void G_BlasterProjectile_Touch(g_entity_t *self, g_entity_t *other,
 	if (!G_IsSky(surf)) {
 
 		G_Damage(other, self, self->owner, self->locals.velocity, self->s.origin, plane->normal,
-				self->locals.damage, self->locals.knockback, 0, MOD_BLASTER);
+				self->locals.damage, self->locals.knockback, DMG_ENERGY, MOD_BLASTER);
 
 		if (G_IsStructural(other, surf)) {
 			vec3_t origin;
--- a/src/game/default/g_combat.c
+++ b/src/game/default/g_combat.c
@@ -137,7 +137,7 @@ static int16_t G_CheckArmor(g_entity_t *ent, const vec3_t pos, const vec3_t norm
 	const int16_t quantity = ent->client->locals.inventory[ITEM_INDEX(armor)];
 
 	vec_t protection = 0.0;
-	vec_t efficiency = 1.0;
+	vec_t burnrate = 1.0;
 
 	// energy weaps cut through armor more
 	const int16_t type = (dflags & DMG_ENERGY) ? armor->tag-1 : armor->tag;
@@ -145,20 +145,20 @@ static int16_t G_CheckArmor(g_entity_t *ent, const vec3_t pos, const vec3_t norm
 	switch (type) {
 		case ARMOR_BODY:
 			protection = 0.75;
-			efficiency = 1.35;
+			burnrate = 1.35;
 			break;
 		case ARMOR_COMBAT:
 			protection = 0.66;
-			efficiency = 1.25;
+			burnrate = 1.25;
 			break;
 		case ARMOR_JACKET:
 			protection = 0.33;
-			efficiency = 1.15;
+			burnrate = 1.15;
 			break;
 	}
 
 	const int16_t saved = Clamp(protection * damage, 0, quantity);
-	ent->client->locals.inventory[ITEM_INDEX(armor)] -= (saved / efficiency);
+	ent->client->locals.inventory[ITEM_INDEX(armor)] -= (saved / burnrate);
 
 	G_SpawnDamage(TE_BLOOD, pos, normal, saved);
 
--- a/src/game/default/g_combat.c
+++ b/src/game/default/g_combat.c
@@ -130,35 +130,21 @@ static int16_t G_CheckArmor(g_entity_t *ent, const vec3_t pos, const vec3_t norm
 		return 0;
 
 	const g_item_t *armor = G_ClientArmor(ent);
+	const g_armor_info_t *armor_info = G_GetArmorInfo(armor); 
 
 	if (!armor)
 		return 0;
 
 	const int16_t quantity = ent->client->locals.inventory[ITEM_INDEX(armor)];
+	int16_t saved;
 
-	vec_t protection = 0.0;
-	vec_t burnrate = 1.0;
+	
+	if (dflags & DMG_ENERGY)
+		saved = Clamp(damage * armor_info->energy_protection, 0, quantity);
+	else
+		saved = Clamp(damage * armor_info->normal_protection, 0, quantity);	
 
-	// energy weaps cut through armor more
-	const int16_t type = (dflags & DMG_ENERGY) ? armor->tag-1 : armor->tag;
-
-	switch (type) {
-		case ARMOR_BODY:
-			protection = 0.75;
-			burnrate = 1.35;
-			break;
-		case ARMOR_COMBAT:
-			protection = 0.66;
-			burnrate = 1.25;
-			break;
-		case ARMOR_JACKET:
-			protection = 0.33;
-			burnrate = 1.15;
-			break;
-	}
-
-	const int16_t saved = Clamp(protection * damage, 0, quantity);
-	ent->client->locals.inventory[ITEM_INDEX(armor)] -= (saved / burnrate);
+	ent->client->locals.inventory[ITEM_INDEX(armor)] -= saved;
 
 	G_SpawnDamage(TE_BLOOD, pos, normal, saved);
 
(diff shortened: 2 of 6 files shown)


src/game/default/g_combat.c
```diff
@@ -137,21 +137,28 @@ static int16_t G_CheckArmor(g_entity_t *ent, const vec3_t pos, const vec3_t norm
...
+			waste = 0.65;
 			break;
 		case ARMOR_COMBAT:
 			protection = 0.66;
+			waste = 0.75;
 			break;
 		case ARMOR_JACKET:
-			protection = 0.5;
+			protection = 0.33;
+			waste = 0.85;
```
jdolan: Nice. I think this makes sense. Might read nicer `s/waste/efficiency/g`

jdolan: Shouldn't the efficiencies be > 1.0, tho? If I'm reading this correctly, for every 1 damage that body armor saves you, it only uses .65 units of armor. I would think it makes more sense for 1 unit in damage saved should cost 1.33 (for example -- just throwing that out there) units of armor.

jar3817:
Funny enough, that's exactly how I had it the first time. But I got hung upon having efficiency be more than 100%. Also I thought it might be better to multiply rather than divide to get the result, but with the frequency in which this is calculated probably makes no different.

At one point the variable was named "heatloss" :)

jdolan:
Hah! Heatloss, burnrate, how about expense? Those would all work. Agreed that an "efficiency" > 1.0 reads kinda odd.

jar3817:
I added the energy flag to blaster damage, seems to make sense (that way in q2 also). Perhaps blaster might even be more fun if it were DMG_NO_ARMOR.

jdolan:
Great catch! I'm on the fence about making the blaster cut through armor. While it would certainly make the blaster more useful, it may seem very strange to have the Hyper and BFG not.. Hm. Let's throw a comment in the code above the G_Damage call indicating that someday we might want to try that.

src/game/default/g_combat.c
```diff
@@ -137,21 +137,28 @@ static int16_t G_CheckArmor(g_entity_t *ent, const vec3_t pos, const vec3_t norm
...
 		case ARMOR_JACKET:
-			protection = 0.5;
+			protection = 0.33;
+			burnrate = 1.15;
 			break;
 	}
 
 	const int16_t saved = Clamp(protection * damage, 0, quantity);
-	ent->client->locals.inventory[ITEM_INDEX(armor)] -= saved;
+	ent->client->locals.inventory[ITEM_INDEX(armor)] -= (saved / burnrate);
```
jdolan: `saved * burnrate` would yield > 1 unit of armor for every point of damage saved, which is what we want, right? I guess I should look at Q2 / Q3 to see how they expend armor..

jdolan: So Q2 just removes `saved` from the appropriate inventory index. But I think, more interestingly, Q2 has an armor info struct that specifies the pickup quantity as well as the max quantity for a given armor. We don't have that distinction anymore (my bad). So in Q2, when you grab the body armor, you pickup 100 units, but you can carry 200 units total. The other armors similarly allow you to carry more than one item's worth. These armor info structs also contain the protection rates for normal as well as energy based damage.

```
typedef struct
{
	int		base_count;
	int		max_count;
	float	normal_protection;
	float	energy_protection;
	int		armor;
} gitem_armor_t;

gitem_armor_t jacketarmor_info	= { 25,  50, .30, .00, ARMOR_JACKET};
gitem_armor_t combatarmor_info	= { 50, 100, .60, .30, ARMOR_COMBAT};
gitem_armor_t bodyarmor_info	= {100, 200, .80, .60, ARMOR_BODY};
```

Something similar might work well for us: an array of `g_armor_info_t`, where the struct looks like:

```
typedef struct {
    const char *item;  // so that we can map these to the items themselves
    vec_t protection;
    vec_t expense;
    vec_t energy_protection;
    vec_t energy_expense;
    int16_t max;
} g_armor_info_t;
```

Places this would impact:

G_CheckArmor
G_PickupArmor
G_Give_f

Thoughts?

src/game/default/g_ballistics.c
```diff
@@ -159,7 +159,7 @@ static void G_BlasterProjectile_Touch(g_entity_t *self, g_entity_t *other,
 	if (!G_IsSky(surf)) {
 
 		G_Damage(other, self, self->owner, self->locals.velocity, self->s.origin, plane->normal,
-				self->locals.damage, self->locals.knockback, 0, MOD_BLASTER);
+				self->locals.damage, self->locals.knockback, DMG_ENERGY, MOD_BLASTER);
```
jdolan: Nice catch.

src/game/default/g_combat.c
```diff
@@ -130,27 +130,20 @@ static int16_t G_CheckArmor(g_entity_t *ent, const vec3_t pos, const vec3_t norm
...
-			break;
-		case ARMOR_JACKET:
-			protection = 0.5;
-			break;
-	}
+	
+	if (dflags & DMG_ENERGY)
+		saved = Clamp(damage * armor_info->energy_protection, 0, quantity);
+	else
+		saved = Clamp(damage * armor_info->normal_protection, 0, quantity);
```
jdolan: Prrty.

src/game/default/g_item.c
```diff
@@ -25,6 +25,10 @@
 const vec3_t ITEM_MINS = { -16.0, -16.0, -16.0 };
 const vec3_t ITEM_MAXS = { 16.0, 16.0, 16.0 };
 
+const g_armor_info_t bodyarmor_info = { 100, 200, 0.8, 0.6, ARMOR_BODY };
```
jdolan: I think I'd like to see these initialized into an array, rather than having 3 globals. Then G_GetArmorInfo could just loop over the array, looking for the tag.

src/game/default/g_item.c
```diff
@@ -332,29 +336,76 @@ static _Bool G_PickupHealth(g_entity_t *ent, g_entity_t *other) {
...
+			taken = true;	
+		} else {
+			salvage = new_info->normal_protection / current_info->normal_protection;
+                        salvage_count = salvage * new_armor->quantity;
+                        new_count = salvage_count + other->client->locals.inventory[ITEM_INDEX(current_armor)];
+			new_count = Clamp(new_count, 0, current_info->max_count);
+			if (other->client->locals.inventory[ITEM_INDEX(current_armor)] >= new_count)
+				return false;
+                        other->client->locals.inventory[ITEM_INDEX(current_armor)] = new_count;
+                        taken = true;
```
jdolan: Whoa, this is cool. Could you add a comment as to what's happening here, and / or where it came from? I'm all for cool rules that add to the game, tho.

src/game/default/g_item.c
```diff
@@ -893,7 +944,7 @@ const g_item_t g_items[] = {
 		EF_ROTATE | EF_BOB | EF_PULSE,
 		"pics/i_bodyarmor",
 		"Body Armor",
-		200,
+		100,
```
jdolan: :+1:

src/game/default/g_item.h
```diff
@@ -45,6 +45,7 @@ const g_item_t *G_FindItem(const char *name);
 const g_item_t *G_FindItemByClassName(const char *class_name);
 const g_item_t *G_ItemByIndex(uint16_t index);
 const g_item_t *G_ClientArmor(const g_entity_t *ent);
+const g_armor_info_t *G_GetArmorInfo(const g_item_t *armor);
```
jdolan: Hate to be a style nazi, but we rarely use get/set nomenclature for functions. Mind naming this just `G_ArmorInfo`?

src/game/default/g_types.h
```diff
@@ -379,6 +379,17 @@ typedef struct g_item_s {
...
+ */
+typedef struct {
+    int     base_count;
+    int     max_count;
+    float   normal_protection;
+    float   energy_protection;
+    int     armor;
+} g_armor_info_t;
+
+/*
```
jdolan: `extern g_armor_info_t g_armor_info[];`

src/game/default/g_item.c
```diff
@@ -25,6 +25,17 @@
...
+	{ 100, 200, 0.8, 0.6, ARMOR_BODY },
+	{ 50, 100, 0.6, 0.3, ARMOR_COMBAT },
+	{ 25, 50, 0.3, 0.0, ARMOR_JACKET }
+};
+
+// move this somewhere
+//armor_info[0] = { 100, 200, 0.8, 0.6, ARMOR_BODY };
+//armor_info[1] = { 50, 100, 0.6, 0.3, ARMOR_COMBAT };
+//armor_info[2] = { 25, 50, 0.3, 0.0, ARMOR_JACKET };
+
```
jdolan: Cool. You know what? These could actually be moved right into G_ArmorInfo as `static const`. It seems like they're only ever accessed via that function. I'll take a crack at this.

jdolan:
Looks great, man. Thanks for knocking this out!

PR #137, opened 2016-06-06
Claire/136 duel

jar3817:
Decent start. Lots of little things...


src/game/default/g_client.c
```diff
@@ -570,6 +570,7 @@ static void G_InitClientInventory(g_entity_t *ent) {
 	// instagib gets railgun and slugs, both in normal mode and warmup
 	if (g_level.gameplay == GAME_INSTAGIB) {
 		G_Give(ent, "Railgun", 1000);
+		G_Give(ent, "Grenades", 50);
 		item = G_FindItem("Railgun");
```
jdolan: NICE.

src/game/default/g_client_stats.c
```diff
@@ -54,6 +54,11 @@ void G_ClientToIntermission(g_entity_t *ent) {
 
 	ent->client->locals.ammo_index = 0;
 	ent->client->locals.pickup_msg_time = 0;
+	
+	// take a screenshot if we're supposed to
+	if (g_force_screenshot->integer == 1){
+		G_ClientStuff(ent, "r_screenshot\n");
+	}
```
jdolan: I think I'd rather see this implemented as a client-side latch with a cvar to enable / disable. The game could send a new command (`SV_CMD_INTERMISSION`), which the cgame could parse and, if the appropriate cvar is enabled (`cg_intermission_screenshot`?), then take the shot. Thoughts?

src/game/default/g_cmd.c
```diff
@@ -663,7 +665,21 @@ static _Bool G_VoteHelp(g_entity_t *ent) {
...
+			"Match mode is required for DUEL gameplay, setting cannot be changed\n");
+		return true;
+	}
+	
+	// teams are required for duel mode
+	if (g_level.gameplay == GAME_DUEL && !g_strcmp0(gi.Argv(1), "g_teams")) {
+		gi.ClientPrint(ent, PRINT_HIGH, 
+			"Teams are required for DUEL gameplay, setting cannot be changed\n");
 		return true;
 	}
```
jdolan: Nice.

src/game/default/g_cmd.c
```diff
@@ -697,11 +713,20 @@ static void G_Vote_f(g_entity_t *ent) {
 			gi.ClientPrint(ent, PRINT_HIGH, "You've already voted\n");
 			return;
 		}
-		if (g_strcmp0(vote, "yes") == 0)
+		if (g_strcmp0(vote, "yes") == 0) {
+			if (ent->client->locals.persistent.admin) {
+				g_level.votes[VOTE_YES] = 100; // admin vote wins immediately
```
jdolan: `g_level.votes[VOTE_YES] = sv_max_clients->integer;` ?

src/game/default/g_cmd.c
```diff
@@ -729,6 +754,10 @@ static void G_Vote_f(g_entity_t *ent) {
 
 	ent->client->locals.persistent.vote = VOTE_YES; // client has implicitly voted
 	g_level.votes[VOTE_YES] = 1;
+	
+	if (ent->client->locals.persistent.admin) {
+		g_level.votes[VOTE_YES] = 100;	// admin votes always succeed...
```
jdolan: g_level.votes[VOTE_YES] = sv_max_clients->integer; ?
 Add a line note

src/game/default/g_cmd.c
```diff
@@ -757,11 +786,16 @@ _Bool G_AddClientToTeam(g_entity_t *ent, const char *team_name) {
 	if (ent->client->locals.persistent.team == team)
 		return false;
 
+	if (g_level.gameplay == GAME_DUEL && G_TeamSize(team) > 0) {
+		gi.ClientPrint(ent, PRINT_HIGH, "Only 1 player per team allowed in Duel mode\n");
+		return false;
+	}
+
```
jdolan: I really appreciate the effort you've gone through to sanitize all of the requisite gameplay vars to enable duel mode. I'm slightly concerned that users might find it too tedious / difficult to setup? For this scenario, it might be better to force all existing clients to no team affiliation when gameplay is changed. Otherwise, enabling duel on a populated server would require full cooperation from everyone connected. What do you think?

src/game/default/g_cmd.c
```diff
@@ -925,6 +959,42 @@ static void G_Teamskin_f(g_entity_t *ent) {
...
+	gi.BroadcastPrint(PRINT_HIGH, "%s is having second thoughts...%s\n", 
+		ent->client->locals.persistent.net_name,
+		(g_level.start_match) ? "countdown aborted" : ""
+	);
+		
+	g_level.start_match = false;
+	g_level.match_time = 0;
+	g_level.match_status = MSTAT_WARMUP;
+}
+
```
jdolan: Nice!

src/game/default/g_cmd.c
```diff
@@ -942,12 +1012,13 @@ static void G_Ready_f(g_entity_t *ent) {
 	}
 
 	if (ent->client->locals.persistent.ready) {
-		gi.ClientPrint(ent, PRINT_HIGH, "You're already ready\n");
+		G_Unready_f(ent);
```
jdolan: Is typing `"ready"` multiple times how Q2 duel mods handle this, or should we map `"unready"` explicitly as well? The way this is now, if I forget that I'm ready, and type "ready", I've just un-readied myself. This is more of a question than a suggestion.

src/game/default/g_cmd.c
```diff
@@ -982,39 +1053,11 @@ static void G_Ready_f(g_entity_t *ent) {
 		return;
 	}
 
-	gi.BroadcastPrint(PRINT_HIGH, "Match starting in 10 seconds...\n");
-	g_level.match_time = g_level.time + 10000;
+	gi.BroadcastPrint(PRINT_HIGH, "Match starting in %d seconds...\n", g_warmup_time->integer);
+	g_level.match_time = g_level.time + (g_warmup_time->integer * 1000);
```
jdolan: Do we clamp `g_warmup_time` somewhere? If not, we definitely should! :)

src/game/default/g_cmd.c
```diff
@@ -1068,6 +1111,73 @@ void G_Score_f(g_entity_t *ent) {
...
+		if (g_strcmp0(gi.Argv(1), g_admin_password->string) == 0){
+			ent->client->locals.persistent.admin = true;
+			gi.BroadcastPrint(PRINT_HIGH, "%s became an admin\n", ent->client->locals.persistent.net_name);
+		} else {
+			gi.ClientPrint(ent, PRINT_HIGH, "Invalid admin password\n");
+		}
+		return;
+	} 
+}
+/*
```
jdolan: Awesome, dude. Love this.

src/game/default/g_entity.c
```diff
@@ -411,6 +411,11 @@ static void G_InitMedia(void) {
 		g_media.sounds.gib_hits[i] = gi.SoundIndex(va("gibs/gib_%i/hit", i + 1));
 	}
 
+	for (i = 1; i < 11; i++) {
```
jdolan: `i < lengthof(g_media.sounds.countdown)`

src/game/default/g_cmd.c
jar3817: It just looks tedious, in reality you just either vote or change gameplay
and it enables teams and matches automatically. We'll have to try it out
when I get back from vacation.

I do need to add some sounds to the quetoo pak though. I have them, just
need to upload.
On Jun 30, 2016 7:58 AM, "Jay Dolan" <notifications@github.com> wrote:

In src/game/default/g_cmd.c
<https://github.com/jdolan/quetoo/pull/137#discussion_r69119173>:

> @@ -757,11 +786,16 @@ _Bool G_AddClientToTeam(g_entity_t *ent, const char *team_name) {
>  	if (ent->client->locals.persistent.team == team)
>  		return false;
>
> +	if (g_level.gameplay == GAME_DUEL && G_TeamSize(team) > 0) {
> +		gi.ClientPrint(ent, PRINT_HIGH, "Only 1 player per team allowed in Duel mode\n");
> +		return false;
> +	}
> +	

I really appreciate the effort you've gone through to sanitize all of the
requisite gameplay vars to enable duel mode. I'm slightly concerned that
users might find it too tedious / difficult to setup? For this scenario, it
might be better to force all existing clients to no team affiliation when
gameplay is changed. Otherwise, enabling duel on a populated server would
require full cooperation from everyone connected. What do you think?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jdolan/quetoo/pull/137/files/de9df12fe5a5cb1073369b0ce0c3d1dffc772747#r69119173>,
or mute the thread
<https://github.com/notifications/unsubscribe/AEB1YccM62FDJGq82AZuYsQ1PXS2O1ZSks5qQ698gaJpZM4IvCRR>
.

src/game/default/g_main.c
```diff
@@ -594,39 +606,7 @@ static void G_CheckRules(void) {
...
-			gi.BroadcastPrint(PRINT_HIGH, "Time limit hit\n");
-			G_EndLevel();
-			return;
-		}
-		time = g_level.time_limit - g_level.time; // count down
-	}
-
-	if (g_level.frame_num % gi.frame_rate == 0) // send time updates once per second
-		gi.ConfigString(CS_TIME, g_level.warmup ? "Warmup" : G_FormatTime(time));
+	G_RunTimers();
```
jdolan: 👍

src/game/default/g_main.c
```diff
@@ -796,21 +799,23 @@ static void G_Frame(void) {
 			return;
 		}
 	}
+		
+	if (!TIMEOUT) {
```
jdolan: This _feels_ dangerous. But I'm assuming you've tested it and nothing goes wonky? Do we effectively suspend `g_level.time` during `TIMEOUT`? Otherwise, plats and things will jump to some new position when `TIMEOUT` expires. I could see lots of things breaking if `g_level.time` is allowed to run during `TIMEOUT`.

src/game/default/g_main.c
```diff
@@ -796,21 +799,23 @@ static void G_Frame(void) {
...
-
-		g_level.current_entity = ent;
+			g_level.current_entity = ent;
 
-		if (ent->client) {
-			G_ClientBeginFrame(ent);
-		} else {
-			G_RunEntity(ent);
+			if (ent->client) {
+				G_ClientBeginFrame(ent);
```
jdolan: Likewise, not acknowledging / running client commands during `TIMEOUT` is also probably dangerous (?)

If you see weirdness in testing this, we might want to explore freezing clients and `func_*` in place during `TIMEOUT`, rather than skipping them altogether.

src/game/default/g_main.c
```diff
@@ -938,6 +948,119 @@ void G_Shutdown(void) {
...
+	g_level.timeout_caller = ent;
+	g_level.timeout_time = g_level.time + (g_timeout_time->integer * 1000);
+	g_level.timeout_frame = g_level.frame_num;
+	
+	// lock players
+	for (int32_t i = 1; i < sv_max_clients->integer; i++) {
+		if (g_game.entities[i].client->locals.persistent.team) {
+			g_game.entities[i].client->ps.pm_state.type = PM_FREEZE;
+		}
+	}
```
jdolan: Ah, good! You're already freezing clients!

src/game/default/g_main.c
```diff
@@ -938,6 +948,119 @@ void G_Shutdown(void) {
...
+
+void G_CallTimeIn(void) {
+	
+	g_level.frame_num = g_level.timeout_frame; // where we were before timeout
+	
+	// unlock players
+	for (int32_t i = 1; i < sv_max_clients->integer; i++) {
+		if (g_game.entities[i].client->locals.persistent.team) {
+			g_game.entities[i].client->ps.pm_state.type = PM_NORMAL;
+		}
```
jdolan: Also, what of new clients who connect during `TIMEOUT`?

src/game/default/g_main.c
```diff
@@ -938,6 +948,119 @@ void G_Shutdown(void) {
...
+				
+				G_TeamCenterPrint(&g_team_good, "%s\n", (!j) ? "Fight!" : va("%d", j));
+				G_TeamCenterPrint(&g_team_evil, "%s\n", (!j) ? "Fight!" : va("%d", j));	
+			}
+		} else { 
+			gi.ConfigString(CS_TIME, G_FormatTime(time));
+		}
+	}
+}
+
```
jdolan: This is great, thanks for moving this into its own function.

src/game/default/g_main.h
```diff
@@ -26,6 +26,11 @@
 
 #ifdef __GAME_LOCAL_H__
 
+#define PLAYING (g_level.match_status & MSTAT_PLAYING)
+#define TIMEOUT (g_level.match_status & MSTAT_TIMEOUT)
+#define COUNTDOWN (g_level.match_status & MSTAT_COUNTDOWN)
+#define WARMUP (g_level.match_status & MSTAT_WARMUP)
+
```
jdolan: Cool. I *think* I'd prefer to see this prefixed with `G_`, but I'm okay with shorthand like this.

src/game/default/g_main.h
```diff
@@ -34,13 +39,16 @@ extern g_export_t ge;
...
 
+extern cvar_t *g_admin_password;
 extern cvar_t *g_ammo_respawn_time;
 extern cvar_t *g_auto_join;
 extern cvar_t *g_capture_limit;
 extern cvar_t *g_cheats;
 extern cvar_t *g_ctf;
 extern cvar_t *g_frag_limit;
 extern cvar_t *g_friendly_fire;
+extern cvar_t *g_force_demo;
```
jdolan: Same comment / question for `g_force_demo` as `g_force_screenshot` -- could / should be moved to client-side preference.

src/game/default/g_types.h
```diff
@@ -515,6 +523,8 @@ typedef struct {
 
 		uint16_t weapon_no_ammo;
 		uint16_t weapon_switch;
+		
+		uint16_t countdown[11];
```
jdolan: `#define NUM_COUNTDOWN_SOUNDS 11`, and then either use `NUM_COUNTDOWN_SOUNDS` or `lengthof(g_media.sounds.countdown)` when iterating, etc.

src/game/default/g_util.c
```diff
@@ -547,6 +551,24 @@ g_team_t *G_OtherTeam(g_team_t *t) {
...
+ */
+size_t G_TeamSize(g_team_t *team) {
+	g_client_t *cl;
+	size_t count = 0;
+	
+	for (int32_t i = 0; i < sv_max_clients->integer; i++){
+		if (!g_game.entities[i + 1].in_use)
+			continue;
+			
+		cl = g_game.entities[i + 1].client;
```
jdolan: Nitpick, but you could move `g_client_t *cl` into this scope, and declare as `const` too. C99, baby.

src/game/default/g_util.c
```diff
@@ -753,3 +775,39 @@ _Bool G_IsAnimation(g_entity_t *ent, entity_animation_t anim) {
...
+	va_start(args, fmt);
+	vsprintf(string, fmt, args);
+	va_end(args);
+	
+	// look through all players
+	for (int32_t i = 0; i < sv_max_clients->integer; i++){
+		if (!g_game.entities[i + 1].in_use)
+			continue;
+		
+		ent = &g_game.entities[i + 1];
```
jdolan: More scope, `const g_entity_t *ent = .. `

jdolan:
Overall, this is great, Joe. Nice work, man. Of all of my comments on the code, the only ones that we should address before merging are the concerns around `TIMEOUT` and not running entities' `Think` functions. As long as we're not breaking things there, please go ahead and merge this.

src/game/default/g_main.h
InFerYes: I agree it should be configurable client-side but also enabling it server-side gives admins more control over this and so they can enforce it where necessary (tournaments).

src/game/default/g_client_stats.c
jar3817: That's a great idea too, but there should be a way to force all active players to take a screen/demo. Having worked on all those euro q2 leagues (edl/edge/adl/etc) in a tournament setting this is a must. I'm constantly being surprised by how retarded some players can be...

src/game/default/g_cmd.c
jar3817: that just makes too much sense...

src/game/default/g_cmd.c
jar3817: That makes sense. They can call a vote, but since the caller automatically votes yes the vote is called and then succeeded. I guess we could just not have the caller vote yes automatically if they're an admin...

src/game/default/g_cmd.c
jar3817: Now that I actually understand what you were saying (I replied right after waking up), yes it should kick everyone off teams if g_gameplay is changed to duel. That sanity check was intended to prevent spectator3 from joining a team that already has a player on it in duel mode.

src/game/default/g_cmd.c
jar3817: Yes, I changed this to a toggle. Perhaps a better way would be to have separate **ready** and **unready** commands and add a **toggleready** command. I find it handy to have a single key bound to ready/unready and toggling is very handy. 

I also often forget I've readied up and then unready which messes with all sorts of other complicated aliases and causes me to swear loudly. Part of the fun :)

src/game/default/g_client_stats.c
jdolan: Oh, okay, cool. I had no idea :) I've never been worthy of competitive play.

src/game/default/g_cmd.c
jdolan: Or revert their vote count to 1, like everyone else, and add another command for forcing the vote to pass or something? `admin vote pass` | `admin vote fail` Just spitballing..

src/game/default/g_main.c
jar3817: `g_level.time` does not stop. Items will continue to rotate and bob during a timeout, and things like water will continue to do whatever it does. From what I've tested, things like lifts, projectiles and temp entities all lock in place and continue where they were upon timing in. Even the grinder in The Pits just stops where it is and then continues on.

Instead of stopping `g_level.time` I just set `g_level.frame_num` back to where it was when timeout was called. Maybe this will screw up something in the back end that I didn't think of, but it seems to work just fine.

src/game/default/g_types.h
jar3817: wait does lengthof() actually work? I figured I had to divide the sizeof of the array by the sizeof one of the elements...

src/game/default/g_util.c
jar3817: Move into the for loop? Doesn't that re-declare it every iteration?

jdolan: No, the stack space for the variable is allocated in function scope, either way. The difference is that, by reducing the variable's AST scope, the code is easier to maintain and follow. Having to declare variables at the beginning of a function is an old C89 requirement, which is why Quake2's original code did this. But reducing variable scope and using const are both best practices for modern c development.

src/game/default/g_main.h
jar3817: prefixed...

src/game/default/g_main.c
jdolan: Hah! That's awesome. I don't think reverting `g_level.frame_num` is dangerous. `g_level.frame_num` is not a protocol-level frame number (like `sv_server.frame_num`).

src/game/default/g_types.h
jdolan: That's exactly what `lengthof` does. It's defined as `(sizeof(x) / sizeof(x[0]))`.

src/game/default/g_main.c
```diff
@@ -938,6 +948,119 @@ void G_Shutdown(void) {
 	gi.FreeTag(MEM_TAG_GAME);
 }
 
+void G_CallTimeOut(g_entity_t *ent) {
+
+	g_level.match_status |= MSTAT_TIMEOUT;
```
jdolan: I would check if `g_timeout_time` is `0` and just return here. Some servers might want to disable calling timeout.

src/game/default/g_main.c
jdolan: I think it's just another line or two in G_ClientConnect / G_ClientBegin / G_ClientRespawn ;)

jdolan:
:shipit:

Issue #176, opened 2016-09-18
Collision bug


tapir:
Here is a demo I've recorded on a test map
[collision_bug.demo.zip](https://github.com/jdolan/quetoo/files/478483/collision_bug.demo.zip)

Player sticks to the walls

jdolan:
Yep, we're well aware ;)

jdolan:
I think this is actually fixed in the latest build. Please retest!

tapir:
It's actually worst now. It doesn't slide at all. The player is forever stuck
[collision.r2758.demo.zip](https://github.com/jdolan/quetoo/files/478600/collision.r2758.demo.zip)

jdolan:
I don't have your `test.bsp` :)

tapir:
ahh sorry I knew I forgot something. Will send it tomorrow

jdolan:
We did some play testing this evening after a major collision fix (4701bac) and there were no hangups / weird clipping issues. Could you please re-test this and let us know if it's fixed for you?

jdolan:
I'm actually going to close this out as it is a duplicate of #149. Also, I can still get stuck in a few very select places (e.g. lava pit in middle arena of The Frag Pipe). So this issue is still valid, but let's track it in that older ticket.

Issue #172, opened 2016-09-18
Players interpentetrate


WickedShell:
No telefragging, if 2 clients jump at each other they will hang in air where ever they meet, even if the ground isn't there. When they randomly slide apart one player can be pushed into the floor.

WickedShell:
Demo from KaadmY https://usercontent.irccloud-cdn.com/file/XmiquUz2/collisionbug_1.demo

jdolan:
Yea, I spent a while tracking this one down tonight. I *think* that this one, and the quirky behavior of hit-scan weapons only impacting the lower half of players, are related. Specifically, I think there is something wrong with `Cm_InitBoxHull`, `Cm_SetBoxHull`, or `Cm_TransformedBoxTrace`. Is this something you could take a look at?

jdolan:
Here's the rev where this shat the bed: 16cc24165a4fcb3129b810306b53159df0106c22

jdolan:
@WickedShell please re-test. This is fixed now.

jdolan:
Aside from minor pmove issues with respect to standing on each other's heads and crouching, I don't see this as a bug anymore. Closing.

Issue #160, opened 2016-09-17
Player and corpse bounding boxes are not clipping correctly


jdolan:
Collision with projectiles, and possibly (probably?) with other players is incorrect. Hopefully this is just a matter of bounding boxes not being set correctly.

jdolan:
![quetoo009](https://cloud.githubusercontent.com/assets/643118/18609928/39760d16-7cdc-11e6-8e26-2383d8fff597.jpg)

New cvar `r_draw_entity_bounds` and major protocol change to send over bounding boxes for all mesh entities reveals that the bounds are correct. There's something else going on here.

WickedShell:
Still a nice debug tool!

jdolan:
This is halfway fixed. Projectiles now use a new solid type, `SOLID_PROJECTILE`, which encapsulates similar logic to Quake2's old `SVF_DEADMONSTER` hack. That is, projectiles can interact with `SOLID_BOX` (players) without clipping their movement.

What is still broken is hit-scan weapons. See #172.

jdolan:
\o/ fixed.

PR #180, opened 2016-09-18
#160, #172, #176

jdolan:


src/collision/cm_model.c
```diff
@@ -172,19 +172,6 @@ static void Cm_LoadBspLeafs(const d_bsp_lump_t *l) {
...
-
-	for (int32_t i = 1; i < cm_bsp.num_leafs; i++) {
-		if (!cm_bsp.leafs[i].contents) {
-			cm_bsp.empty_leaf = i;
-			break;
-		}
-	}
-
-	if (cm_bsp.empty_leaf == -1)
-		Com_Error(ERR_DROP, "Map does not have an empty leaf\n");
```
jdolan: This is just useless noise, honestly. `solid_leaf` was literally never used, and `empty_leaf` is the same as `leafs[num_leafs]`.

src/collision/cm_test.c
```diff
@@ -202,6 +202,9 @@ void Cm_InitBoxHull(void) {
  */
 int32_t Cm_SetBoxHull(const vec3_t mins, const vec3_t maxs, const int32_t contents) {
 
+	VectorCopy(mins, cm_box.brush->mins);
+	VectorCopy(maxs, cm_box.brush->maxs);
+
```
jdolan: And here it is, ladies and gentlemen. The source of an entire weekend full of frustration. 2 lines of code. It's also worth noting that this bug is almost 2 years old.

PR #181, opened 2016-09-19
Added FOV interpolation; expect glitches

kaadmy:
This still has to be tested on Mac and Windows, and works for me on Linux.
This is my first change to Quetoo, so I was learning about the codebase a bit, too.

### DIFF
--- a/src/cgame/default/cg_main.c
+++ b/src/cgame/default/cg_main.c
@@ -47,6 +47,7 @@ cvar_t *cg_draw_vitals_pulse;
 cvar_t *cg_draw_vote;
 cvar_t *cg_fov;
 cvar_t *cg_fov_zoom;
+cvar_t *cg_fov_interpolate;
 cvar_t *cg_third_person;
 cvar_t *cg_third_person_yaw;
 
@@ -106,6 +107,8 @@ static void Cg_Init(void) {
 
 	cg_fov = cgi.Cvar("cg_fov", "110.0", CVAR_ARCHIVE, NULL);
 	cg_fov_zoom = cgi.Cvar("cg_fov_zoom", "55.0", CVAR_ARCHIVE, NULL);
+	cg_fov_interpolate = cgi.Cvar("cg_fov_interpolate", "100", CVAR_ARCHIVE,
+			"Speed to change the fov at in x100 degrees/second. 0 disables, 1 is 100 deg/sec, 0.5 is 50 deg/sec");
 
 	cg_third_person = cgi.Cvar("cg_third_person", "0.0", CVAR_ARCHIVE | CVAR_LO_ONLY,
 			"Activate third person perspective.");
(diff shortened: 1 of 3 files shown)


src/cgame/default/cg_main.c
```diff
@@ -106,6 +107,8 @@ static void Cg_Init(void) {
 
 	cg_fov = cgi.Cvar("cg_fov", "110.0", CVAR_ARCHIVE, NULL);
 	cg_fov_zoom = cgi.Cvar("cg_fov_zoom", "55.0", CVAR_ARCHIVE, NULL);
+	cg_fov_interpolate = cgi.Cvar("cg_fov_interpolate", "100", CVAR_ARCHIVE,
+			"Speed to change the fov at in x100 degrees/second. 0 disables, 1 is 100 deg/sec, 0.5 is 50 deg/sec");
```
jdolan: Description is wrong base on default value.

src/cgame/default/cg_view.c
```diff
@@ -27,23 +27,53 @@
...
+			cg_fov->modified = false;
+		}
+	} else {
+		fov = target_fov;
+		cg_fov->modified = false;
+	}
+
+
+	cgi.view->fov[0] = fov / 2.0;
+
```
jdolan: I'm pretty sure this could be dramatically simplified into a couple lines, and a `VectorLerp`.

src/cgame/default/cg_main.c
kaadmy: What do you mean by wrong base for the description?

kaadmy:
> Use a single static uint32_t for time tracking

That's just how Cg_UpdateBob does it, I copied that.

I'll try using VectorLerp, I didn't know if I should/knew how to.

src/cgame/default/cg_main.c
jdolan: The description string describes values 0.0-1.0. The default value is 100. Those don't jive.

kaadmy: Oh, didn't notice that.

WickedShell: Not to make it more complicated, but why not just express it as the amount of time it takes to complete the zoom in? Slightly easier for users/editing I think, (especially since 1.0 = 100 deg/sec is a extremely weird mapping).

kaadmy: @WickedShell I originally did that, but it can't work because it's stateless.
The zoom is just an alias to change the FOV, so it can go from anywhere to anywhere, including changing the FOV target while already in the interpolation state.
Having a fixed time to interpolate between the default FOV and zoom FOV would require the zooming to be remade as an FOV modifier/factor.

WickedShell: Yeah, the alias has always seemed slightly weird to me, ideally it would just be a normal command (ideally one that just gets eaten locally by cgame without making it to the server). But that's scope creep on what you are trying to achieve :) It just has the advantage of not being dependent on the binds restoring fov.

jdolan:
@kaadmy Thanks for the patch. I took your code and rolled with it, ironing out any glitches (I think I got them all).

PR #183, opened 2016-09-19
Fix #154 Use SDL_GetTicks()

### LINKED ISSUE #154: Use SDL_GetTicks instead of platform-dependent Sys_Milliseconds code
I just realized that SDL has `SDL_GetTicks` which returns millis since `SDL_Init`. Perfect! Let's use that and kill off some non-portable code (`Sys_Milliseconds`).

tapir:


src/client/cl_cgame.c
```diff
@@ -127,7 +129,7 @@ void Cl_InitCgame(void) {
 	import.Warn_ = Com_Warn_;
 	import.Error_ = Cl_CgameError;
 
-	import.Time = Sys_Milliseconds;
+	import.Time = SDL_GetTicks;
```
jdolan: Hm. I'm glad you pointed this one out in the changeset. Rather than expose an SDL library function to the cgame, I think we can remove this pointer. The cgame can already access this value through `cgi.client->systime`.

src/console.c
```diff
@@ -46,7 +47,7 @@ static console_string_t *Con_AllocString(int32_t level, const char *string) {
 	str->size = strlen(str->chars);
 	str->length = StrColorLen(str->chars);
 
-	str->timestamp = Sys_Milliseconds();
+	str->timestamp = SDL_GetTicks();
```
jdolan: I wonder if this one could be `quetoo.time`, which is updated each frame (see `src/main/main.c`).

jdolan:
Thanks for taking a look at this one! What do you think of the changes I proposed?

src/console.c
tapir: Indeed quetoo.time should be enough for this purose :+1:

deps/minizip/zip.c
```diff
@@ -519,15 +519,17 @@ local ZPOS64_T zip64local_SearchCentralDir(const zlib_filefunc64_32_def* pzlib_f
...
       }
+    }
 
-      if (uPosFound!=0)
-        break;
+    if (uPosFound!=0)
+      break;
   }
   TRYFREE(buf);
   return uPosFound;
```
jdolan: 👍

src/console.c
```diff
@@ -47,7 +46,7 @@ static console_string_t *Con_AllocString(int32_t level, const char *string) {
 	str->size = strlen(str->chars);
 	str->length = StrColorLen(str->chars);
 
-	str->timestamp = SDL_GetTicks();
+	str->timestamp = quetoo.time;
 
 	return str;
 }
```
jdolan: 👍

tapir:
Sorry zip.c fix mixed in this one. I'll take that back and put it on the other PR to keep things clean

PR #184, opened 2016-09-19
Fix all warnings

tapir:
I fixed all the warnings and a potential bug. Feel free to nitpick :+1:

### DIFF
--- a/deps/minizip/zip.c
+++ b/deps/minizip/zip.c
@@ -526,8 +526,8 @@ local ZPOS64_T zip64local_SearchCentralDir(const zlib_filefunc64_32_def* pzlib_f
         break;
       }
 
-      if (uPosFound!=0)
-        break;
+    if (uPosFound!=0)
+      break;
   }
   TRYFREE(buf);
   return uPosFound;
--- a/deps/minizip/zip.c
+++ b/deps/minizip/zip.c
@@ -519,12 +519,14 @@ local ZPOS64_T zip64local_SearchCentralDir(const zlib_filefunc64_32_def* pzlib_f
       break;
 
     for (i=(int)uReadSize-3; (i--)>0;)
+    {
       if (((*(buf+i))==0x50) && ((*(buf+i+1))==0x4b) &&
         ((*(buf+i+2))==0x05) && ((*(buf+i+3))==0x06))
       {
         uPosFound = uReadPos+i;
         break;
       }
+    }
 
     if (uPosFound!=0)
       break;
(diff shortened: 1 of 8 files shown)


deps/minizip/zip.c
```diff
@@ -526,8 +526,8 @@ local ZPOS64_T zip64local_SearchCentralDir(const zlib_filefunc64_32_def* pzlib_f
         break;
       }
 
-      if (uPosFound!=0)
-        break;
+    if (uPosFound!=0)
+      break;
   }
```
tapir: Actually this gives a warning because there is a for statement without curly braces on top of it. So the compiler warns to show that this doesn't belong to the for. Maybe it's eve nbetter to guard the for with {}. Which one do you prefer?

src/client/cl_console.c
```diff
@@ -282,7 +282,7 @@ void Cl_InitConsole(void) {
 
 	memset(&cl_chat_console, 0, sizeof(cl_chat_console));
 	cl_chat_console.level = PRINT_CHAT | PRINT_TEAM_CHAT;
-	
+
 	cl_draw_chat = Cvar_Get("cl_draw_chat", "1", 0, "Draw recent chat messages");
 	cl_draw_notify = Cvar_Get("cl_draw_notify", "1", 0, "Draw recent console activity");
```
tapir: Oh... I have vscode which supposedly trims trailing space. I'll pay attention. Thanks.

src/client/renderer/r_material.c
```diff
@@ -311,7 +310,7 @@ static void R_DrawBspSurfaceMaterialStage(const r_bsp_surface_t *surf, const r_s
 		const vec_t *v = &r_model_state.world->verts[surf->index * 3 + i * 3];
 		const vec_t *st = &r_model_state.world->texcoords[surf->index * 2 + i * 2];
 
-		R_StageVertex(surf, stage, v, &r_state.vertex_array_3d[i * 3]);
+		R_StageVertex(v, &r_state.vertex_array_3d[i * 3]);
 
 		R_StageTexCoord(stage, v, st, &texunit_diffuse.texcoord_array[i * 2]);
```
tapir: :+1:

deps/minizip/zip.c
jdolan: I always prefer `{}` :)

tapir:
Okay, fixed your proposals. The whitespaces was not a mistake by my part it turns out. vscode sucessfully trimmed the unnecassary white space :+1: 

In the mean time I was sidetracked by a crash. If a target_light entity is pot on a map and the initial color key is not set, quetoo crashes on the strchr line. Do you try to avoid those kind of things or you just say map makers should not make silly mistakes like this? @jdolan

jdolan:
This is great, thank you!

PR #185, opened 2016-09-19
Fix proposals by jdolan

tapir:
Just I was reverting back the zip.c change you merged the PR :D Anyway here is the rest of it :+1:


jdolan:
Hah! Whoops :)

Issue #171, opened 2016-09-18
Sound spatialization code still sucks


jdolan:
The sound mixing is fucking horrible. Either rewrite it, or replace it when OpenAL.

tapir:
:+1:  for OpenAL. Good 3D mixing is a very hard job to do correctly.

jdolan:
I found some definite bugs here around channels never being freed up for new voices to play on them. With some initial play testing with 5 players, I didn't notice any missing samples. Marking this one fixed. We can revisit OpenAL when we have time / specific interest in it.

PR #194, opened 2016-09-22
#171 Big ol' s_mix refactor

jdolan:


src/client/sound/s_mix.c
```diff
@@ -69,120 +96,112 @@ _Bool S_SpatializeChannel(s_channel_t *ch) {
...
+				s_env.num_active_channels++;
+			} else {
+				
+				if (ch->start_time) {
+					Mix_HaltChannel(i);
+				} else {
+					S_FreeChannel(i);
+				}
+			}
+		}
```
jdolan: Here's the real stinker. Mix_HaltChannel doesn't call the Mix_ChannelFinished callback if the channel was never started. So failing to free channels that were never started causes us to "leak" culled channels, and eventually this blows our available channels.

PR #203, opened 2016-09-25
Added handicap, should fix #32

### LINKED ISSUE #32: Add a handicap feature (q3a-style)
The title says it all. Would be cool to have a handicap feature to allow people to play against others who are on a different skill level. I can't play some of my friends because it makes no sense dut to our differences in skill (match results are usually something like 30:0).

How handicap works: * each player has a handicap value that he can set in his config. It defaults to 100. * the weaker player does nothing (so he has 100) * the stronger player specifies a lower handicap value (like 70). This affects (weakens) him in several ways:

    he starts with <handycap-value> health instead of 100
    he can't pick up more than <handycap-value> health with normal medkits (but more with +5 bubbles and MH)
    if he hurts another player, he only does <handycap-value>% (per cent) of the normal damage 

There's no need to implement all of these (and others may also be introduced, of course), but a feature that does something similar to the things described here would allow me to attract a few of my friends to q2w. :)

Changed 14 months ago by jdolan

    owner changed from jdolan to wickedshell
    status changed from new to assigned

A userinfo cvar on the client could set this for the server, and the game module could then honor it in G_ClientRespawn_.

kaadmy:
As the title says, this adds handicap in the form of reducing reducing max health and max overheal health.

### DIFF
--- a/src/client/cl_main.c
+++ b/src/client/cl_main.c
@@ -44,6 +44,7 @@ cvar_t *name;
 cvar_t *password;
 cvar_t *rate;
 cvar_t *skin;
+cvar_t *handicap;
 
 cvar_t *qport;
 
@@ -514,6 +515,7 @@ static void Cl_InitLocal(void) {
 	password = Cvar_Get("password", "", CVAR_USER_INFO, NULL);
 	rate = Cvar_Get("rate", "0", CVAR_USER_INFO | CVAR_ARCHIVE, NULL);
 	skin = Cvar_Get("skin", "qforcer/default", CVAR_USER_INFO | CVAR_ARCHIVE, NULL);
+	handicap = Cvar_Get("handicap", "100", CVAR_USER_INFO | CVAR_ARCHIVE, NULL);
 
 	qport = Cvar_Get("qport", va("%d", Random() & 0xff), 0, NULL);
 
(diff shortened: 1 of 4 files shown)


src/client/cl_main.c
```diff
@@ -44,6 +44,7 @@ cvar_t *name;
 cvar_t *password;
 cvar_t *rate;
 cvar_t *skin;
+cvar_t *handicap;
```
jdolan: Should be a corresponding extern declaration of this in cl_main.h.

src/game/default/g_types.h
```diff
@@ -698,6 +698,8 @@ typedef struct {
 	char skin[MAX_QPATH];
 	g_hand_t hand;
 
+	vec_t handicap; // handicap inverse percentage from 0 to 100
+
```
jdolan: Will this eventually influence G_Damage?

kaadmy: I don't know, I should be able to add it easily.

jdolan: I never used Q3's handicap feature, so I'm not familiar with all of it's functionality. But from what @maci0 said above, apparently your handicap would also make your weapons that much less effective. I do not have a strong opinion on whether that is needed or not.

kaadmy: I'll add it to allow RJing still :)

src/client/cl_main.c
kaadmy: Okay, it's fixed

kaadmy:
Oh, I was still changing this branch :(

jdolan:
Restored.

PR #207, opened 2016-09-25
Paril's visual studio work

jdolan:


src/server/sv_console.c
```diff
@@ -20,10 +20,23 @@
...
+#endif
+
+#include <curses.h>
+
+// Visual Studio hack
+#if defined(_MSC_VER)
+#undef bool
+#define bool _Bool
+#endif
+
```
jdolan: What's the deal here? Do I introduce "bool", or does VS not like curses.h or something?

src/server/sv_console.c
```diff
@@ -210,6 +223,7 @@ static void Sv_DrawConsole_Buffer(void) {
 	}
 
 	Sv_DrawConsole_Color(CON_COLOR_DEFAULT);
+	Mem_Free(lines);
```
jdolan: Another g_malloc0 use case.

src/sys.c
```diff
@@ -23,13 +23,28 @@
...
 #include <shlobj.h>
 #define RTLD_NOW 0
 #define dlopen(file_name, mode) LoadLibrary(file_name)
-#define dlerror() "Windows.. go figure."
+
+const char *dlerror()
+{
+	static char num_buffer[32];
+
+	DWORD err = GetLastError();
```
jdolan: const DWORD ? ;)

src/sys.c
```diff
@@ -178,6 +193,52 @@ void Sys_Backtrace(void) {
...
+		}
+
+		fflush(stderr);
+		Mem_Free(symbol);
+	}
+	else
+	{
+		fprintf(stderr, "Couldn't get stack trace.\n");
+		fflush(stderr);
+	}
```
jdolan: This is awesome! Because the game is crashing, I would refrain from using Mem_Malloc here. There be dragons.

src/tools/quemap/main.c
```diff
@@ -19,10 +19,10 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  */
 
-#include <SDL2/SDL.h>
-
 #include "quemap.h"
 
+#include <SDL2/SDL.h>
+
```
jdolan: Just curious, we can't import SDL.h first? Typically I include external deps first, then our includes.

src/tools/quemap/main.c
```diff
@@ -552,4 +552,6 @@ int32_t main(int32_t argc, char **argv) {
 	Com_Print("%d Seconds\n", (int32_t) (duration % 60));
 
 	Com_Shutdown(NULL);
+
+	return 0;
```
jdolan: Hah, good one.

src/tools/quemap/monitor.c
```diff
@@ -259,4 +277,5 @@ void Mon_Shutdown(const char *msg) {
 	mon_backlog = NULL;
 
 	Net_Shutdown();
+#endif
```
jdolan: Was libxml2 too difficult to wrangle into VS? I gotta be honest, without the XML monitoring, quemap is pretty useless on Windows. The XML monitoring is how it talks to GtkRadiant. It's absolutely stupid, but this is now GtkRadiant does it.

src/tools/update/main.c
```diff
@@ -113,6 +126,9 @@ static char *convert_cygwin_path(const char *path) {
 		}
 	}
 
+	*out++ = '"';
+	*out++ = 0;
+
 	return cygpath;
```
jdolan: Ah, we have double quotes here, now. I added the call to `va` down below, which I think is a little cleaner (fewer lines).

jdolan: This is resolved now. I just reverted the path function to that of master.

Quetoo.vs15/src/config.h
```diff
@@ -0,0 +1,86 @@
...
+#define __thread __declspec(thread)
+#endif
+
+#ifndef usleep
+#define usleep(t) Sleep(t / 1000)
+#endif
+
+#ifndef S_ISDIR
+#define S_ISDIR(m) (((m) & S_IFDIR) == S_IFDIR)
+#endif
```
Paril: MSVC has stat.h, it just doesn't include S_ISDIR for some reason. I don't think I can cascade the includes either (as in, include MSVC' stat.h from our stat.h), which is why I didn't make a stat.h.

unistd is a slight exception; I specifically designed this PR to not require code modifications to the third party libraries. I added unistd so that one of the libraries which tries to include it won't fail (and will get ssize_t, which is why it requested unistd in the first place.. on MSVC, size_t is implicitly defined by the compiler but ssize_t does not exist. Go figure :) )

Quetoo.vs15/src/main/winmain.c
```diff
@@ -0,0 +1,17 @@
...
+
+int CALLBACK WinMain(
+  _In_ HINSTANCE hInstance,
+  _In_ HINSTANCE hPrevInstance,
+  _In_ LPSTR     lpCmdLine,
+  _In_ int       nCmdShow
+)
+{
+	return main(__argc, __argv);
+}
```
Paril: No, this is to prevent requiring SDLmain.lib; it's basically the same thing that SDLmain does, but apparently ming doesn't need it/has it automatically.

deps/minizip/ioapi.h
```diff
@@ -38,6 +38,10 @@
         #ifndef _FILE_OFFSET_BIT
                 #define _FILE_OFFSET_BIT 64
         #endif
+#else
+#if (defined(_MSC_VER))
+		#define __attribute__(...)
+#endif
```
Paril: Up to you, Minizip is third party but at least you are able to modify it!

src/client/cl_console.c
```diff
@@ -89,6 +89,8 @@ static void Cl_DrawConsole_Buffer(void) {
 		g_free(lines[i]);
 		y += ch;
 	}
+
+	Mem_Free(lines);
 }
```
Paril: I was worried I missed one of these, I thought they were all buffered. In the case of one called per frame, I'd recommend using alloca instead - it allocates on the stack, and is the closest to what the VLA accomplished. If need be, we can #if _MSC_VER it as well so that it's not required for non-MSVC.

src/console.c
```diff
@@ -175,10 +175,12 @@ void Con_Append(int32_t level, const char *string) {
...
 		}
 	} else {
-		char stripped[strlen(string + 1)];
+		char *stripped = Mem_Malloc(sizeof(char) * strlen(string) + 1);
 
 		StripColors(string, stripped);
 		fputs(stripped, stdout);
+
+		Mem_Free(stripped);
 	}
```
Paril: Or alloca, see above

src/image.c
```diff
@@ -251,4 +254,7 @@ _Bool Img_WriteJPEG(const char *path, byte *data, uint32_t width, uint32_t heigh
 
 	fclose(f);
 	return true;
+#else
+	return false;
+#endif
```
Paril: Yep, that's why I disabled it for now. No sense in me getting jpeglib working if we don't need it

src/server/sv_console.c
Paril: pdcurses breaks the definition for bool. It's an ugly fix, and I didn't want to edit the third party lib so we can update if need be.

src/sys.c
Paril: Yep I realized that a bit too late, lol

src/tools/quemap/main.c
Paril: This file was a special exception. I needed config.h to be included to be able to tell SDL to stop trying to get SDL_main working.. it's an unconditional include on Windows, even if you want to handle WinMain yourself (as I do in winmain.c)

src/tools/quemap/main.c
Paril: It wouldn't stop complaining

src/tools/quemap/monitor.c
Paril: Oh, I wasn't sure if it was necessary or just an external cool thingy. I'll see about getting libxml included in a PR.

src/tools/update/main.c
Paril: Sounds good!

Quetoo.vs15/src/config.h
jdolan: This is probably another case where, like in Objectively/MVC, we want to keep POSIX compat stuff outside of config.h.

src/client/cl_console.c
```diff
@@ -157,7 +157,7 @@ void Cl_DrawNotify(void) {
 		con.whence = cl.systime - cl_notify_time->value * 1000;
 	}
 
-	char *lines[con.height];
+	char **lines = alloca(sizeof(char*) * con.height);
```
jdolan: Looks like we can roll some of these back to VLA.

src/net/net_udp.c
```diff
@@ -19,7 +19,9 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  */
 
+#if !defined(_MSC_VER)
```
jdolan: This would be a good spot to introduce config.h and check `HAVE_SYS_TIME_H`.

PR #218, opened 2016-09-26
Fixes

tapir:
Some small fixes

### DIFF
--- a/src/game/default/g_client_stats.c
+++ b/src/game/default/g_client_stats.c
@@ -54,7 +54,7 @@ void G_ClientToIntermission(g_entity_t *ent) {
 
 	ent->client->locals.ammo_index = 0;
 	ent->client->locals.pickup_msg_time = 0;
-	
+
 	// take a screenshot if we're supposed to
 	if (g_force_screenshot->integer == 1){
 		G_ClientStuff(ent, "r_screenshot\n");
@@ -120,7 +120,7 @@ static size_t G_UpdateScores(g_score_t *scores) {
 
 	// and optionally concatenate the team scores
 	if (g_level.teams || g_level.ctf) {
-		memset(s, 0, sizeof(s) * 2);
+		memset(s, 0, sizeof(*s) * 2);
 
 		s->client = MAX_CLIENTS;
 		s->score = g_team_good.score;
--- a/src/server/sv_entity.c
+++ b/src/server/sv_entity.c
@@ -177,7 +177,7 @@ static void Sv_ClientVisibility(const vec3_t org, byte *pvs, byte *phs) {
 		byte cluster_pvs[MAX_BSP_LEAFS >> 3];
 		Cm_ClusterPVS(clusters[i], cluster_pvs);
 
-		for (size_t i = 0; i < sizeof(pvs); i++) {
+		for (size_t i = 0; i < sizeof(cluster_pvs); i++) {
 			pvs[i] |= cluster_pvs[i];
 		}
 	}
--- a/src/tools/quemap/bspfile.c
+++ b/src/tools/quemap/bspfile.c
@@ -577,7 +577,7 @@ void ParseEntities(void) {
 void UnparseEntities(void) {
 	char *buf, *end;
 	epair_t *ep;
-	char line[2048];
+	char line[2096];
 	int32_t i;
 	char key[1024], value[1024];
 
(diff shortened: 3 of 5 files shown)


src/game/default/g_client_stats.c
```diff
@@ -120,7 +120,7 @@ static size_t G_UpdateScores(g_score_t *scores) {
 
 	// and optionally concatenate the team scores
 	if (g_level.teams || g_level.ctf) {
-		memset(s, 0, sizeof(s) * 2);
+		memset(s, 0, sizeof(g_score_t));
 
 		s->client = MAX_CLIENTS;
 		s->score = g_team_good.score;
```
jdolan: Ah, there *is* a bug here, tho. It should be: `memset(s, 0, sizeof(*s) * 2);`

tapir: yeap size of pointer not struct :+1:

src/server/sv_entity.c
```diff
@@ -177,7 +177,7 @@ static void Sv_ClientVisibility(const vec3_t org, byte *pvs, byte *phs) {
 		byte cluster_pvs[MAX_BSP_LEAFS >> 3];
 		Cm_ClusterPVS(clusters[i], cluster_pvs);
 
-		for (size_t i = 0; i < sizeof(pvs); i++) {
+		for (size_t i = 0; i < MAX_BSP_LEAFS >> 3; i++) {
 			pvs[i] |= cluster_pvs[i];
 		}
 	}
```
tapir: so without really knowing what the code does, do we really wanto get the size of a byte pointer?

jdolan: sizeof(byte[64]) is 64.

src/shared.c
```diff
@@ -1026,6 +1026,14 @@ _Bool ValidateUserInfo(const char *s) {
...
 	char newi[MAX_USER_INFO_STRING], *v;
 
+	if (!value) {
+		return;
+	}
+
+	if (*value == '\0') {
+		return;
+	}
+
```
tapir: Interesting. It must be a compiler thing. Because both sides of the || is evaluated right? How come when value is NULL,  *value == 0 doesn't crash?

jdolan: No, just look at the previous ordering: the old checks take place _after_ the call to DeleteUserInfo. That was on purpose.

src/tools/quemap/bspfile.c
```diff
@@ -577,7 +577,7 @@ void ParseEntities(void) {
 void UnparseEntities(void) {
 	char *buf, *end;
 	epair_t *ep;
-	char line[2048];
+	char line[2096];
 	int32_t i;
```
tapir: Not really based on something but if look at this line

```
sprintf(line, "\"%s\" \"%s\"\n", key, value);
```
key and value are allowed to hold 2048 bytes in total and we still have " chars in between.
In real life this is not a problem probably because no key/value is 1024 byte long but it's a possible overflow

jdolan: I see, okay.

src/server/sv_entity.c
tapir: sizeof(*byte) = 8 on 64 bit systems and 4 on 32 bit systems (I think)
I've just tested this on my system and sizeof(pvs) was 8.

src/shared.c
tapir: I see how passing null returns the function early (just after del user info) so the intended use `SetUserInfo(s, "active", NULL);` to delete user info is met nicely.

The thing is when `value = NULL` doing a `*value` will segfault. In expression `!value || *value == '\0'` one might tought the same thing might happen. If it doesn't happen I think its because compiler is smart enough to see it and not evaluate right hand side of the || when `value == NULL`

tapir: So we can still keep the order after DelUserInfo and separate || to two different ifs.

src/server/sv_entity.c
jdolan: Oh, shit. Yes == `sizeof(cluster_pvs)` is what I meant here. Which yields 8192.

tapir: ah :+1:

src/shared.c
jdolan: There's no reason to separate them. All compilers short-circuit compound conditionals.

tapir:
Changed back some commits. Cleaned history. Ready to roll!

PR #221, opened 2016-09-27
Fix #201 by disabling TextInput until it's absolutely necessary

### LINKED ISSUE #201: Tilde sometimes causes a write of backtick to the console
Hitting tilde (on MSVC build, at least) causes a backtick to get written to the console almost every time. It seems that after the text input is initialized, it immediately registers a text input event for the backtick.

tapir:
What a little bug.

While trying to debug this, I've found out that SDL_StopTextInput() doesn't necessarily work. Somewhere along the Init() sequence it gets undone eventhough there are no explicit SDL_StartTextInput() calls and that SDL is already initialized.

### DIFF
--- a/src/client/cl_keys.c
+++ b/src/client/cl_keys.c
@@ -79,7 +79,8 @@ void Cl_SetKeyDest(cl_key_dest_t dest) {
  */
 static void Cl_KeyConsole(const SDL_Event *event) {
 
-	SDL_StartTextInput();
+	if (SDL_GetEventState(SDL_TEXTINPUT) != SDL_ENABLE)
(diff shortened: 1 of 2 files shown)


tapir:
KeyConsole and KeyChat are also called when KEY_UP event is emitted. That's why, the first text event is not missed. It goes like this:
- Console or chat key id down
- Dest is set accordingly
- Console or chat key is up
- Because dest was already set KeyConsole or KeyChat are called thus TextInput started.

tapir:
@Paril tested 959df6e3922863a8be7a00ce1f1fdbd484c47e9c and it works so this is not needed anymore.

PR #228, opened 2016-10-01
Added very, very buggy water ripples. This is for archive/testing pur…

jdolan:
…poses only.


jdolan:
Not bad! The effect texture needs some work, and I'm sure the effect emission in `Cg_WaterRippleEffect` could use tweaking. The ripples are placed incorrectly for client entities. But this is a good start, and I'd like to see it eventually merged.

kaadmy:
No plz
This is incredibly broken, you should try it first...
Stuff like hand grenades emmiting a water ripple when thrown if the map's center point is in water, shooting underwater leaves floating ripples in the water, and water ripples that supposedly are on the surface are either above/below the actual surface.

PR #240, opened 2016-10-13
Feature/gl30

jdolan:
Opening a PR for Paril's OpenGL 3.0 branch for review purposes only.


Quetoo.vs15/src/config.h
```diff
@@ -34,6 +34,10 @@
 		g_list_foreach(list, (GFunc)func, NULL); \
 		g_list_free(list);
 
+#define g_slist_free_full(list, func)	\
+		g_slist_foreach(list, (GFunc)func, NULL); \
+		g_slist_free(list);
+
```
jdolan: As a low-priority item, it would be great to either move this into a glib-specific include, or really, just snag an updated glib.dll and glib.h that includes this stuff. It has to exist, no? Tons of GTK+ software is available on Windows..

src/cgame/default/Makefile.am
```diff
@@ -66,5 +66,6 @@ cgame_la_LDFLAGS = \
 cgame_la_LIBADD = \
 	viewcontrollers/libviewcontrollers.la \
 	$(top_builddir)/src/game/default/libpmove.la \
-	$(top_builddir)/src/libswap.la
+	$(top_builddir)/src/libswap.la \
+	$(top_builddir)/src/libmatrix.la
```
jdolan: I'm okay with adding libmatrix.la as a dependency for cgame. We do want to be careful, in general, with introducing deps to cgame / game, but this one probably makes sense.

src/cgame/default/cg_entity_trail.c
```diff
@@ -456,15 +456,15 @@ static void Cg_LightningTrail(cl_entity_t *ent, const vec3_t start, const vec3_t
...
 
 		p->part.scale = 8.0;
 		p->part.scroll_s = -8.0;
 
 		VectorCopy(pos, p->part.org);
 
 		if (dist <= 48.0) {
-			VectorScale(dir, dist, delta);
+			VectorScale(dir, -dist, delta);
 			offset = 0.0;
```
jdolan: Hm. Was this a bug?

src/cgame/default/views/PlayerModelView.c
```diff
@@ -185,7 +191,9 @@ static void render(View *self, Renderer *renderer) {
 		glMatrixMode(GL_PROJECTION);
 		glLoadIdentity();
 
-		glOrtho(0, cgi.context->window_width, cgi.context->window_height, 0, -1, 1);
+		Matrix4x4_FromOrtho(&mat, 0.0, cgi.context->window_width, cgi.context->window_height, 0.0, -1.0, 1.0);
+
+		glLoadMatrixf((GLfloat *) mat.m);
 
 		glMatrixMode(GL_MODELVIEW);
 		glLoadIdentity();
```
jdolan: Juicy. There is also some slop in this View for calculating the matrix of the linked models (head, torso, weapon). Not sure if any of this `matrix4x4_t` cleanup will allow us to remove that (doubtful). But just so you're aware of it..

src/client/renderer/r_array.c
```diff
@@ -245,3 +245,17 @@ void R_ResetArrayState(void) {
 	r_array_state.arrays = arrays;
 }
 
+/**
+ * @brief
+ */
+void R_DrawArrays(GLenum type, GLint start, GLsizei count) {
+
+	if (!r_state.active_program)
+		return;
```
jdolan: A Com_Warn / R_Error here is probably appropriate before returning.

src/client/renderer/r_array.c
```diff
@@ -245,3 +245,17 @@ void R_ResetArrayState(void) {
...
+
+	// TODO: should we figure a better way to "dirty" the matrices so that the memcmp doesn't run every time?
+	if (r_state.active_program->UseMatrices)
+		r_state.active_program->UseMatrices(&r_view.projection_matrix, &r_view.modelview_matrix, &r_view.texture_matrix);
+
```
jdolan: Sure. I would say, introduce R_SetModelViewMatrix, R_SetProjectionMatrix, R_SetTextureMatrix, just like the other state management functions in r_state.c. Those functions can do the comparison, and flip a flag if the new values are dirty.

src/client/renderer/r_bsp.c
```diff
@@ -320,7 +320,7 @@ void R_DrawBspLeafs(void) {
 			if ((*s)->vis_frame != r_locals.vis_frame)
 				continue;
 
-			glDrawArrays(GL_POLYGON, (*s)->index, (*s)->num_edges);
+			R_DrawArrays(GL_TRIANGLE_FAN, (*s)->index, (*s)->num_edges);
 		}
```
jdolan: This is great. After this refactor, I'm eager to tear into the renderer and move world geometry (at least) to indexed arrays, so we'll have R_DrawElements as well. I dig it.

src/client/renderer/r_draw.c
```diff
@@ -301,7 +301,7 @@ static void R_DrawChars(void) {
 		R_BindArray(GL_TEXTURE_COORD_ARRAY, GL_FLOAT, chars->texcoords);
 		R_BindArray(GL_VERTEX_ARRAY, GL_SHORT, chars->verts);
 
-		glDrawArrays(GL_QUADS, 0, chars->vert_index / 2);
+		R_DrawArrays(GL_QUADS, 0, chars->vert_index / 2);
```
jdolan: Missed a GL_TRIANGLE_FAN.

src/client/renderer/r_draw.c
```diff
@@ -378,7 +378,7 @@ static void R_DrawFills(void) {
 	R_BindArray(GL_VERTEX_ARRAY, GL_SHORT, r_draw.fill_arrays.verts);
 	R_BindArray(GL_COLOR_ARRAY, GL_UNSIGNED_BYTE, r_draw.fill_arrays.colors);
 
-	glDrawArrays(GL_QUADS, 0, r_draw.fill_arrays.vert_index / 2);
+	R_DrawArrays(GL_QUADS, 0, r_draw.fill_arrays.vert_index / 2);
```
jdolan: Another.

src/cgame/default/views/PlayerModelView.c
```diff
@@ -150,16 +150,22 @@ static void render(View *self, Renderer *renderer) {
...
-		glRotatef(-90.0, 1.0, 0.0, 0.0);
-		glRotatef( 90.0, 0.0, 0.0, 1.0);
-		glTranslatef(64.0, 0.0, -8.0);
+		Matrix4x4_ConcatRotate(&mat, -90.0, 1.0, 0.0, 0.0);
+		Matrix4x4_ConcatRotate(&mat,  90.0, 0.0, 0.0, 1.0);
+		Matrix4x4_ConcatTranslate(&mat, 64.0, 0.0, -8.0);
+
+		glLoadMatrixf((GLfloat *) mat.m);
 		
 		glPushMatrix();
```
jdolan: It looks like you've added R_Push / PopMatrix to the renderer. We should therefore expose those via cg_export_t, and use `cge.PushMatrix` and `cge.PopMatrix` here.

src/client/renderer/r_entity.c
```diff
@@ -285,9 +285,9 @@ static void R_DrawEntityBounds(const r_entities_t *ents, const vec4_t color) {
 
 		memcpy(r_state.vertex_array_3d, verts, sizeof(verts));
 
-		glDrawArrays(GL_QUADS, 0, lengthof(verts));
+		R_DrawArrays(GL_QUADS, 0, lengthof(verts));
```
jdolan: Hm.. I guess I'll have to refactor these to use GL_LINES.

src/client/renderer/r_material.c
```diff
@@ -127,43 +127,43 @@ static void R_StageTextureMatrix(const r_bsp_surface_t *surf, const r_stage_t *s
...
 	if (stage->flags & STAGE_SCROLL_S)
-		glTranslatef(stage->scroll.ds, 0.0, 0.0);
+		Matrix4x4_ConcatTranslate(&r_view.texture_matrix, stage->scroll.ds, 0.0, 0.0);
 
 	if (stage->flags & STAGE_SCROLL_T)
-		glTranslatef(0.0, stage->scroll.dt, 0.0);
+		Matrix4x4_ConcatTranslate(&r_view.texture_matrix, 0.0, stage->scroll.dt, 0.0);
 
 	identity = false;
 }
```
jdolan: 👍

src/client/renderer/r_program.c
```diff
@@ -155,23 +155,23 @@ void R_ProgramParameter4fv(r_uniform4fv_t *variable, const GLfloat *value) {
...
 
 	if (memcmp(&variable->value.mat4, value, sizeof(variable->value.mat4)) == 0)
-		return;
+		return false;
 
 	memcpy(&variable->value.mat4, value, sizeof(variable->value.mat4));
 	qglUniformMatrix4fv(variable->location, 1, false, (GLfloat *) variable->value.mat4.m);
 
 	R_GetError(variable->name);
+	return true;
```
jdolan: Is this return value used? If not, I'd be happy just leaving it as void.

src/client/renderer/r_program.c
```diff
@@ -267,6 +270,55 @@ void R_ShutdownPrograms(void) {
...
+
+	GError *error = NULL;
+	gchar *output = g_regex_replace_eval(shader_preprocess_regex, emplaced->str, emplaced->len, 0, 0, R_PreprocessShader_EvalCallback, NULL, &error);
+
+	if (error)
+		Com_Warn("Error preprocessing shader: %s", error->message);
+
+	return output;
+}
+
```
jdolan: 👍

src/client/renderer/r_state.c
```diff
@@ -558,6 +571,40 @@ void R_UseMaterial(const r_material_t *material) {
...
+	if (r_state.matrix_stack_index == 0)
+		Com_Error(ERR_DROP, "Matrix stack underflow");
+
+	Matrix4x4_Copy(&r_view.modelview_matrix, &r_state.matrix_stack[--r_state.matrix_stack_index]);
+}
+
+/**
+ * @brief Uploads matrices to the currently loaded program.
+ */
+void R_UploadMatrices(void) {
```
jdolan: Maybe name this one R_UseMatrices.

src/cgame/default/Makefile.am
Paril: It may not be required in the end, but an doing my best to keep it up to date with my projects.

src/cgame/default/cg_entity_trail.c
Paril: Yep, caused lightning bolt at end of chain to go backwards. Now you can shoot lightning bolt at a close wall and still see zappyzap

src/client/renderer/r_array.c
Paril: Not a bad idea. The only thing is that it has to be per+program because one program may have the old matrix still, so I have to think about that one. Also want to avoid creating too many matrix4x4 instances

src/client/renderer/r_draw.c
Paril: Yeah my IDE kept crashing because of a preview intellisense update so I forgot about these lol

src/client/renderer/r_array.c
jdolan: I would propose that R_UseProgram invalidates the matrices, then, too.

Paril:
That'd still end up attempting a matrix upload pretty much every time since
shaders do switch quite often. Tbf the only near-constant array is the
projection one.. unless it becomes a perf concern maybe it's just
microoptimization  at its worst

src/client/renderer/r_array.c
Paril: That seems too micro-optimize for my liking, especially since UseProgram will happen between every major step anyhow and no two entities will use the same matrix. I'll only look into that TODO if it shows to be a performance concern.

src/client/renderer/r_draw.c
Paril: Also it won't work for these anyways - TRIANGLE_FAN will only work in spots where GL_POLYGON would work. GL_QUADS will eventually have to be changed into GL_TRIANGLES. Sooooup

src/client/renderer/r_light.c
```diff
@@ -182,29 +185,19 @@ void R_EnableLights(uint64_t mask) {
 	if (mask) { // enable up to MAX_ACTIVE_LIGHT sources
 		const r_light_t *l = r_view.lights;
 
-		vec4_t position = { 0.0, 0.0, 0.0, 1.0 };
-		vec4_t diffuse = { 0.0, 0.0, 0.0, 1.0 };
-
 		for (uint16_t i = 0; i < r_view.num_lights; i++, l++) {
 
-			if (j == MAX_ACTIVE_LIGHTS)
+			if (j == r_state.max_lights)
```
Paril: Not sure about that one. You should never call UseLight with a value that is beyond r_state.max_lights, it will crash. The game is set up to not push more than the max hardware lights that is allowed by the program.

src/client/renderer/r_program.c
Paril: See r_program_default

src/client/renderer/r_light.c
jdolan: Sure, I'm just saying that the name is kind of misleading. The previous constant was named `MAX_ACTIVE_LIGHTS`, which is more accurate and doesn't conflate "how many lights can be in a scene" with "how many lights can I enable for a given primitive."

Paril: Oh I see what you mean, yeah. Will do.

src/client/renderer/r_array.c
```diff
@@ -41,14 +41,26 @@ static r_array_state_t r_array_state;
 int32_t R_ArraysMask(void) {
 	uint32_t mask = R_ARRAY_MASK_VERTEX;
 
+	_Bool do_interpolation = r_view.current_entity && r_view.current_entity->old_frame != r_view.current_entity->frame && IS_MESH_MODEL(r_view.current_entity->model);
```
jdolan: To be truly safe, I would validate the frame indices against the model. Or at least `assert(ent->frame < md3->num_frames)`. etc..

src/client/renderer/r_mesh.c
```diff
@@ -157,82 +157,6 @@ void R_UpdateMeshModelLighting(const r_entity_t *e) {
...
-	}
-
-	R_UploadToBuffer(&r_state.buffer_vertex_array, 0, vert_index * sizeof(float), r_state.vertex_array);
-
-	if (r_state.lighting_enabled) {
-	
-		R_UploadToBuffer(&r_state.buffer_normal_array, 0, vert_index * sizeof(float), r_state.normal_array);
-	}
-}
-
```
jdolan: CYA!

src/client/renderer/r_array.c
```diff
@@ -197,51 +182,107 @@ void R_ResetArrayState(void) {
...
+	R_SetupAttributes();
+
+	R_UseMatrices();
+
+	R_UseAlphaTest();
+
+	R_UseCurrentColor();
+
+	R_UseFog();
+}
```
jdolan: Nice!

src/client/renderer/r_context.c
```diff
@@ -40,6 +40,19 @@ static void R_SetWindowIcon(void) {
...
 /**
+ * @brief Attempts to create the specified versioned context, returns true on success and false on failure.
+ */
+static _Bool R_InitGLContext(const int major, const int minor) {
+
+	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, major);
+	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, minor);
+		
+	return (r_context.context = SDL_GL_CreateContext(r_context.window)) != NULL;
+}
```
jdolan: This helper seems kind of superfluous now that we're only trying one GL version.

src/client/renderer/r_draw.c
```diff
@@ -22,40 +22,65 @@
...
 	uint32_t vert_index;
+	r_buffer_t vert_buffer;
 
-	GLfloat texcoords[MAX_CHARS * 4 * 2];
+	GLfloat texcoords[MAX_CHARS * sizeof(vec2_t)];
 	uint32_t texcoord_index;
+	r_buffer_t texcoord_buffer;
 
-	GLbyte colors[MAX_CHARS * 4 * 4];
+	GLfloat colors[MAX_CHARS * sizeof(vec4_t)];
```
jdolan: I can see why you're suggesting unsigned bytes for color attributes.. that's a lot of memory.

src/client/renderer/r_draw.c
```diff
@@ -324,42 +419,63 @@ static void R_DrawChars(void) {
...
+
+		color[3] = a;
 	}
 
 	// duplicate color data to all 4 verts
-	memcpy(&r_draw.fill_arrays.colors[r_draw.fill_arrays.color_index + 0], color, 4);
-	memcpy(&r_draw.fill_arrays.colors[r_draw.fill_arrays.color_index + 4], color, 4);
-	memcpy(&r_draw.fill_arrays.colors[r_draw.fill_arrays.color_index + 8], color, 4);
-	memcpy(&r_draw.fill_arrays.colors[r_draw.fill_arrays.color_index + 12], color, 4);
+	memcpy(&r_draw.fill_arrays.colors[r_draw.fill_arrays.color_index + 0], color, 4 * sizeof(float));
```
jdolan: Probably use `sizeof(vec4_t)` here?

src/client/renderer/r_draw.c
```diff
@@ -370,59 +486,72 @@ static void R_DrawFills(void) {
...
+			color[i] = ((img_palette[c] >> (i * 8)) & 0xFF) / 255.0;
+
+		color[3] = a;
 	}
 
 	// duplicate color data to all 4 verts
-	memcpy(&r_draw.line_arrays.colors[r_draw.line_arrays.color_index + 0], color, 4);
-	memcpy(&r_draw.line_arrays.colors[r_draw.line_arrays.color_index + 4], color, 4);
+	memcpy(&r_draw.line_arrays.colors[r_draw.line_arrays.color_index + 0], color, 4 * sizeof(float));
+	memcpy(&r_draw.line_arrays.colors[r_draw.line_arrays.color_index + 4], color, 4 * sizeof(float));
```
jdolan: And here.

src/client/renderer/r_main.c
```diff
@@ -64,24 +64,25 @@ cvar_t *r_lighting;
 cvar_t *r_line_alpha;
 cvar_t *r_line_width;
 cvar_t *r_materials;
+cvar_t *r_max_lights;
```
jdolan: Did we decide on a better name for this sucker? `r_lighting_max_dynamic` ?

src/client/renderer/r_main.c
```diff
@@ -512,17 +518,78 @@ static void R_InitConfig(void) {
...
+
+			// up here, we are assumed to be either at EOF or at a GL_ boundary.
+			// just be sure first.
+			while (*extensions_in && *extensions_in != 'G') {
+				++extensions_in;
+			}
+
+			// done, print what's left
+			if (!*extensions_in) {
+
```
jdolan: A lot of work for something that we only ever print in `verbose` mode ;)

src/client/renderer/r_main.h
```diff
@@ -44,26 +44,30 @@ extern cvar_t *r_lighting;
...
-extern cvar_t *r_vertex_buffers;
 extern cvar_t *r_warp;
 extern cvar_t *r_width;
 extern cvar_t *r_windowed_height;
 extern cvar_t *r_windowed_width;
 
 extern r_view_t r_view;
 
+// convenience shortcuts for matrices
+#define projection_matrix r_view.active_matrices[R_MATRIX_PROJECTION]
```
jdolan: I'd maybe prefix these constants with `r_`, but it's not a huge deal.

src/client/renderer/r_mesh.c
```diff
@@ -157,96 +157,32 @@ void R_UpdateMeshModelLighting(const r_entity_t *e) {
...
-				VectorCopy(r_mesh_state.normals[tri[1]], (&r_state.normal_array[vert_index + 3]));
-				VectorCopy(r_mesh_state.normals[tri[2]], (&r_state.normal_array[vert_index + 6]));
-			}
-
-			vert_index += 9;
-		}
-	}
-}
-
-/**
```
jdolan: YES!

src/client/renderer/r_state.h
```diff
@@ -37,43 +47,95 @@ void R_Setup2D(void);
...
 // texunits maintain multitexture state
+#define MAX_GL_TEXUNITS			8
+
 typedef struct r_texunit_s {
-	_Bool enabled; // GL_TEXTURE_2D on / off
-	GLenum texture; // e.g. GL_TEXTURE0_ARB
+	_Bool enabled; // on / off (off uses null texture)
+	GLenum texture; // e.g. GL_TEXTURE0 + x
 	GLuint texnum; // e.g 123
+	GLuint bound; // the actual bound value regardless of enabled state
```
jdolan: Are both `texture` and `bound` actually used?

src/client/renderer/r_types.h
```diff
@@ -545,18 +570,16 @@ typedef struct r_model_s {
...
+	GLsizei num_elements; // number of vertex elements, if element_buffer is to be used
+	GLsizei num_tris; // cached num_tris amount
+
+	// vertex buffer objects
+	r_buffer_t *vertex_buffers;
+	r_buffer_t texcoord_buffer;
+	r_buffer_t lightmap_texcoord_buffer;
+	r_buffer_t *normal_buffers;
+	r_buffer_t *tangent_buffers;
+	r_buffer_t element_buffer;
```
jdolan: Why are some of these plural?

src/client/renderer/r_types.h
```diff
@@ -858,6 +895,7 @@ typedef struct {
 	uint32_t num_mesh_tris;
 
 	_Bool update; // inform the client of state changes
+	_Bool scissor_enabled;
```
jdolan: Isn't this duplicate from r_state_t?

src/client/renderer/r_types.h
```diff
@@ -841,6 +877,7 @@ typedef struct {
 
 	const r_entity_t *current_entity; // entity being rendered
 	const r_shadow_t *current_shadow; // shadow being rendered
+	vec4_t current_shadow_light, current_shadow_plane;
```
jdolan: Still necessary?

src/client/ui/ui_types.h
```diff
@@ -22,6 +22,9 @@
 #ifndef __UI_TYPES_H__
 #define __UI_TYPES_H__
 
+// This is just to silence error about gl.h already being included by Objectively.
+#define USE_SDL_GL
+
```
jdolan: Also not right imo. I'm happy to look into this.

src/common.c
```diff
@@ -134,12 +145,19 @@ void Com_Print(const char *fmt, ...) {
 		fputs(msg, stdout);
 		fflush(stdout);
 	}
+
+	// FIXME: not sure how to resolve this one, because
+	// prints happen before command buf is run and whatnot.
+	// developer may never be non-zero if crash happens very early.
```
jdolan: Stale comment?

src/cvar.h
```diff
@@ -49,6 +49,7 @@ _Bool Cvar_Command(void);
 char *Cvar_UserInfo(void);
 char *Cvar_ServerInfo(void);
 void Cvar_WriteAll(file_t *f);
+_Bool Cvar_EmplaceValues(const char *input, const size_t in_size, GString **output);
```
jdolan: If you're not thrilled with this name, I'd suggest Cvar_ExpandString. But up to you.

src/cgame/default/views/PlayerModelView.c
```diff
@@ -136,59 +71,56 @@ static void render(View *self, Renderer *renderer) {
...
+		Matrix4x4_ConcatRotate(&mat, cgi.client->systime * 0.08, 0.0, 0.0, 1.0);
 
-		glRotatef(cgi.client->systime * 0.08, 0.0, 0.0, 1.0);
-
-		glDepthRange(0.0, 0.1);
-		glEnable(GL_DEPTH_TEST);
-		glEnable(GL_TEXTURE_2D);
+		Matrix4x4_Copy(&cgi.view->active_matrices[R_MATRIX_MODELVIEW], &mat);
+		
+		cgi.EnableTextureID(0, true);
```
Paril: I couldn't find any, but we can easily introduce one.

src/client/renderer/Makefile.am
```diff
@@ -69,7 +71,9 @@ librenderer_la_SOURCES = \
 	r_model.c \
 	r_particle.c \
 	r_program.c \
+	r_program_corona.c \
```
Paril: That's something better suited to master methinks. It's not exactly related to GL2.

src/client/renderer/r_array.c
```diff
@@ -38,143 +38,128 @@ static r_array_state_t r_array_state;
...
-	// lightmap coords
-	if (texunit_lightmap.enabled) {
+	R_BindArray(R_ARRAY_COLOR, NULL);
 
-		if (mask & R_ARRAY_TEX_LIGHTMAP) {
-			R_SelectTexture(&texunit_lightmap);
+	_Bool do_interpolation = r_view.current_entity && IS_MESH_MODEL(mod) && mod->mesh->num_frames > 1 && r_view.current_entity->old_frame != r_view.current_entity->frame;
+	uint16_t old_frame = r_view.current_entity ? r_view.current_entity->old_frame : 0;
+	uint16_t frame = r_view.current_entity ? r_view.current_entity->frame : 0;
+
```
Paril: I had a crash there, I'll see about making that more readable

src/client/renderer/r_draw.c
Paril: That and it's always made more sense to me from a visibility perspective. 0.0115 r and 0.0114 r are not going to be distinguishable by 99% of the population.

src/client/renderer/r_main.c
Paril: It looked even uglier just printing the blank string, because of wrapping. I wouldn't mind going back to just printing the raw extension string, but there's a bug relating to colors on huge strings that I'd need to fix as well.

src/client/renderer/r_main.h
Paril: I followed the same concept as texunit_diffuse, texunit_lightmap, etc.

src/client/renderer/r_types.h
Paril: Because they hold more than one buffer (multiple frames), hence why they are pointers.

src/client/renderer/r_types.h
Paril: Those were added to resolve the todo in r_mesh_shadow, they store the state of the "light" and "plane" uniforms essentially. I'll see if I can find a way to keep these contained in the shadow program though.

src/client/renderer/r_types.h
Paril: There is no scissor enable bit in r_state_t, but you're right in that it should be there.

src/common.c
Paril: Kind of. It's moreso just noting that there's no way to print -everything- out to the attached debugger because the command buffer that enables debug happens several steps down the chain (if you remove the !debug check, for instance, it will crash).

PR #241, opened 2016-10-13
Smaller epsilon for points

tapir:


tapir:
@kaadmy could you test this with your maps. From my tests, it looks like it doesn't mess up anything but more eyes the better

jdolan:
Gehenna and Wake would be good test cases for this.

jdolan:
Did some testing with this. Massive improvements on Jester's crazy complex geometry.

@Panjoo can you think of any existing maps that have brush errors (cracks between faces, etc) that would benefit from a recompile?

Awesome find and fix @tapir.

jdolan:
Gehenna could definitely use a recompile. The self-jerking-off pipes in the main atrium have numerous small brush errors.

kaadmy:
What exactly does this fix?

tapir:
fixing missing narrow surfaces in jgothic and gevennah

PR #243, opened 2016-10-30
Remove part of the rocket trail

WickedShell:
The effect is really hard to see (at least for me), but it has the significant problem where the effect shoots through a wall a large distance. (Shoot a rocket against a near corner and keep moving around it quickly and you can see it shoot through).

### DIFF
--- a/src/cgame/default/cg_entity_trail.c
+++ b/src/cgame/default/cg_entity_trail.c
@@ -295,33 +295,6 @@ static void Cg_RocketTrail(cl_entity_t *ent, const vec3_t start, const vec3_t en
 
 	Cg_SmokeTrail(ent, start, end);
 
-	if (time < ent->time) { // time to add new particles
-		vec3_t delta;
-
-		VectorSubtract(end, start, delta);
-		const vec_t dist = VectorNormalize(delta);
-
-		vec_t d = 0.0;
-		while (d < dist) {
-			cg_particle_t *p;
-
-			if (!(p = Cg_AllocParticle(PARTICLE_NORMAL, cg_particles_flame)))
-				break;
-
-			cgi.ColorFromPalette(EFFECT_COLOR_ORANGE + (Random() & 5), p->part.color);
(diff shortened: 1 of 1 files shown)


Paril:
The other trail-like particles do this (blaster, grenade) - it's because the particles move forward by the direction the entity is going in, so once the blaster hits a wall the particles keep going inside of it.

I was thinking of a few different ways of addressing this. The easiest way would be to have the particles trace periodically and kill themselves if they touch a wall. I don't think removing the effect itself is a good solution, because there's still blaster and grenades that do this too.

WickedShell:
Ah, hadn't noticed it on the blaster one actually. Why would grenade have that small flame effect?

I think if the velocity was lowered and we instead increased the alpha/size velocities and then emitted them more commonly along the trail it would have much the same visual effect without the wall penetration. In this case though it was removing a particle effect that I can not observe/distinguish in the rockets trail, at least not from a first person perspective.

Paril:
I mean ,grenade smoke (and rocket smoke) also goes through walls. It's because they move in the same direction of the rocket/grenade/blaster, so when the blaster hits the wall the particles that spawned where the blaster hit keep moving forward.

WickedShell:
Upon further reflection it makes sense in principle for the blaster trail to shoot forward, but from a physics perspective that just doesn't make sense on projectiles, a rocket goes a direction because it shoots particles out the other end at high speed and then newton's third law kicks in... I haven't played with it yet but I imagine we could fix up all the rocket like cases to actually shoot particles the other way with a much shorter life span. Blaster we could may just play with the particle life times so that a shot doesn't continue through the wall?

At a high level I'm okay with the fact that particles will make it through walls, what I'm not okay with is the distance that this effect over penetrates. It's at least an entire room on Torn as well as other small maps, and can be really surprising a wall over in Lilith. We just need faster decays so you don't see it shoot through like that.

jdolan:
I quite like your explanation and suggestion. If you wanna take a crack at tuning this effect so that the particles either shoot "backwards" from the rocket, or even just hang in place in order to create the correct effect, I'm down with that.

PR #246, opened 2016-10-31
Feature/particles

Paril:
This is a fairly large PR, but here's the skinny.

Some cleanup of the particle system, gives more room to add more types. Honestly I think that the cgame should only use its own type for all of it and only setup r_particle_t to send to the renderer instead of storing r_particle_t as part of the particle and building it as you go. It would also simplify some other code, but r_element_t still needs a pointer to the particle which is a little annoying. Need to think more about that connection.

Frustum calculation was wrong, so I copied and cleaned up the code from Quake 3 as it seems to work properly.

Cgame now has the ability to do culling, which helps in particle intensive scenes. It'd be cool if this could be done on another thread somehow, but for now it just checks culling after it updates and pushes if it's on screen. This cuts down visible particles in most areas on Fractures by like 60%.

Element sorting is done in two passes now - first pass sorts by depth, second pass is done solely on particles and sorts them by type and image. This keeps state changes down.

A system for creating atlases has been introduced. It works, and can be used to atlas together any textures you'd like. It creates a media entry for itself (with all of the images in the atlas as dependencies) and combines all of the images into one giant texture (with the max mip size being the smallest image available). It creates a new, local-only media entry for each sub-image of the atlas (of type IT_ATLAS_IMAGE) which contains the texture coordinates of the sub-image. There's still some improvements that need to be made to the system, but they are not required for our current implementation:
* Better stitching algorithm. Right now it just packs them on X axis and ignores Y completely.
* Be sure that the mipping works on most hardware.
* Functions for removing images from atlases, dirty checks and see if you can recompile at runtime without many issues
* Cache atlases locally? This would be a nice addition, and could be done with a bit of metadata to cache atlases locally and not recompile them (and just load the mips from compiled TGAs) if the mod didn't add/remove anything new from the files. Not sure if this is something @jdolan would be interested in though since we have nothing else cached in the game.

Lightning beams now come from the gun in first person. Crouch check was unnecessary since view org is view org, not client org. Didn't adjust third person code, but I think we should have the beam come from the barrel tag on the client-side (start position is unnecessary to convey where they're aiming, so it wouldn't affect gameplay). If/when we have the capability to do that from the cgame (query tags for positions, etc) then this beam system should be changed to use that as well even in first person.

Bullets now have a spark-like particle.

Coronas are now particles, batched together and sorted by depth as well.

Removed "end" being pushed for beam particles. This was only used for BFG lasers and bullet trails, but it caused them to go inside walls which looked funky. It looks fine without it.


jdolan:
General impressions: great stuff. I'm sure the atlas will help in busy scenes, and I love the new bullet sparks. We should also use that particle type for TE_SPARKS / cg_emit sparks.

src/cgame/cgame.h
```diff
@@ -472,6 +472,28 @@ typedef struct cg_import_s {
...
+	 *
+	 * @param point The central point of the sphere.
+	 * @param radius The radius of the sphere.
+	 *
+	 * @return Returns true if the specified sphere is completely culled by the
+	 * view frustum, false otherwise.
+	 */
+	_Bool (*CullSphere)(const vec3_t point, const float radius);
+
+	/**
```
jdolan: 👍

src/cgame/cgame.h
```diff
@@ -499,6 +521,44 @@ typedef struct cg_import_s {
...
+	
+	/**
+	 * @brief Resolve an atlas image from an atlas and image.
+	 *
+	 * @param atlas The atlas to fetch the stitched image from.
+	 * @param image The original image you wish to query.
+	 *
+	 * @return An r_atlas_image_t which contains the stitched coordinates of that image.
+	 */
+	const r_atlas_image_t *(*GetAtlasImageFromAtlas)(const r_atlas_t *atlas, const r_image_t *image);
```
jdolan: GetImageFromAtlas is probably adequate. EDIT: Oh, I see that it returns `r_atlas_image_t`.

src/cgame/cgame.h
```diff
@@ -499,6 +521,44 @@ typedef struct cg_import_s {
...
+
+	/**
+	 * @brief Stitches the images together into an atlas.
+	 *
+	 * @param atlas The atlas to stitch together.
+	 */
+	void (*StitchAtlas)(r_atlas_t *atlas);
+
```
jdolan: Stitch is kind of a weird word. CompileAtlas or FinalizeAtlas or something maybe?

src/cgame/cgame.h
```diff
@@ -499,6 +521,44 @@ typedef struct cg_import_s {
...
```
jdolan: In general, this is awesome.

src/cgame/cgame.h
```diff
@@ -524,11 +584,6 @@ typedef struct cg_import_s {
 	r_model_t *(*WorldModel)(void);
 
 	/**
-	 * @brief Adds a corona to the scene for the current frame.
-	 */
-	void (*AddCorona)(const r_corona_t *c);
-
-	/**
```
jdolan: CYA

src/cgame/default/cg_entity_trail.c
```diff
@@ -646,20 +636,16 @@ void Cg_EntityTrail(cl_entity_t *ent, r_entity_t *e) {
...
 					break;
 			}
 
-			if (cgi.client->frame.ps.pm_state.flags & PMF_DUCKED) {
-				VectorMA(start, -6.0, cgi.view->up, start);
-			} else {
-				VectorMA(start, -12.0, cgi.view->up, start);
-			}
+			VectorMA(start, -8.0, cgi.view->up, start);
 		}
```
jdolan: Seems to work!

src/cgame/default/cg_particle.c
```diff
@@ -184,21 +241,50 @@ void Cg_AddParticles(void) {
...
+
+				if (free) {
 					p = Cg_FreeParticle(p, &ps->particles);
 					continue;
 				}
 			}
 
-			cgi.AddParticle(&p->part);
+			// add the particle if it's visible on our screen
+			float radius;
```
jdolan: vec_t, and could be moved into each block and made const.

src/client/renderer/r_atlas.c
```diff
@@ -0,0 +1,237 @@
...
+ */
+void R_AddImageToAtlas(r_atlas_t *atlas, const r_image_t *image) {
+
+	// ignore duplicates
+	if (g_hash_table_contains(atlas->hash, image))
+		return;
+
+	// add to list
+	g_array_append_val(atlas->images, (const r_atlas_image_t){
+		.input_image = image
```
jdolan: Whoa, is this safe? Appending a stack variable to a GList?

src/client/renderer/r_atlas.c
```diff
@@ -0,0 +1,237 @@
...
+
+		SDL_LockSurface(mip_surface);
+
+		R_BindTexture(atlas->image.texnum);
+
+		glTexImage2D(GL_TEXTURE_2D, i, GL_RGBA, mip_width, mip_height, 0, GL_RGBA, GL_UNSIGNED_BYTE, mip_surface->pixels);
+
+		SDL_UnlockSurface(mip_surface);
+
+		SDL_FreeSurface(mip_surface);
```
jdolan: Damn. I'm slightly surprised that this much work is necessary here. Just curious: why can't we simple rely on glGenerateMipmap?

src/client/renderer/r_atlas.c
```diff
@@ -0,0 +1,237 @@
...
+	
+	// check for errors
+	R_GetError(atlas->image.media.name);
+
+	// setup texcoords
+	for (uint16_t i = 0, j = 0; i < atlas->images->len; i++) {
+		r_atlas_image_t *image = &g_array_index(atlas->images, r_atlas_image_t, i);
+
+		Vector4Set(image->texcoords, 
+			(float)j / (float)width,
```
jdolan: vec_t

src/client/renderer/r_particle.c
```diff
@@ -293,12 +312,22 @@ void R_DrawParticles(const r_element_t *e, const size_t count) {
 	R_BindArray(R_ARRAY_ELEMENTS, &r_particle_state.element_buffer);
 
 	const GLuint base = (uintptr_t) e->data;
+	r_particle_type_t last_type = -1;
+	GLuint last_texnum = -1;
```
jdolan: 0?

src/quetoo.h
```diff
@@ -73,6 +73,11 @@ typedef dvec_t dvec2_t[2];
 typedef dvec_t dvec3_t[3];
 typedef dvec_t dvec4_t[4];
 
+typedef uint16_t u16vec_t;
+typedef u16vec_t u16vec2_t[2];
+typedef u16vec_t u16vec3_t[3];
+typedef u16vec_t u16vec4_t[4];
+
```
jdolan: Hm. Cool. Probably a few places we could use this.

src/client/renderer/r_atlas.c
Paril: It's a g_array, it copies the data from the input.

src/client/renderer/r_atlas.c
Paril: You can't generate mipmaps of atlases. If you have 4 separate textures together in an image, scaling them down will cause texels to leak into the surrounding texels. The only way to guarantee that the mips are correct is to generate them by hand.

src/client/renderer/r_particle.c
Paril: Type 0 is a valid type, same same with last_texnum (particle with no image may be valid, like coronas)

src/cgame/cgame.h
Paril: Yeah I wanted to separate "regular image" from "atlas image". The atlas image is a special media entry owned by the atlas that just points to where "image" is on the actual "atlas" while still letting it be passed around as an actual image.

src/cgame/cgame.h
Paril: Can do. In the texturing world, stitching is the term used for these kinds of things. Initially it was "Upload" but that didn't seem right either. Compile works for me.

src/cgame/default/cg_particle.c
Paril: Will do.

jdolan:
:shipit:

PR #258, opened 2016-11-11
Feature/gl types

Paril:


src/client/renderer/r_state.h
```diff
@@ -67,28 +67,14 @@ typedef struct r_matrix_stack_s {
...
-
-	// built-in buffers for the above vertex arrays
-	r_buffer_t buffer_vertex_array;
-	r_buffer_t buffer_color_array;
-	r_buffer_t buffer_normal_array;
-	r_buffer_t buffer_tangent_array;
-	r_buffer_t buffer_element_array;
-
-	r_buffer_t buffer_interleave_array;
-
```
jdolan: 👍

src/client/renderer/r_types.h
```diff
@@ -94,6 +94,99 @@ typedef enum {
...
+	R_ATTRIB_SHORT,
+	R_ATTRIB_UNSIGNED_SHORT,
+	R_ATTRIB_INT,
+	R_ATTRIB_UNSIGNED_INT,
+	R_ATTRIB_INT_2_10_10_10_REV,
+
+	R_ATTRIB_TOTAL_TYPES
+} r_attrib_type_t;
+
+// NOTE: if any of the enums or data types
```
jdolan: I'd rather see Doxygen `@brief` `@details` and `@remarks` comments for stuff like this, and 100 char line wrap. Nothing to hold up the pull request for. Just a standard I'd like to move to for block-level comments.

src/client/renderer/r_types.h
```diff
@@ -275,12 +374,11 @@ typedef struct {
...
 typedef struct {
-	vec3_t origin;
 	vec_t radius;
-	const r_image_t *image;
-	vec3_t color;
 	uint32_t time;
 	vec_t alpha;
+
+	r_particle_t particle;
 } r_bsp_flare_t;
```
jdolan: Cool.

jdolan: Is it possible to remove the flare's alpha and timestamps, and just use those of the backing particle?

src/game/default/g_client_view.c
```diff
@@ -185,109 +185,106 @@ static void G_ClientWorldAngles(g_entity_t *ent) {
...
+	const vec_t frac = (g_level.time - ent->client->locals.kick_angles_time) / 64.0;
+	if (frac < 1.0) {
+		AngleLerp(kick, ent->client->locals.kick_angles, Clamp(frac, 0.0, 1.0), kick);
+	} else {
+		// otherwise, decay the kick, and use the resulting value directly
+		G_ClientDecayKick(ent);
+		VectorCopy(ent->client->locals.kick_angles, kick);
 	}
 
 	PackAngles(kick, kick_angles);
```
jdolan: Well this is fucking weird. Why is this showing up here? I pushed this to `master` already.

src/quetoo.h
```diff
@@ -63,20 +63,40 @@
...
+#define VECTOR_TYPENAME_N(name, n) name ## n ## _t[n]
+#define VECTOR_TYPE(type, typename) \
+	typedef type VECTOR_TYPENAME(typename); \
+	typedef VECTOR_TYPENAME(typename) VECTOR_TYPENAME_N(typename, 2); \
+	typedef VECTOR_TYPENAME(typename) VECTOR_TYPENAME_N(typename, 3); \
+	typedef VECTOR_TYPENAME(typename) VECTOR_TYPENAME_N(typename, 4);
+
+VECTOR_TYPE(float, vec);
+VECTOR_TYPE(double, dvec);
+VECTOR_TYPE(_Bool, bvec);
```
jdolan: Do we use bvec anywhere? If not, I'd recommend we lose this so that we don't slip up and think it's 'byte vector'. Just my $0.02. Otherwise, this is nice.

src/client/renderer/r_state.c
```diff
@@ -1101,39 +1174,17 @@ void R_InitState(void) {
...
-
-			texunit->texcoord_array = Mem_TagMalloc(len, MEM_TAG_RENDERER);
-			R_CreateBuffer(&texunit->buffer_texcoord_array, GL_DYNAMIC_DRAW, R_BUFFER_DATA, len, NULL);
-
-			R_EnableTexture(texunit, true);
-
-			R_BindDefaultArray(R_ARRAY_TEX_DIFFUSE);
-
-			R_EnableTexture(texunit, false);
 		}
```
jdolan: Sick!

src/client/renderer/r_state.c
```diff
@@ -917,6 +985,16 @@ void R_UseInterpolation(const vec_t lerp) {
...
 /**
+ * @brief Uploads shell offset value to the currently loaded program.
+ */
+void R_UseShellOffset(const vec_t offset) {
+
+	if (r_state.active_program->UseShellOffset) {
+		r_state.active_program->UseShellOffset(offset);
+	}
+}
+
```
jdolan: Feels a little gross to expose this here, when only one program will ever use this. Is there a way for the shell program's generic Use function to check the current entity and its flags, and modify the uniform accordingly?

jdolan: Not a deal breaker, btw.

src/client/renderer/r_state.c
```diff
@@ -434,13 +459,52 @@ void R_CreateBuffer(r_buffer_t *buffer, const GLenum hint, const r_buffer_type_t
...
 }
 
 /**
  * @brief Destroy a GPU-allocated buffer.
  */
 void R_DestroyBuffer(r_buffer_t *buffer) {
-	assert(buffer->bufnum != 0);
+
+	if (buffer->bufnum == 0) {
+		Com_Warn("Attempting to free empty buffer");
```
jdolan: A Com_Debug would be fine here. This is a development bug trap, and I typically only use Warn for things a user might see under normal operation.

src/client/renderer/r_state.c
```diff
@@ -391,41 +358,99 @@ void R_UploadToSubBuffer(r_buffer_t *buffer, const size_t start, const size_t si
...
  */
-void R_CreateBuffer(r_buffer_t *buffer, const GLenum hint, const r_buffer_type_t type, const size_t size,
+void R_CreateBuffer(r_buffer_t *buffer, const r_attrib_type_t element_type, const GLubyte element_count,
+                    const _Bool element_normalized, const GLenum hint,
+                    const r_buffer_type_t type, const size_t size,
                     const void *data) {
 
-	assert(buffer->bufnum == 0);
+	if (buffer->bufnum != 0) {
+		Com_Warn("Attempting to reclaim non-empty buffer");
```
jdolan: This one, too, is not something users should typically see. Making it a Com_Debug makes sense imo.

src/client/renderer/r_state.c
```diff
@@ -391,41 +358,99 @@ void R_UploadToSubBuffer(r_buffer_t *buffer, const size_t start, const size_t si
...
 
 /**
+ * @brief
+ */
+static GLubyte R_GetElementSize(const GLenum type) {
+
+	switch (type) {
+		case R_ATTRIB_BYTE:
+		case R_ATTRIB_UNSIGNED_BYTE:
+		case R_ATTRIB_INT_2_10_10_10_REV:
```
jdolan: Wait, isn't this 4?

src/client/renderer/r_state.c
```diff
@@ -25,11 +25,11 @@ static cvar_t *r_get_error;
...
-	0.0, 0.0,
-	1.0, 0.0,
-	1.0, 1.0,
-	0.0, 1.0
+const vec2_t default_texcoords[4] = { // useful for particles, pics, etc..
+	{ 0.0, 0.0 },
+	{ 1.0, 0.0 },
+	{ 1.0, 1.0 },
+	{ 0.0, 1.0 }
 };
```
jdolan: 👍  ;)

src/client/renderer/r_state.c
Paril: I was grossed out by it too, and I was trying to think of a way to directly access the state of the shell program without causing too much trouble but I just couldn't do it elegantly. My first thought was just as yours, to hijack the Use function, but Use is called wayyyy before entities get managed and we can't call Use for each entity because it will skip it (it won't try to Use a program that is already Use'd) and seems kind of inefficient.

src/client/renderer/r_state.c
Paril: Fixing.

src/client/renderer/r_state.c
Paril: Fixing.

src/client/renderer/r_state.c
Paril: I'm adding a note to this, but basically it's something of a hack. GL requires that INT_2_10_10_10_REV is defined as 4 bytes wide but also have a count of 4, even though the actual component is 4 bytes, not (4 * 4) bytes. Instead of having a completely separate if/else just to fix that one specific case where the GL spec is stupid as shit, I tell it that it's 1 byte so that the stride/size is correctly calculated as 4 bytes wide.

src/quetoo.h
Paril: Yeah I'm removing the ones we don't use.

src/game/default/g_client_view.c
Paril: My fault, I squashed when I shouldn't have.

src/client/renderer/r_types.h
Paril: Will do. I wish we could have astyle wrap comments by 100 automatically instead of the 120.

src/client/renderer/r_types.h
jdolan: Oh, I see. I thought the `alpha` field here was the final, "ramped" alpha after the visibility check. You already moved that into the particle. Great.

src/client/renderer/r_state.c
jdolan: Is there some way to "overload" glPolygonOffset for this, maybe? And if not, really, don't worry about it.

Issue #237, opened 2016-10-07
Bug fix / potential optimization for r_shadows 2


jdolan:
The planar shadows make extensive use of the stencil buffer. Each surface is rendered with a stencil value derived from its plane number. When shadows impact a plane, we draw that shadow with the stencil test for that plane number. The stencil buffer is only 8 bits, which leaves a maximum number of unique stencil values of 255. In fact, I think the way this worked out, I was only able to use 127 unique values:

```
    glStencilFunc(GL_ALWAYS, (surf->plane->num % 0xff) + 1, ~0);
```

This results in shadows sometimes hitting a false positive for surfaces (planes) across the map. If you've ever seen a stray, nonsensical shadow, this is why. Exiting the water on Edge is an easy way to see this.

So.. what if we flipped a reserved bit on `surf->flags` whenever a shadow impacts that plane, and then check that reserved bit before applying glStencilFunc? For surfaces that are not impacted by a shadow, we would write `0` to the stencil buffer (or we could even disable the stencil buffer, if that yields a performance boost). We would have to add a simple loop in R_MarkBspSurfaces to reset this flag every frame.

This should fix the stray shadows, and maybe even provide a small framerate boost.

jdolan:
The tricky part here is, in the shadow code, when a trace yields a shadow, you must mark the _renderer's_ copy of the impacted plane, so that the renderer can easily check the plane when drawing surfaces.

We currently do not have a `r_bsp_plane_t` type, but I would add one, that is essentially a `cm_bsp_plane_t` with a bit mask field for renderer state / flags such as this.

jdolan:
@Paril handing this one back to you. Now that the fix itself is in there, if you wanted to try optimizing stencil state management, be my guest. And if not, we can close this out.

Paril:
Old profile: http://i.imgur.com/UHpi2WB.png
New profile: http://i.imgur.com/4G4DCD5.png

R_Stencil ops are barely noticable now. I also get about 15-20 FPS improvement on wake. Nice!

PR #260, opened 2016-11-13
#237 Flip a bit on shadow-impacted frames, use this to govern stencil…

jdolan:
This addresses a limitation of an 8 bit stencil buffer, and using the stencil test to apply shadows to planes in the map. Because of the 8 bit limitation, every 256 planes share the same stencil mask. This can cause shadows to "jump" to seemingly random planes across the map, as the stencil test will yield a positive result.

A new type was introduced for renderer planes (`r_bsp_plane_t`), which extends `cm_bsp_plane_t` to include a bit mask of renderer-specific flags. A flag indicating that a mesh shadow impacted a plane is flipped in `R_CastShadows`. The flag is consulted as surfaces are drawn: if set, the plane number is written to the stencil buffer as before; if not set, `0` is written so that the subsequent stencil tests will always fail.

This should (and seems to) prevent shadows from jumping across the map. I have not yet explored optimizing the stencil state management, so that perhaps the stencil write is disable by default, and only enabled when we're dealing with a shadowed plane. @Paril if you wanted to experiment there.. you have better profiling tools than I do.

Also, I'll concede that the "index" trick to align collision and renderer planes is a bit ugly, but it was simple and efficient. Because the collision model trace structure (`cm_trace_t`) works with _copies_ of planes, and not pointers to them, simple pointer arithmetic would not work -- and so the `index` field was necessary.

### DIFF
--- a/src/client/renderer/r_bsp.c
+++ b/src/client/renderer/r_bsp.c
@@ -156,7 +156,7 @@ static void R_DrawBspInlineModel_(const r_entity_t *e) {
 	surf = &r_model_state.world->bsp->surfaces[e->model->bsp_inline->first_surface];
 
 	for (i = 0; i < e->model->bsp_inline->num_surfaces; i++, surf++) {
-		const cm_bsp_plane_t *plane = surf->plane;
+		const r_bsp_plane_t *plane = surf->plane;
 		vec_t dot;
 
 		// find which side of the surf we are on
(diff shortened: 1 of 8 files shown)


Paril:
I think my only gripe is that the new structure could contain a cm_bsp_plane_t instead of re-declaring the 5 thingies, but other than that it's great.

jdolan:
Yea, I definitely thought about that, but the thought of `plane->plane` made me decide against it. I dunno, tho. I could go either way, really.

Paril:
((cm_bsp_plane_t *) plane)->normal :p

PR #263, opened 2016-11-14
Issue/262 shadows flicker

jdolan:
@Paril I have to get busy with work-work, but I could use your help on this one.

The change refactors r_bsp_plane_t to include a cm_bsp_plane_t, like we discussed previously. So there's some "noise" in the PR about that. But the "fix" here to eliminate the shadow flickering is to change how the "plane was impacted by a shadow" state is managed per frame.

In the initial commit for #237, I flipped the R_PLANE_SHADOW bit off in R_MarkBspSurfaces_. However, static lighting is cached over multiple frames. So I was effectively discarding the shadow each frame, and it would not be cast again until the entity's lighting cache was marked dirty.

The fix in this commit is to instead keep a count of shadows on r_bsp_plane_t, and decrement that count when recalculating shadows. This all seems to work correctly; however, there is a crash that I couldn't get my head around this morning.

Spawn into Edge, take the hyperblaster, and go fire it at a wall near some mesh entities. You'll get a crash in R_FreeShadow. It will be for the torso entity of your player model -- which shares a lighting struct with the other player model segments. Still, I don't understand the crash. HALP PLZ.

### DIFF
--- a/src/client/renderer/r_lighting.c
+++ b/src/client/renderer/r_lighting.c
@@ -367,7 +367,7 @@ static void R_CastShadows(r_lighting_t *l, const r_illumination_t *il) {
 		r_shadow_t *s = l->shadows;
 		while (s->illumination) {
 
-			if (s->illumination == il && s->plane.num == tr.plane.num) {
+			if (s->illumination == il && s->plane->num == tr.plane.num) {
 				s->shadow = MAX(s->shadow, shadow);
 				break;
 			}
@@ -385,7 +385,8 @@ static void R_CastShadows(r_lighting_t *l, const r_illumination_t *il) {
 		}
 
 		s->illumination = il;
-		s->plane = tr.plane;
+		s->plane = &r_model_state.world->bsp->planes[tr.plane.index];
+		s->plane->flags |= R_PLANE_SHADOW;
 		s->shadow = shadow;
 	}
 }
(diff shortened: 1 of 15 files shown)


src/client/renderer/r_lighting.c
```diff
@@ -386,12 +386,26 @@ static void R_CastShadows(r_lighting_t *l, const r_illumination_t *il) {
...
 }
 
 /**
+ * @brief Frees the specified shadow, decrementing its plane's shadow count.
+ */
+static void R_FreeShadow(r_shadow_t *s) {
+
+	if (s->plane) {
+		if (s->plane->num_shadows) {
+			s->plane->num_shadows--;
```
jdolan: And here is where she'll crash now, on the torso segment of your player model. Why?

src/client/renderer/r_lighting.c
```diff
@@ -409,7 +423,9 @@ static void R_UpdateShadows(r_lighting_t *l) {
 	}
 
 	// otherwise, refresh all shadow information based on the new illuminations
-	memset(l->shadows, 0, sizeof(l->shadows));
+	for (size_t i = 0; i < lengthof(l->shadows); i++) {
+		R_FreeShadow(&l->shadows[i]);
+	}
```
jdolan: Instead of blindly wiping all shadows, we iterate them, and decrement the shadow count of any impacted planes. This should be safe. I can't conceive of how a shadow could hold an invalid reference to a plane -- at least not on the first map load. I don't get it.

PR #267, opened 2016-11-15
Feature/lightmap atlas

Paril:
I was told to, and I quote, "fuck it".

### DIFF
--- a/src/client/renderer/r_atlas.c
+++ b/src/client/renderer/r_atlas.c
@@ -96,38 +96,60 @@ const r_atlas_image_t *R_GetAtlasImageFromAtlas(const r_atlas_t *atlas, const r_
 	return (const r_atlas_image_t *) g_hash_table_lookup(atlas->hash, image);
 }
 
-typedef struct {
-	uint16_t width, height;
-	uint16_t x, y;
+/**
+ * @brief Initialize an r_packer_t structure. If the packer is already
+ * created, clears the packer back to an initial state.
+ */
+void R_AtlasPacker_InitPacker(r_packer_t *packer, const r_pixel_t max_width, const r_pixel_t max_height, const r_pixel_t root_width, const r_pixel_t root_height, const uint32_t initial_size) {
 
-	uint32_t right, down; // nodes to the right/below this one
-	_Bool used; // whether this node is used by an image or not yet
-} r_packer_node_t;
+	packer->max_width = max_width;
+	packer->max_height = max_height;
 
-typedef struct {
-	GArray *nodes;
-	uint32_t root;
-} r_packer_t;
+	if (packer->nodes == NULL) {
+		packer->nodes = g_array_sized_new(false, false, sizeof(r_packer_node_t), initial_size + 1);
+	} else {
+		packer->nodes->len = 0;
+	}
+
+	g_array_append_vals(packer->nodes, &(const r_packer_node_t) {
+		.width = root_width,
+		 .height = root_height,
+		  .right = -1,
+		   .down = -1
+	}, 1);
+
+	packer->root = 0;
+}
+
+/**
+ * @brief Free data created by R_AtlasPacker_InitPacker
+ */
+void R_AtlasPacker_FreePacker(r_packer_t *packer) {
+
+	g_array_free(packer->nodes, true);
+	packer->nodes = NULL;
+}
 
 /**
  * @brief Finds a free node that is big enough to hold us.
  */
-static r_packer_node_t *R_AtlasPacker_FindNode(r_packer_t *packer, r_packer_node_t *root, r_atlas_image_t *image) {
+r_packer_node_t *R_AtlasPacker_FindNode(r_packer_t *packer, r_packer_node_t *root, const r_pixel_t width, const r_pixel_t height) {
 
 	if (root->used) {
 		r_packer_node_t *node = R_AtlasPacker_FindNode(packer, &g_array_index(packer->nodes, r_packer_node_t, root->right),
-		                        image);
+		                        width, height);
 
 		if (node) {
 			return node;
 		}
 
-		node = R_AtlasPacker_FindNode(packer, &g_array_index(packer->nodes, r_packer_node_t, root->down), image);
+		node = R_AtlasPacker_FindNode(packer, &g_array_index(packer->nodes, r_packer_node_t, root->down), width, height);
 
 		if (node) {
 			return node;
 		}
-	} else if (image->input_image->width <= root->width && image->input_image->height <= root->height) {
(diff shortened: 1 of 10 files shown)


src/client/renderer/r_atlas.c
```diff
@@ -96,38 +96,60 @@ const r_atlas_image_t *R_GetAtlasImageFromAtlas(const r_atlas_t *atlas, const r_
...
 }
 
-typedef struct {
-	uint16_t width, height;
-	uint16_t x, y;
+/**
+ * @brief Initialize an r_packer_t structure. If the packer is already
+ * created, clears the packer back to an initial state.
+ */
+void R_AtlasPacker_InitPacker(r_packer_t *packer, const r_pixel_t max_width, const r_pixel_t max_height, const r_pixel_t root_width, const r_pixel_t root_height, const uint32_t initial_size) {
```
jdolan: Astyle (line length).

jdolan: Also, I typically don't bother with `const` for non-pointer arguments. But I guess it can't hurt.

src/client/renderer/r_lightmap.c
```diff
@@ -62,74 +56,23 @@ static r_image_t *R_AllocLightmap_(r_image_type_t type) {
...
-#define R_AllocLightmap() R_AllocLightmap_(IT_LIGHTMAP)
-#define R_AllocDeluxemap() R_AllocLightmap_(IT_DELUXEMAP)
-
-/**
- * @brief
- */
-static void R_UploadLightmapBlock(r_bsp_model_t *bsp) {
-
-	R_UploadImage(r_lightmap_state.lightmap, GL_RGB, r_lightmap_state.sample_buffer);
-	R_UploadImage(r_lightmap_state.deluxemap, GL_RGB, r_lightmap_state.direction_buffer);
```
jdolan: Do we lose the ability to upload lightmaps as GL_RGB and not GL_RGBA with this change?

src/client/renderer/r_lightmap.c
```diff
@@ -235,86 +174,160 @@ static void R_BuildLightmap(const r_bsp_model_t *bsp, const r_bsp_surface_t *sur
...
+		
+		surf->lightmap = lightmap;
+		surf->deluxemap = deluxemap;
+
+		start = start->next;
+	} while (start != end);
+
+	// upload!
+	R_UploadImage(lightmap, GL_RGB, sample_buffer);
+	R_UploadImage(deluxemap, GL_RGB, direction_buffer);
```
jdolan: Ah! Yes, we do ;)

src/client/renderer/r_lightmap.c
```diff
@@ -235,86 +174,160 @@ static void R_BuildLightmap(const r_bsp_model_t *bsp, const r_bsp_surface_t *sur
...
+			}
+		} while (node == NULL);
+	}
+
+	// free stuff
+	R_AtlasPacker_FreePacker(&packer);
+
+	g_slist_free(r_lightmap_state.blocks);
+	r_lightmap_state.blocks = NULL;
+}
```
jdolan: Beautiful!

src/client/renderer/r_main.c
```diff
@@ -479,8 +478,6 @@ static void R_InitLocal(void) {
 	r_height = Cvar_Add("r_height", "0", CVAR_ARCHIVE | CVAR_R_CONTEXT, NULL);
 	r_invert = Cvar_Add("r_invert", "0", CVAR_ARCHIVE | CVAR_R_MEDIA,
 	                    "Inverts the RGB values of all world textures");
-	r_lightmap_block_size = Cvar_Add("r_lightmap_block_size", "4096", CVAR_ARCHIVE | CVAR_R_MEDIA,
-	                                 NULL);
```
jdolan: Yay, no more guessing! Do we need a max, tho? For maps like Wake (again, it's an outlier, but still..), does the engine perform better with 2x 2048 lightmaps, or 1x 4096 lightmap?

src/client/renderer/r_mesh_model.c
```diff
@@ -875,7 +875,7 @@ static void R_LoadObjShellVertexArrays(r_model_t *mod, r_obj_t *obj, GLuint *ele
 	}
 	
 	GHashTable *index_remap_table = g_hash_table_new(g_direct_hash, g_direct_equal);
-	GArray *unique_vertex_list = g_array_new(true, false, sizeof(r_obj_shell_interleave_vertex_t));
+	GArray *unique_vertex_list = g_array_new(false, false, sizeof(r_obj_shell_interleave_vertex_t));
```
jdolan: This fix a crash or something?

src/client/renderer/r_types.h
```diff
@@ -452,7 +452,7 @@ typedef struct {
 	vec2_t st_mins;
 	vec2_t st_maxs;
 	vec2_t st_center;
-	vec2_t st_extents;
+	s16vec2_t st_extents;
```
jdolan: Oh nice, saves some mem.

src/client/renderer/r_types.h
```diff
@@ -462,7 +462,11 @@ typedef struct {
 	r_bsp_flare_t *flare;
 
 	r_pixel_t lightmap_s, lightmap_t; // lightmap texture coords
-	r_image_t *lightmap;
+
+	union {
+		r_image_t *lightmap;
+		const byte *lightmap_input; // this is only used by r_lightmap
+	};
```
jdolan: Then why doesn't r_lightmap.c just cast?

src/client/renderer/r_atlas.c
Paril: Sorry, it's a thing I can't unlearn from C++. I had const correctness drilled into my head.

src/client/renderer/r_lightmap.c
Paril: No sir, they are still GL_RGB.

src/client/renderer/r_mesh_model.c
Paril: It allocated an extra vertex, no biggie.

src/client/renderer/r_types.h
Paril: I could have done that but it seemed weird to hijack an r_image_t pointer just to store a temporary byte pointer for a single step. Do you think the cast would be less disconcerting?

jdolan: Oh, I see (I think). This is the pointer to the sample offset in the lightmap lump? I would either go ahead and waste the pointer on it (because we're going to require the original lightmap samples for stainmaps -- potentially), or add a helper function that returns this pointer for a given surface. Make sense?

PR #271, opened 2016-11-17
Shadow plane limiting with reduced memory usage and fixed threading issue.

jdolan:
Some noise here due to Objectively documentation updates, but the real change here is that an array of shadow counts is now allocated on `r_bsp_model_t`, and is of the smallest possible length. The previous implementation here allocated a `uint16_t` for each plane, which is wasteful.

The other change here is to remove the parallelization of `R_CullEntities`. This function is responsible for updating lighting information (i.e. casting shadows). So entities need to be fully culled before we can draw world surfaces correctly. Otherwise, it's quite likely that shadows will appear a frame late, as the check against `plane_shadows` will fail until the shadows are actually cast.

### DIFF
--- a/src/cgame/default/viewcontrollers/CreateServerViewController.c
+++ b/src/cgame/default/viewcontrollers/CreateServerViewController.c
@@ -24,7 +24,6 @@
 #include "cg_local.h"
 
 #include "CreateServerViewController.h"
-#include "MapListCollectionItemView.h"
 
 #define _Class _CreateServerViewController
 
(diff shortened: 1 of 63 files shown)


src/cgame/default/viewcontrollers/CreateServerViewController.c
```diff
@@ -67,36 +67,36 @@ static void selectTeamsplay(Select *select, Option *option) {
...
+//	g_list_free(selectedMaps);
+//
+//	if (firstmap[0] == '\0') {
+//		cgi.Warn("No map selected\n");
+//
+//		return;
+//	}
+//
+//	cgi.CvarSet("g_map_rotation", maplist);
+//	cgi.Cbuf(va("map %s\n", firstmap));
```
jdolan: Will fix this up tonight.

src/client/renderer/r_bsp_model.c
```diff
@@ -667,11 +667,11 @@ static void R_LoadBspPlanes(r_bsp_model_t *bsp, const d_bsp_lump_t *l) {
...
 		out->dist = LittleFloat(in->dist);
 		out->type = LittleLong(in->type);
-		out->sign_bits = Cm_SignBitsForPlane(out);
+		out->sign_bits = Cm_SignBitsForPlane((cm_bsp_plane_t *) out);
 		out->num = (i >> 1) + 1;
 	}
 
-	r_shadow_state.plane_shadow_counts = Mem_LinkMalloc(((count >> 1) + 1) * sizeof(int16_t), bsp);
+	bsp->plane_shadows = Mem_LinkMalloc(((count >> 1) + 1) * sizeof(uint16_t), bsp);
 }
```
jdolan: This can live inside of the BSP again, now that the memory corruption has been tracked down and eliminated.

src/client/renderer/r_main.c
```diff
@@ -162,15 +162,14 @@ void R_DrawView(void) {
 
 	R_AddSustainedLights();
 
-	// dispatch threads to cull entities and sort elements while we draw the world
-	thread_t *cull_entities = Thread_Create(R_CullEntities, NULL);
-
 	R_AddFlares();
 
-	thread_t *sort_elements = Thread_Create(R_SortElements, NULL);
+	R_CullEntities();
```
jdolan: We have to cull entities before we can draw the world :/

PR #286, opened 2016-12-10
Ensure every server update is fully processed by the client.

jdolan:
This change is to fix the scenario where the client receives multiple updates from the server in a single client frame iteration (Cl_ReadPackets). This scenario is increasingly common at high server tick rates, but it's entirely possible even at 30Hz, depending on client framerate and network connection quality. A temporary dip in client framerate, or a jittery network connection, can both produce this scenario even at 30Hz, and so we should deal with it.

The solution is to call Cl_Interpolate from within Cl_ParseFrame. Cl_Interpolate has been refactored to loop over all new frames from the last one that it saw. It fully populates the view at each iteration, tho only the last iteration will be rendered. Populating the view is not very expensive, since all we do is delta the entity state, and copy entities, particles, etc.. into the view. So while there is some "throw-away" work with this approach, it's minimal.

This actually fixes a lot of issues:

 * Animation changes falling through the cracks because a server frame was never seen by the cgame.
 * Sounds, explosions and other events simply falling through the cracks and never being heard or seen.
 * This *should* enable an elegant fix for prediction error interpolation, altho right now this is not yet corrected.

One very obvious glitch with this change as-is, is that particles flicker for some reason. If you could help me understand that one, that'd be great!

### DIFF
--- a/src/cgame/default/cg_entity.c
+++ b/src/cgame/default/cg_entity.c
@@ -53,109 +53,15 @@ _Bool Cg_IsSelf(const cl_entity_t *ent) {
 /**
  * @return True if the entity is ducking, false otherwise.
  */
-_Bool Cg_IsDucking(const entity_state_t *ent) {
+_Bool Cg_IsDucking(const cl_entity_t *ent) {
 
 	vec3_t mins, maxs;
-	UnpackBounds(ent->bounds, mins, maxs);
+	UnpackBounds(ent->current.bounds, mins, maxs);
 
 	const vec_t standing_height = (PM_MAXS[2] - PM_MINS[2]) * PM_SCALE;
 	const vec_t height = maxs[2] - mins[2];
 
 	return standing_height - height > PM_STOP_EPSILON;
-}
-
-/**
- * @brief Adds client-side player effects such as breath puffs in rain or snow
- */
-static void Cg_AddBreathPuffs(cl_entity_t *ent) {
-
-	if (ent->animation1.animation < ANIM_TORSO_GESTURE) { // death animations
-		return;
-	}
-
-	if (cgi.client->ticks < ent->breath_puff_time) {
-		return;
-	}
-
-	cg_particle_t *p;
-
-	vec3_t pos;
-	VectorCopy(ent->origin, pos);
-
-	if (Cg_IsDucking(&ent->current)) {
-		pos[2] += 18.0;
-	} else {
-		pos[2] += 30.0;
-	}
-
-	vec3_t forward;
-	AngleVectors(ent->angles, forward, NULL, NULL);
-
-	VectorMA(pos, 8.0, forward, pos);
-
-	const int32_t contents = cgi.PointContents(pos);
-
-	if (contents & MASK_LIQUID) {
-		if ((contents & MASK_LIQUID) == CONTENTS_WATER) {
-
-			if (!(p = Cg_AllocParticle(PARTICLE_BUBBLE, cg_particles_bubble))) {
-				return;
-			}
-
-			p->lifetime = 1000 - (Randomf() * 100);
-			p->effects |= PARTICLE_EFFECT_COLOR | PARTICLE_EFFECT_SCALE;
-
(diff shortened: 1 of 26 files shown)


src/cgame/default/cg_main.c
```diff
@@ -283,7 +283,6 @@ cg_export_t *Cg_LoadCgame(cg_import_t *import) {
 	cge.ParseMessage = Cg_ParseMessage;
 	cge.PredictMovement = Cg_PredictMovement;
 	cge.UpdateView = Cg_UpdateView;
-	cge.PopulateView = Cg_PopulateView;
 	cge.DrawFrame = Cg_DrawFrame;
 
 	return &cge;
```
jdolan: PopulateView and UpdateView are now one.

src/client/cl_entity.c
```diff
@@ -348,38 +348,61 @@ static void Cl_UpdateLerp(void) {
...
+			} else {
+				VectorCopy(ent->current.angles, ent->angles);
+			}
 		}
+
+		Cl_UpdateView();
 	}
+
+	cl.render_frame = &cl.frames[cl.frame.frame_num & PACKET_MASK];
 }
```
jdolan: This is the meat of the change. Basically, after reading all updates from the server, enter a loop where we start at the last frame we simulated, and run up to the most recently received server frame. At each iteration, we fully populate the view. This ensures nothing falls through the cracks, so to speak.

src/client/cl_screen.c
```diff
@@ -343,8 +343,6 @@ void Cl_UpdateScreen(void) {
 
 	if (cls.state == CL_ACTIVE) {
 
-		Cl_UpdateView();
-
```
jdolan: This is now called [sometimes multiple times per frame] from Cl_Interpolate.

src/client/cl_view.c
```diff
@@ -194,20 +189,15 @@ void Cl_UpdateView(void) {
...
 
-	const cl_frame_t *frame = &cl.frames[(cl.frame.frame_num - 1) & PACKET_MASK];
-	const player_state_t *from = frame->frame_num == cl.frame.frame_num - 1 ? &frame->ps : &cl.frame.ps;
+	const player_state_t *ps = cl.delta_frame ? &cl.delta_frame->ps : &cl.frame.ps;
 
-	Cl_UpdateOrigin(from, &cl.frame.ps);
+	Cl_UpdateOrigin(ps, &cl.frame.ps);
 
-	Cl_UpdateAngles(from, &cl.frame.ps);
+	Cl_UpdateAngles(ps, &cl.frame.ps);
```
jdolan: This delta player state logic was always unnecessarily ugly. Just use cl.delta_frame.

src/client/cl_view.c
```diff
@@ -194,20 +189,15 @@ void Cl_UpdateView(void) {
...
 
 	Cl_UpdateViewSize();
 
-	// allow client game to modify view (bob, third person, etc)
 	cls.cgame->UpdateView(&cl.frame);
-
-	// create the thread which populates the view
-	r_view.thread = Thread_Create((ThreadRunFunc) cls.cgame->PopulateView, &cl.frame);
 }
```
jdolan: The view is no longer asynchronously populated. I'm okay with this -- this was never the "right" way to do async, anyway. In fact, this change moves us closer to the correct model of "prepare a view, and send it off to be rendered."

src/cgame/default/cg_entity.c
```diff
@@ -65,100 +65,6 @@ _Bool Cg_IsDucking(const entity_state_t *ent) {
...
-
-		for (int32_t i = 0; i < 3; i++) {
-			p->vel[i] += 2.0 * Randomc();
-		}
-
-		ent->breath_puff_time = cgi.client->ticks + 3000;
-	}
-}
-
-/**
```
jdolan: Just moved the breath stuff into cg_entity_trail. It's kind of a trail.

PR #293, opened 2016-12-15
Features/cm split

Paril:
Not entirely ready yet, this is for review.

This PR has the material system moved into libcollision. I have some code in there left over to ensure there are no leaks in the ref system.

Basically, it's set up to allow the server, client and renderer each to load and store material references without allocating 3 sets of materials. They can all call Cm_LoadMaterial to get a cm_material_t, which contains all of the parsed data from the file. In the future, there will also be a Cm_SaveMaterial routine to write the material back to the disk. The cm_material_t is reference counted, meaning the server and client can both request the same material and get the same reference without worrying about leaks or dupes. When you're done with a reference, you call UnrefMaterial to get rid of it.

The renderer is the only system that requires special code because it stores images and whatnot. The renderer has an "extension" of cm_material_t and cm_stage_t, called r_material_t and r_stage_t (I am so clever right?) and R_LoadMaterial which essentially "converts" a cm_material_t into an r_material_t. The r_material_t retains a reference to the cm_material_t and loads all of the images. When the r_material_t media gets freed, it unrefs the cm_material_t. This will actually save us a lot of time on an r_restart, because the cm_material_t doesn't need to go anywhere if the server has already loaded the collision model of the map which will have a reference to it already! This isn't implemented in the PR, only the renderer uses the cm_material_t, but soon(tm), my friends.

I don't know if this is the best way of approaching the system, but hey, it made sense in my head and it might work out. In theory, we can do the same thing for cm_bsp and r_bsp down the road. I like the pattern, personally!


src/client/renderer/r_mesh_model.c
```diff
@@ -31,7 +31,11 @@ static void R_LoadMeshMaterial(r_model_t *mod) {
 	Dirname(mod->media.name, skin);
 	strcat(skin, "skin");
 
-	mod->mesh->material = R_LoadMaterial(skin);
+	cm_material_t *cm = Cm_LoadMaterial(skin);
+
+	mod->mesh->material = R_LoadMaterial(cm);
+
+	Cm_UnrefMaterial(cm);
```
jdolan: Seems to me like a little R_LoadMaterial wrapper which does this unref would be useful. I think this is the 3rd time I've seen the "grab, load, unref" triplet :)

src/collision/cm_material.h
```diff
@@ -0,0 +1,155 @@
...
+	cm_stage_pulse_t pulse;
+	cm_stage_stretch_t stretch;
+	cm_stage_rotate_t rotate;
+	cm_stage_scroll_t scroll;
+	cm_stage_scale_t scale;
+	cm_stage_terrain_t terrain;
+	cm_stage_dirt_t dirt;
+	cm_stage_anim_t anim;
+	struct cm_stage_s *next;
+} cm_stage_t;
```
jdolan: I wonder, actually, if collision and quemap actually need the stages.. I think it's entirely possible that all of the stage definition stuff could remain purely in the renderer. Collision and quemap would only ever care about "primary" material properties like hardness, contents / surface flags, light emission, etc.

Of course, if this code is to write / save materials to disk, I guess it needs the full structure, or maybe we could use a callback approach to help with saving materials. Idk. Something to think about.

Paril: Only place that will need it is the client, yeah. The only thing is that, right now, once the material is loaded it's done for and there's no way to trigger "load more" down the road. There'd need to be a nice way to pass the loading off to another routine to continue reading or something.. and who knows, maybe collision will eventually need a stage down the road?

src/client/renderer/r_mesh_model.c
Paril: Yep!

src/collision/cm_material.h
jdolan: Yea. Let's keep it in here for now then.

src/collision/cm_model.c
```diff
@@ -422,6 +423,15 @@ static void Cm_LoadBspAreaPortals(const d_bsp_lump_t *l) {
...
 
+	// unload materials
+	for (int32_t i = 0; i < cm_bsp.num_surfaces; ++i) {
+		cm_bsp_surface_t *surf = &cm_bsp.surfaces[i];
+
+		if (surf->material) {
+			Cm_UnrefMaterial(surf->material);
+		}
+	}
+
```
jdolan: So in theory, the collision model could unload the map, but the materials could stay resident if quemap or the renderer or something still needed them? Seems robust.. hopefully doesn't provide rope to hang ourselves with :)

src/collision/cm_model.c
```diff
@@ -471,6 +486,14 @@ cm_bsp_model_t *Cm_LoadBspModel(const char *name, int64_t *size) {
 	Cm_LoadBspAreas(&header.lumps[BSP_LUMP_AREAS]);
 	Cm_LoadBspAreaPortals(&header.lumps[BSP_LUMP_AREA_PORTALS]);
 
+	// unref the materials from the list
+	for (uint32_t i = 0; i < materials->len; ++i) {
+		cm_material_t *material = g_array_index(materials, cm_material_t *, i);
+		Cm_UnrefMaterial(material);
+	}
+
+	g_array_free(materials, true);
```
jdolan: What's the motivation for using `GArray` instead of an array of structs here? This is different from how we load everything else.

PR #296, opened 2016-12-20
Add simple water ripples when most entities enter water

kaadmy:
Ripples when entering water (should work with lava/slime, too) that work with most entities, such as all current weapons and the player. Ripples don't work on the flag in CTF, dunno why...

### DIFF
--- a/src/cgame/default/cg_types.h
+++ b/src/cgame/default/cg_types.h
@@ -86,4 +86,3 @@ typedef struct cg_particles_s {
 #endif /* __CG_LOCAL_H__ */
 
 #endif /* __CG_TYPES_H__ */
-
--- a/src/game/default/g_physics.c
+++ b/src/game/default/g_physics.c
@@ -26,26 +26,27 @@
  * @brief Spawn a liquid ripple between a start and end point, setting pos1 to
  * NULL will not test a trace
  */
-void G_LiquidRipple(g_entity_t *ent, const vec3_t pos1, const vec3_t pos2, const vec_t size) {
-	vec3_t end;
+void G_LiquidRipple(g_entity_t *ent, const vec3_t start, const vec3_t end, const vec_t size) {
+	vec3_t pos;
 
-	_Bool doadd = true;
+	if (start != NULL) {
+		const cm_trace_t tr = gi.Trace(start, end, NULL, NULL, ent, MASK_LIQUID);
 
-	if (pos1 != NULL) {
-		cm_trace_t tr = gi.Trace(pos1, pos2, NULL, NULL, ent, MASK_LIQUID);
-		doadd = (tr.fraction != 1.0 && !tr.all_solid);
-		VectorCopy(tr.end, end);
+		if (!(tr.fraction != 1.0 && !tr.all_solid))
+			return;
+
+		VectorCopy(tr.end, pos);
 	} else {
-		VectorCopy(pos2, end);
+		VectorCopy(end, pos);
 	}
 
-	if (doadd) {
-		gi.WriteByte(SV_CMD_TEMP_ENTITY);
-		gi.WriteByte(TE_RIPPLE);
-		gi.WritePosition(end);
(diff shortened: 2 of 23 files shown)


src/game/default/g_physics.c
```diff
@@ -23,6 +23,33 @@
...
+ * @brief Spawn a liquid ripple between a start and end point, setting pos1 to
+ * NULL will not test a trace
+ */
+void G_LiquidRipple(g_entity_t *ent, const vec3_t pos1, const vec3_t pos2, const vec_t size) {
+	vec3_t end;
+
+	_Bool doadd = true;
+
+	if (pos1 != NULL) {
+		cm_trace_t tr = gi.Trace(pos1, pos2, NULL, NULL, ent, MASK_LIQUID);
```
jdolan: `tr` can be `const`. Also, I'd prefer a `return` instead of the `doadd` business.

src/game/default/g_physics.c
```diff
@@ -23,6 +23,33 @@
...
+	} else {
+		VectorCopy(pos2, end);
+	}
+
+	if (doadd) {
+		gi.WriteByte(SV_CMD_TEMP_ENTITY);
+		gi.WriteByte(TE_RIPPLE);
+		gi.WritePosition(end);
+		gi.WriteVector(size);
+		gi.Multicast(pos2, MULTICAST_PVS, NULL);
```
jdolan: Does `pos2` guarantee that we broadcast on the top-side of the water? I imagine, for opaque water, the PVS broadcast position is pretty important here.

src/game/default/g_physics.h
```diff
@@ -26,6 +26,7 @@
 
 #ifdef __GAME_LOCAL_H__
 	#define DEFAULT_GRAVITY 800.0
+	void G_LiquidRipple(g_entity_t *ent, const vec3_t pos1, const vec3_t pos2, const vec_t size);
```
jdolan: I'd probably name these `org` and `prev` instead of `pos1` and `pos2`. Also, if we're passing in `ent`, isn't `pos1` just `ent->s.origin`?

src/cgame/default/cg_types.h
```diff
@@ -86,4 +86,3 @@ typedef struct cg_particles_s {
 #endif /* __CG_LOCAL_H__ */
 
 #endif /* __CG_TYPES_H__ */
-
```
kaadmy: Yep, it does. I was going to amend this commit without this change, but git was complaining.

src/game/default/g_physics.c
kaadmy: I don't know why I didn't do that in the first place.

src/game/default/g_physics.c
kaadmy: `pos2` should be flush with the water, in `cg_temp_entity.c`, it's lifted up 1u.

src/game/default/g_physics.h
kaadmy: I wanted `ent` optional, because some things like railgun beams *might* not work with that.

src/game/default/g_physics.c
jdolan: Okay, flush with the water will probably be problematic for the broadcast. For opaque water, you'll need to be safely above the water to correctly send the event to clients. Otherwise, the PVS check might fail. So I'd lift it up 1u on the server, and leave it alone on the client.

src/game/default/g_physics.h
jdolan: Ah, makes sense!

src/game/default/g_physics.c
kaadmy: I just realized you were asking something else, but I'll do that instead :D

kaadmy:
I'm using the same particle image from  last time that's already in quetoo-data.

`target/particles/ripple.tga`

Haven't tested with lava/slime, but it looks pretty good with Edge's water.

PR #297, opened 2016-12-21
Liquid caustics

kaadmy:
Title.

Don't have to merge just yet, performance seems okay-ish, but doesn't seem to change much from normal if you're not looking at water.


kaadmy:
Should I still try to optimize this by batching underwater surfaces or is performance good enough as is?

jdolan:
Unless you move caustics to its own program, then batching won't yield a tremendous performance increase. So I'll leave that up to you. I didn't see a noteworthy hit with this change as-is. It's good enough for 1.0.

PR #298, opened 2016-12-21
Stainmarps!

Paril:
Just for discussion currently.


jdolan:
Are you planning on migrating all of the burn marks to stains? Blood stains too? 👍

jdolan:
There's a serious bug around lightmap retention when reloading the same map. I spawned into Frag Pipe, and for some reason `mouse 1` was not firing my weapon. I've no idea if this is related to your PR, but regardless... I reloaded the map `map fragpipe` to see if it would start behaving. That yielded this:

quetoo022

Kinda cool, but clearly busted. So I `map fragpipe`'d again, and got this:

quetoo023

All subsequent reloads remained like that.

As a separate note: I think the stain intensity for burn marks should be increased.

src/cgame/default/cg_particle.c
```diff
@@ -208,6 +208,33 @@ static _Bool Cg_UpdateParticle_Spark(cg_particle_t *p, const vec_t delta, const
...
+
+	if (p->blood.time < cgi.client->ticks) {
+		const vec4_t blood_color = {
+			0.9 + Randomf() * 0.1,
+			0.0,
+			0.0,
+			0.3 + Randomc() * 0.25
+		};
+
+		const vec3_t blood_pos = {
```
jdolan: This variable is unused.

jdolan:
Another note: there is logic in `G_BurnMark` to not leave burns on moving objects, because the shitty decal particles would float in space if we did. That is not the case with stain maps, right? They should recurse even BSP sub-models, and stain accordingly, so that we could leave burns on plats, doors and other objects. This is just one more reason why I would really like to abandon the decal burn marks and just use stains for these kinds of effects.

src/collision/cm_trace.c
```diff
@@ -246,91 +246,168 @@ static void Cm_TestInLeaf(cm_trace_data_t *data, int32_t leaf_num) {
 }
 
 /**
+ * @brief Temporary structured used for a single trace node
```
jdolan: Dude, this `cm_*` stuff really has to be done in another branch. We can't risk regression like this in trying to get stain maps in before release. Do you agree?

src/client/renderer/r_bsp_light.c
```diff
@@ -322,3 +325,292 @@ void R_DrawBspLights(void) {
...
+
+			const r_pixel_t smax = surf->st_extents[0];
+			const r_pixel_t tmax = surf->st_extents[1];
+
+			vec_t ftacc = 0.0;
+
+			byte *pfBL = surf->stainmap_buffer;
+			_Bool changes = false;
+
+			for (uint16_t t = 0; t < tmax; t++, ftacc += 16.0) {
```
jdolan: I think you want to increment by lightmap_scale here, instead of 16.0. Might this be part of the problem with scaling stains correctly?

src/client/renderer/r_bsp_light.c
```diff
@@ -322,3 +325,292 @@ void R_DrawBspLights(void) {
...
+				const uint32_t td = (uint32_t) fabs(local[1] - ftacc);
+
+				uint16_t s = 0;
+				vec_t fsacc = 0.0;
+
+				for (; s < smax; s++, fsacc += 16.0, pfBL += 3) {
```
jdolan: And again, lightmap_scale vs 16.0.

src/client/renderer/r_bsp_light.c
```diff
@@ -322,3 +325,292 @@ void R_DrawBspLights(void) {
...
+	for (uint16_t i = 0; i < r_view.num_lights; i++, l++) {
+		if (e) {
+			Matrix4x4_Transform(&e->inverse_matrix, light_origins[i], l->origin);
+			R_MarkLight(l, nodes + e->model->bsp_inline->head_node);
+		} else {
+			VectorCopy(light_origins[i], l->origin);
+		}
+	}
+}
+		*/
```
jdolan: Is this not needed?

src/client/renderer/r_bsp_light.c
```diff
@@ -322,3 +325,292 @@ void R_DrawBspLights(void) {
...
+			if (!entity) {
+				continue;
+			}
+
+			Matrix4x4_Transform(&entity->inverse_matrix, org, origin);
+		} else {
+			VectorCopy(org, origin);
+		}
+
+		R_StainNode(origin, color, size, node, &num);
```
jdolan: It'd be cleaner and more efficient to loop once over `cl.frame.entities` and deal with ones that reference BSP models, than to iterate the BSP models and then iterate the entities in loop.

src/client/renderer/r_bsp_light.c
```diff
@@ -322,3 +325,292 @@ void R_DrawBspLights(void) {
...
+							pfBL[i] = new_color;
+							changes = true;
+							(*num)++;
+						}
+					}
+				}
+			}
+
+			if (!changes) {
+				continue;
```
jdolan: Rather than tracking `changes` can't you simply check `surf->stainmap_dirty`?

PR #320, opened 2017-01-05
Paril/dll support

Paril:


src/cgame/default/viewcontrollers/CreateServerViewController.c
```diff
@@ -282,13 +282,19 @@ static void initialize(Class *clazz) {
...
-	.instanceSize = sizeof(CreateServerViewController),
-	.interfaceOffset = offsetof(CreateServerViewController, interface),
-	.interfaceSize = sizeof(CreateServerViewControllerInterface),
-	.initialize = initialize,
-};
+Class *_CreateServerViewController(void) {
+	static Class clazz;
+	
+	if (!clazz.name) {
+		clazz.name = "CreateServerViewController";
```
jdolan: Instead of `!clazz.name`, we should use `do_once` to make these all thread-safe. See URLSession.c or Boole.c for an example of `do_once`.

Paril: Good point, I didn't consider that.

Issue #313, opened 2016-12-30
Too many BSP representations


Paril:
So this kind of goes back to #63 - our regular usage of Quetoo and Quemap have three separate BSP implementations.

Quemap works solely in d_bsp_*, which is the "bspfile" stuff that is basically a raw representation of the BSP as it exists on the disk.

Quetoo Server works in cm_bsp_*, which is basically a copy of d_bsp_* except the structures have additions to it, certain bits are expanded/transformed, but for the most part it's exactly the same as d_bsp_*.

Quetoo Client works in r_bsp_*, which is basically a copy of cm_bsp_* except some useless lumps are dropped, some new lumps are loaded, and the structures are set up just to hold data that only pertains to rendering.

As I was doing the material subsystem, I realized that I really like the concept I had with cm_material_t which r_material holds a reference to and just contains the additions to it. I'm thinking maybe we can do this with BSP as well so that we have one global "BSP load" function living in common (honestly I think the cm_bsp_* system can be used in both server and quemap with no issues at all) and then have the renderer just tack on the stuff it needs.

The only real downside I can think of is the question of how to manage lumps that only certain systems need - for instance, cm_* doesn't need lightmap data, but Quemap and renderer do - rather, quemap needs all lumps since it could write to them.

I have some ideas for all of this, but would be nice to have some input first before I decide to do anything about this. It's more of a rainy day thing since it doesn't really affect what we have.

jdolan:
In looking over this, I am slightly tempted to move `src/collision/cm_bsp.{c,h}` to `src/bsp.{c,h}` to build `libbsp.la`, and have collision and friends all link against it. But that's for a rainy day.

Paril:
That is something I thought about, same for materials (libmaterial)

PR #321, opened 2017-01-07
Paril/bsp reps

Paril:
Mostly for review, I need to run a few more tests to be sure it's ready for prime.
This is a resolution to #313 which was a rainy day thing for me. This adds a new set of Bsp_* functions, currently settling in libcollision (should this be a separate lib? libbsp? 😄), that reflect the disk layout of a BSP file, and provide operations to load, save, and edit lumps in raw memory.

The class handles swapping for you, so you don't have to ever touch swapping from loading or saving BSP files (it even only swaps if the CPU is indeed big-endian, so on little-endian there's no overhead). Currently, it doesn't verify the BSP lumps as being legit, so the cm_bsp and r_bsp systems still do their own verification, but I think this is a much cleaner approach.

So, while we still have three BSP representations, the main memory still resides in bsp_file_t; cm_bsp_t only contains additions related to collision (it copies the data out so that, in theory, you can unload certain lumps from bsp_file_t when you're done with them, but that's not done yet) and r_bsp_t only contains additions related to rendering.

This also starts to resolve an issue @jdolan had with shadows - r_bsp_plane_t is now just a typedef to cm_bsp_plane_t, and bsp->planes now just references the cm_bsp's "planes" ptr. Renderer and collision have the same plane ptrs and can be compared directly.

Ideally I wanted there to be some sort of pseudo-inheritance going on where the r_bsp stuff can inherit from the cm_bsp ones and add the stuff it needs, but then we'd have some stupid typings like "texinfo->cm_texinfo->scale" and I wanted to avoid that, so the data is just copied over.

This commit also moves all of the old stack-allocated memories (both quemap and libcollision allocated about 300 MB of static heap memory) and uses a heap-allocated setup. The savings don't mean much for quemap because quemap still allocates that amount of memory as scratch for the BSP process, but for libcollision it's a huge saving.


src/client/renderer/r_model.c
```diff
@@ -153,12 +152,26 @@ r_model_t *R_LoadModel(const char *name) {
...
-		// free the file
-		Fs_Free(buf);
+			Fs_Load(file_name, &buf);
+
+			// load it
+			format->Load(mod, buf);
+
+			// free the file
+			Fs_Free(buf);
+		}
```
jdolan: A little gross that the format loaders aren't taking the same kind of data, but I get it.

src/tools/quemap/brush.c
```diff
@@ -43,7 +43,7 @@ static void BoundBrush(bsp_brush_t *brush) {
 /**
  * @brief
  */
-static void CreateBrushWindings(bsp_brush_t *brush) {
+static void CreateBrushWindings(qbsp_brush_t *brush) {
```
jdolan: Why the q? We don't do that anywhere else.

jdolan: Oh.. does Quemap have its own representation of this struct?

src/tools/quemap/qbsp.c
```diff
@@ -297,6 +297,32 @@ static void ProcessModels(void) {
...
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_LEAFS, MAX_BSP_LEAFS);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_LEAF_FACES, MAX_BSP_LEAF_FACES);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_LEAF_BRUSHES, MAX_BSP_LEAF_BRUSHES);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_EDGES, MAX_BSP_EDGES);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_FACE_EDGES, MAX_BSP_FACE_EDGES);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_MODELS, MAX_BSP_MODELS);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_BRUSHES, MAX_BSP_BRUSHES);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_BRUSH_SIDES, MAX_BSP_BRUSH_SIDES);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_AREAS, MAX_BSP_AREAS);
+	Bsp_AllocLump(&bsp_file, BSP_LUMP_AREA_PORTALS, MAX_BSP_AREA_PORTALS);
```
jdolan: I like this.

src/client/renderer/r_model.c
Paril: Yeah I wasn't sure how to best approach this. I can rework the rest of the loaders to stream the data instead of loading the whole thing into memory maybe?

src/tools/quemap/brush.c
Paril: Quemap has bsp_brush_t which represents the CSG data, from what I can tell. It's then compiled down into regular bsp brushes.

The only reason this collision happened was because of the naming of the BSP stuff. I didn't want to keep the d_bsp_ prefixes (I wanted the structs to match the functions and I didn't want to have the funcs named D_Bsp_Load...), I didn't want to call them Cm either because Cm has its own BSP stuff, and map_brush_t already existed so I couldn't use that either. Any ideas would be appreciated!

src/tools/quemap/qbsp.c
Paril: It's probably the part I hate the most, but this basically replaces the part where they're statically allocated. In an ideal world, these would only be allocated as the system needs to expand the amount (in, say, EmitBrushes, it would allocate only as many as it needs; in EmitFace it would reallocate if we've run out, etc). It's a minor detail though.

Paril:
Cool, so I still have a few things left to do on this branch (a few fixmes/todos) but it's crashless and no memory leaks that I can find.

src/client/renderer/r_model.c
jdolan: Is it just not practical / convenient for the BSP loader to take a buffer of the entire file? Is that too large an allocation to deal with elegantly? I know some BSPs can come in at ~50MB.

src/tools/quemap/brush.c
jdolan: Just call Quemap's `brush_t` then?

jdolan: I only say that because Quemap suffers from short, nondescript naming pretty much throughout ;D

src/tools/quemap/qbsp.c
jdolan: I was being sincere. It's very readable.

src/client/renderer/r_model.c
Paril: It's just unnecessary, especially if you're only going to load a single lump. I know we're not exactly memory-bottlenecked these days but if I can avoid loading an entire file in memory and streaming instead I always take that route.

src/tools/quemap/brush.c
Paril: we can do that, yeah

Paril:
I think this is ready now. Mind giving my newer pushes the once-over?

PR #324, opened 2017-01-08
Issue/reduce client traces

jdolan:


src/client/cl_entity.c
```diff
@@ -42,16 +42,25 @@ static void Cl_ParsePlayerState(const cl_frame_t *delta_frame, cl_frame_t *frame
...
 	if (from->model1 != to->model1) {
 		return false;
 	}
 
 	VectorSubtract(from->origin, to->origin, delta);
 
-	if (VectorLength(delta) > 256.0) {
+	if (VectorLength(delta) > 64.0) { // MAX_SPEED * QUETOO_TICK_SECONDS = 60.0
 		return false;
 	}
```
jdolan: Just a couple minor adjustments to `Cl_ValidDeltaEntity` to make it all-encompassing. This makes the conditional that relies on it below simper.

src/client/cl_entity.c
```diff
@@ -87,23 +96,8 @@ static void Cl_ReadDeltaEntity(cl_frame_t *frame, entity_state_t *from, uint16_t
...
-				Matrix4x4_Invert_Simple(&ent->inverse_matrix, &ent->matrix);
-			}
-		} else { // bounding-box entities
-			if (bits & U_ORIGIN) {
-				Matrix4x4_CreateFromEntity(&ent->matrix, ent->current.origin, vec3_origin, 1.0);
-				Matrix4x4_Invert_Simple(&ent->inverse_matrix, &ent->matrix);
-			}
-		}
-	}
-
```
jdolan: Rather than only updating these matrices when the entity is updated here, I've moved this code into `Cg_Interpolate` so that (when we resolve the prediction issues), we can actually interpolate the collision matrix, and have pixel-perfect client-side collision.

src/client/cl_entity.c
```diff
@@ -87,23 +96,8 @@ static void Cl_ReadDeltaEntity(cl_frame_t *frame, entity_state_t *from, uint16_t
...
 	// mark the lighting cache as dirty
-	if (bits & U_ORIGIN) {
+	if (bits & (U_ORIGIN | U_TERMINATION | U_ANGLES | U_MODELS | U_BOUNDS)) {
 		ent->lighting.state = LIGHTING_DIRTY;
 	}
```
jdolan: It's pointless to update lighting every single renderer frame. Updating it every server frame, if something changes, is totally adequate. This should shave the time spent in `Cl_Trace`, and it also tidies up `Cl_Interpolate` below.

src/client/cl_entity.c
```diff
@@ -401,6 +392,39 @@ void Cl_Interpolate(void) {
...
+				angles = vec3_origin;
+				UnpackBounds(ent->current.bounds, ent->mins, ent->maxs);
+			}
+
+			Cm_EntityBounds(ent->current.solid, ent->current.origin, angles, ent->mins, ent->maxs, ent->abs_mins, ent->abs_maxs);
+
+			Matrix4x4_CreateFromEntity(&ent->matrix, ent->current.origin, angles, 1.0);
+			Matrix4x4_Invert_Simple(&ent->inverse_matrix, &ent->matrix);
+		}
 	}
```
jdolan: Comment says it all, really. Changing `ent->current.var` to `ent->var` here yields pixel-perfect collision for shadows and whatnot, but this makes the prediction / error interpolation code go awry. I was seeing strange error correction when riding the elevator on Edge, particularly downward. So I've left it so that it works as the old code did, snapping entities' collision states against the server frame.

In thinking about how the prediction code works, I think I have some understanding of the problem. The server "thinks" in discrete 25ms snaps, while the client runs at e.g. 8ms snaps (125fps). Therefore, the client is going to predict against different snapshots of entity state than the server is. In theory, you'd think this would all even out in the end, but I guess it doesn't.

src/client/cl_main.c
```diff
@@ -622,16 +618,19 @@ void Cl_Frame(const uint32_t msec) {
...
 
-	if (cls.state == CL_ACTIVE) {
 		Cl_Interpolate();
 
 		Cl_PredictMovement();
 
 		Cl_UpdateView();
+	} else {
+		Cl_SendCommands();
 	}
```
jdolan: Simply updated `Cl_Frame` to not bother with `Cl_UpdateMovementCommand` until we're `CL_ACTIVE`.

src/client/cl_predict.c
```diff
@@ -37,13 +37,12 @@ static int32_t Cl_HullForEntity(const entity_state_t *ent) {
...
+	const cl_entity_t *ent = &cl.entities[s->number];
 
-	if (ent->client) {
-		return Cm_SetBoxHull(emins, emaxs, CONTENTS_MONSTER);
+	if (s->client) {
+		return Cm_SetBoxHull(ent->mins, ent->maxs, CONTENTS_MONSTER);
 	} else {
-		return Cm_SetBoxHull(emins, emaxs, CONTENTS_SOLID);
+		return Cm_SetBoxHull(ent->mins, ent->maxs, CONTENTS_SOLID);
 	}
```
jdolan: Entities now have their `mins` and `maxs` readily available, so there's no need to unpack them here anymore.

src/client/cl_predict.c
```diff
@@ -64,9 +63,14 @@ int32_t Cl_PointContents(const vec3_t point) {
...
 
-		const int32_t head_node = Cl_HullForEntity(s);
 		const cl_entity_t *ent = &cl.entities[s->number];
 
+		if (ent == cl.entity) {
+			continue;
+		}
+
+		const int32_t head_node = Cl_HullForEntity(s);
+
```
jdolan: This was slightly alarming: `Cl_PointContents` did not skip self! I can't imagine that's a good thing, so it does now.

src/client/cl_predict.c
```diff
@@ -102,16 +107,16 @@ static void Cl_ClipTraceToEntities(cl_trace_t *trace) {
 			continue;
 		}
 
-		if (s->number == cl.client_num + 1) {
-			continue;
-		}
-
```
jdolan: This was redundant, as the "self-check" is now handled by the cgame with `Cg_Self` and `cl.entity` (see below).

src/client/cl_predict.c
```diff
@@ -166,6 +171,8 @@ cm_trace_t Cl_Trace(const vec3_t start, const vec3_t end, const vec3_t mins, con
 	trace.skip = skip;
 	trace.contents = contents;
 
+	Cm_TraceBounds(start, end, mins, maxs, trace.box_maxs, trace.box_maxs);
+
 	Cl_ClipTraceToEntities(&trace);
```
jdolan: And finally, borrow the box intersection logic from the server so that we only clip traces to entities that reside within the trace's bounding box.

src/collision/cm_trace.c
```diff
@@ -501,3 +501,73 @@ cm_trace_t Cm_TransformedBoxTrace(const vec3_t start, const vec3_t end, const ve
...
+	for (int32_t i = 0; i < 3; i++) {
+		if (end[i] > start[i]) {
+			bounds_mins[i] = start[i] + mins[i] - 1.0;
+			bounds_maxs[i] = end[i] + maxs[i] + 1.0;
+		} else {
+			bounds_mins[i] = end[i] + mins[i] - 1.0;
+			bounds_maxs[i] = start[i] + maxs[i] + 1.0;
+		}
+	}
+}
```
jdolan: This was all lifted from `sv_world.c` and commonized.

PR #334, opened 2017-01-09
Paril/remove mat refs

Paril:


src/collision/cm_main.c
```diff
@@ -33,5 +33,4 @@ void Cm_Init(void) {
  */
 void Cm_Shutdown(void) {
 
-	Cm_ShutdownMaterials();
```
jdolan: Looks like we could do away with Cm_Init and Cm_Shutdown entirely.

src/client/renderer/r_types.h
```diff
@@ -359,7 +359,7 @@ typedef struct r_material_s {
 	// from media
 	r_media_t media;
 
-	const struct cm_material_s *cm; // link to cm material
+	struct cm_material_s *cm; // copy of cm material
```
jdolan: Then why is it a pointer?

src/client/renderer/r_material.c
```diff
@@ -706,22 +712,27 @@ r_material_t *R_ConvertMaterial(cm_material_t *cm, const _Bool unref) {
...
+	char material_key[MAX_QPATH];
+	StripExtension(name, material_key);
+	Cm_NormalizeMaterial(material_key, material_key, sizeof(material_key));
+	g_snprintf(material_key, sizeof(material_key), "%s_mat", material_key);
+
+	if (!(mat = (r_material_t *) R_FindMedia(material_key))) {
+		mat = R_ConvertMaterial(Cm_LoadMaterial(name), false);
+	}
+
+	return mat;
```
jdolan: I'd love it if the renderer simply asked Cm_ for it's copy of the material, copied it into r_material_t, and continued along its merry way.

src/client/renderer/r_types.h
Paril: It was a copy initially and then I just realized I can reference the data like I did in the Bsp subsystem, so I did that instead.

src/client/renderer/r_material.c
Paril: But it's so elegant as it is, no extra allocations - the only thing that needs the special code is map materials since Cm loads them already, so we can just use the materials they already loaded/parsed.

src/collision/cm_main.c
Paril: Oh weird, I removed them locally but it didn't commit or something.

src/client/renderer/r_material.c
jdolan: I don't understand your obsession with this. LOL. The renderer loads its own copies of planes, nodes, texinfos, VIS, .. What is the issue?

src/collision/cm_material.h
```diff
@@ -83,71 +83,112 @@ typedef struct cm_stage_s {
...
+
+	/**
+	 * @brief The specular factor to use for the specular map.
+	 */
 	vec_t specular;
+
+	/**
+	 * @brief Pointer to the first stage in the stage list. NOT an array;
+	 * be sure to use ->next to traverse.
+	 */
```
jdolan: 👍

src/client/renderer/r_material.c
Paril: It doesn't any more - it only loads copies if it needs to, the rest it just reads from the cm pointer.

Paril:
So, interestingly, I tried having the renderer make its own copies but it introduced some problems because of the way sub-materials work. Multiple materials may reference the same r_material_t (mossyrockies on Edge for instance is referenced twice), and it becomes hard to free the r_material_ts then because the sub-references aren't connected to the final material list (they are mem_linked to their parent). It would either crash because multiple materials reference the same r_material_t whose cm_material_t was freed by the prior Free call, or you would have a memory leak because some of the sub-reference cm_material_ts will get discarded and lost along the way since they will resolve to an already-loaded r_material_t reference.

The only easy way to resolve this is to not free BSP materials (like I already do) and free them separately during BSP free, which would still require that whole map_material thing to be in place. I honestly don't think the system is going to get any more elegant than it is right now.

jdolan:
Looks like there's a memory leak with loaded files. I see this on exit after `quetoo +map torn`:

```
quit
Server quit
Server shutdown...
Game shutdown...
  Game shutdown...
Game down
Server down
Disconnecting from localhost...
Client game shutdown...
  Client game shutdown...
Client game down
Client console shutdown
Fs_PrintLoadedFiles: models/weapons/shotgun/tris.mat @ 0x7fb2bb70f0c0
Fs_PrintLoadedFiles: models/ammo/shells/tris.mat @ 0x7fb2bb748cd0
Fs_PrintLoadedFiles: models/health/medium/tris.mat @ 0x7fb2bb719b60
Fs_PrintLoadedFiles: models/weapons/railgun/tris.mat @ 0x7fb2bb709320
Fs_PrintLoadedFiles: models/weapons/rocketlauncher/tris.mat @ 0x7fb2bb70cb20
Fs_PrintLoadedFiles: models/armor/body/tris.mat @ 0x7fb2bb744b10
Fs_PrintLoadedFiles: models/weapons/supershotgun/tris.mat @ 0x7fb2bb735840
Fs_PrintLoadedFiles: models/weapons/lightning/tris.mat @ 0x7fb2bb726390
Fs_PrintLoadedFiles: models/ammo/slugs/tris.mat @ 0x7fb2b55e7ca0
Fs_PrintLoadedFiles: models/health/large/tris.mat @ 0x7fb2bd0e43c0
Fs_PrintLoadedFiles: models/weapons/machinegun/tris.mat @ 0x7fb2bb734a40
Fs_PrintLoadedFiles: models/weapons/grenadelauncher/tris.mat @ 0x7fb2bb709ad0
Fs_PrintLoadedFiles: models/ammo/bolts/tris.mat @ 0x7fb2bd0e4e80
Fs_PrintLoadedFiles: models/weapons/hyperblaster/tris.mat @ 0x7fb2bb7252f0
Fs_PrintLoadedFiles: materials/torn.mat @ 0x7fb2b6135e20
Fs_PrintLoadedFiles: models/armor/jacket/tris.mat @ 0x7fb2bb71cb60
Fs_PrintLoadedFiles: models/weapons/blaster/tris.mat @ 0x7fb2bb72b3f0
Fs_PrintLoadedFiles: models/armor/shard/tris.mat @ 0x7fb2bb712010
Fs_PrintLoadedFiles: models/ammo/cells/tris.mat @ 0x7fb2bb7225f0
Fs_PrintLoadedFiles: models/health/small/tris.mat @ 0x7fb2bb702210
Fs_PrintLoadedFiles: models/ammo/nukes/tris.mat @ 0x7fb2bb7368d0
Fs_PrintLoadedFiles: models/ammo/rockets/tris.mat @ 0x7fb2bb707720
Fs_PrintLoadedFiles: models/weapons/bfg/tris.mat @ 0x7fb2bb736ac0
Fs_PrintLoadedFiles: models/ammo/grenades/tris.mat @ 0x7fb2bb719f20
Fs_PrintLoadedFiles: models/health/mega/tris.mat @ 0x7fb2bb70c610
Fs_PrintLoadedFiles: models/ammo/bullets/tris.mat @ 0x7fb2bb716130
Fs_PrintLoadedFiles: models/armor/combat/tris.mat @ 0x7fb2bb703810
```

jdolan:
Fixed the memory leak, adjusted build for Autotools and Xcode. This is good for merge.

PR #338, opened 2017-01-12
Issue/330 particle trail entity delta

jdolan:
The issue here was that, at high latency, `delta_frame` is not the previous frame the client received. That is, as soon as `net_loop_latency> QUETOO_TICK_MILLIS`, `Cl_ValidDeltaEntity` would begin to fail for what should have been valid deltas. I couldn't figure out why, until I unwound how the delta entity code works. The server can only delta from what you've ack'ed, and at high latency, what you've ack'ed can be several updates old (100ms ping means your delta acks are 4+ frames old).

Therefore, determining valid delta needed significant adjustment. Quake2 always used a hack for this, where it subtracted the delta-from origin with the delta-to origin, and if that vector was > 256.0, it disabled lerp. But at say 2-300ms latency, it's entirely possible for a delta entity to move 256 units (in fact, rockets and other projectiles *always* do).

Because of Quake2's higher net frame interval, this was less apparent and less problematic, but our simulation is a lot more sensitive. The correct approach, then, is to determine delta validity against `ent->current`, not `from`. This has the added bonus of spawning trails earlier for everyone (even localhost / listen server clients) because the check for `ent->prev.model1` is now `ent->current.model1` like it should be.

### DIFF
--- a/src/client/cl_entity.c
+++ b/src/client/cl_entity.c
@@ -179,7 +179,7 @@ static void Cl_ParseEntities(const cl_frame_t *delta_frame, cl_frame_t *frame) {
 			}
 
 			if (from_number != number) {
-				Com_Warn("U_REMOVE: %u != %u\n", from_number, number);
+				Com_Debug(DEBUG_CLIENT, "U_REMOVE: %u != %u\n", from_number, number);
 			}
 
 			index++;
--- a/src/client/cl_entity.c
+++ b/src/client/cl_entity.c
@@ -40,7 +40,8 @@ static void Cl_ParsePlayerState(const cl_frame_t *delta_frame, cl_frame_t *frame
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/client/cl_entity.c
```diff
@@ -71,7 +74,7 @@ static _Bool Cl_ValidDeltaEntity(const cl_frame_t *frame, const cl_entity_t *ent
  * @brief Reads deltas from the given base and adds the resulting entity to the
  * current frame.
  */
-static void Cl_ReadDeltaEntity(cl_frame_t *frame, entity_state_t *from, uint16_t number, uint16_t bits) {
+static void Cl_ReadDeltaEntity(cl_frame_t *frame, const entity_state_t *from, uint16_t number, uint16_t bits) {
```
jdolan: Everything from here down is literally just variable renaming. The fix is all above.

PR #349, opened 2017-01-22
Paril/footsteps

Paril:
Ready to go!


Paril:
@Panjoo this merge now means you can have custom footstep sounds for all of your maps down the road. Currently we're waiting on Anthony to get us a good set of footsteps, but if you want to test it out, the new material argument is "footsteps" - you set it to a small string for the sound files (ie, "wood"), and then it searches for players/common/step_wood_n.wav, where n is 1-as many unique sounds as you want.

When we have a bunch of sounds from Anthony to use, we'll begin shimming them into maps (starting with grass probably, since that's an easy one).

PR #352, opened 2017-01-23
Issue/234 java aws update

jdolan:
Where to begin..

The Java based updater is now integrated into the Apple, Linux and MinGW builds. It is installed to the `lib` directory of the application bundle via `aws s3 copy ...` during the bundle creation.

`quetoo-update` is now a minimal executable that resolves the relative path to `quetoo-update-small.jar`, and invokes `java` to run it. `quetoo-update` lives in `bin` while `quetoo-update-small.jar` lives in `lib`.

I'd recommend wiping out the current release directories for_pc-windows_ on S3 as you hammer this branch into shape. Otherwise, we'll likely end up with some stale artifacts sneaking into folks' installs.


mingw-cross/Makefile
```diff
@@ -1,25 +1,22 @@
-# Makefile for Quetoo Windows distributable.
```
jdolan: @Paril If there's something you're looking for as a guide, I guess the MinGW Makefile would be the most appropriate thing.

src/mem.c
```diff
@@ -172,7 +172,7 @@ static void *Mem_Malloc_(size_t size, mem_tag_t tag, void *parent) {
 	void *data = (void *) (b + 1);
```
jdolan: Just fixing compiler warnings.

src/tools/quetoo-update/main.c
```diff
@@ -0,0 +1,96 @@
...
+#include <assert.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <glib.h>
+
+#ifdef _WIN32
+	#ifndef realpath
+		#define realpath(rel, abs) _fullpath(abs, rel, MAX_PATH)
```
jdolan: I tried the `realpath` thing you described here. Hope it works.

src/tools/quetoo-update/main.c
```diff
@@ -0,0 +1,96 @@
...
+
+	const char *jar = g_getenv("QUETOO_UPDATE_JAR") ?: DEFAULT_UPDATE_JAR;
+
+	gchar *quetoo_update_jar = get_quetoo_update_jar(argv[0], jar);
+	if (quetoo_update_jar) {
+
+		gchar *args[argc + 3];
+		args[0] = "java", args[1] = "-jar", args[2] = quetoo_update_jar;
+		memcpy(args[3], argv[1], (argc - 1) * sizeof(char *));
+		args[argc + 2] = NULL;
```
jdolan: So this is a little cheeky, but we're actually executing Java with a copy of our own `argv`, *after* the `java -jar ../path/to/jar` stuff.

src/tools/quetoo-update/main.c
```diff
@@ -0,0 +1,96 @@
...
+
+		const GSpawnFlags flags = G_SPAWN_SEARCH_PATH | G_SPAWN_CHILD_INHERITS_STDIN;
+
+		GError *error;
+
+		if (g_spawn_async(NULL, args, NULL, flags, NULL, NULL, NULL, &error)) {
+			status = 0;
```
jdolan: glib takes care of the `spawnlp` vs `execlp` stuff for us.

src/tools/update/main.c
```diff
@@ -1,274 +0,0 @@
...
-
-	printf("Press any key to close.\n");
-	getchar();
-
-#if defined(_WIN32)
-	_spawnlp(_P_NOWAIT, "post-update.bat", "post-update.bat", NULL);
-#endif
-
-	return status;
-}
```
jdolan: CYA!

jdolan:
@Panjoo We're close to merging in a new Java-based updater that has a little GUI and everything. No more console window and rsync woes. But before we do, I wanted to make sure you're able to run Java 8.

jdolan:
@Panjoo This has been merged but we're still sorting out Windows updater AND MinGW build issues. Please hold off on attempting to update until I follow up. Thanks!

PR #379, opened 2017-02-03
Ran make style, I may regret this

kaadmy:
Just ran make style to keep stuff cleaner later on


jdolan:
`make style` isn't perfect as-is. This is worth doing, but we have to be more selective. I'll run one tomorrow. Thanks for reminding me.

PR #387, opened 2017-02-05
Fix console linewrap colors

literallyvoid:
This fixes the linewrapping color issues in the chat log and the console.


kaadmy:
Don't you mean re-re-re-fix?

jdolan:
Oh, wtf. I must have copied / pasted. Thank you.

PR #392, opened 2017-02-06
Changes to address some annoying issues with cvar/cmd subsystems.

Paril:
- Get rid of the extraneous "keys" state list, which is just ripe for us messing things up in the future. It was unnecessary since hash tables internally keep their own list of keys, which we could query with g_hash_table_get_keys, but instead I went with iterators because they are a ton safer to iterate through.
- Add GlobiMatch, which is case-insensitive GlobMatch
- Make cvar and cmd hash tables case-insensitive and store a backing queue. Basically, this allows the system to store competing variables (such as "+zoom" and "+ZOOM") in the same entry in the hash table, and you can fetch the one you want from Cvar_Get. Cvar_Get is case-sensitive, to not break backwards compatibility, but internally case-insensitivity is used when executing commands, changing cvar values or for tab-completion. Adding new cvars or aliases is case-sensitive, for obvious reasons. This allows us to tab complete or even change variables with different casing. If you have competing casings, you must match that exact casing in order to change it. If there's only 1 command in the list, then it will work regardless of casing.
- Cvar uses GSList backing table and Cmd uses GQueue. Cvar never needs to remove at runtime, so GSList is safe - Cmd required GQueue so that it's easy to remove values and query the list's length without worrying about the list ptr disappearing because of removing it.
- Added copies of g_str_hash and g_str_equals which are case-insensitive.


src/cmd.c
```diff
@@ -354,39 +385,70 @@ static cmd_t *Cmd_Alias(const char *name, const char *commands) {
...
  * @brief Removes all commands which match the specified flags mask.
  */
 void Cmd_RemoveAll(uint32_t flags) {
-	g_hash_table_foreach_remove(cmd_state.commands, Cmd_RemoveAll_, &flags);
+	GHashTableIter iter;
+	gpointer key, value;
+	g_hash_table_iter_init(&iter, cmd_state.commands);
+
+	while (g_hash_table_iter_next (&iter, &key, &value)) {
+		GQueue *queue = (GQueue *) value;
```
jdolan: const?

src/cmd.c
```diff
@@ -354,39 +385,70 @@ static cmd_t *Cmd_Alias(const char *name, const char *commands) {
...
-	uint32_t flags = *((uint32_t *) data);
+void Cmd_Remove(const char *name) {
+	cmd_t *cmd = Cmd_Get_(name, true);
 
-	if (cmd->flags & flags) {
-		cmd_state.keys = g_list_remove(cmd_state.keys, key);
-		return true;
-	}
+	if (cmd) {
+		GQueue *queue = Cmd_Remove_(cmd);
```
jdolan: const?

src/console.h
```diff
@@ -206,4 +206,4 @@ void Con_AddConsole(const console_t *console);
 void Con_RemoveConsole(const console_t *console);
 void Con_AutocompleteInput_f(const uint32_t argi, GList **matches);
 void Con_Init(void);
-void Con_Shutdown(void);
+void Con_Shutdown(void);
```
jdolan: Missing newline?

src/cvar.c
```diff
@@ -53,16 +48,43 @@ static _Bool Cvar_InfoValidate(const char *s) {
 /**
  * @return The variable by the specified name, or `NULL`.
  */
-cvar_t *Cvar_Get(const char *name) {
+static cvar_t *Cvar_Get_(const char *name, const _Bool case_sensitive) {
+
+	if (cvar_vars) {
+		GSList *list = (GSList *) g_hash_table_lookup(cvar_vars, name);
```
jdolan: const

src/cvar.c
```diff
@@ -96,16 +118,15 @@ const char *Cvar_GetString(const char *name) {
...
-			func(var, data);
-		} else {
-			Com_Error(ERROR_FATAL, "Missing variable: %s\n", (char * ) key->data);
+	GHashTableIter iter;
+	gpointer key, value;
+	g_hash_table_iter_init(&iter, cvar_vars);
+
+	while (g_hash_table_iter_next (&iter, &key, &value)) {
+		
+		for (GSList *list = (GSList *) value; list; list = list->next) {
```
jdolan: const

src/shared.c
```diff
@@ -814,6 +822,50 @@ _Bool GlobMatch(const char *pattern, const char *text) {
...
+ * character hyphen character(as in 0-9 or A-Z).
+ * [0-9a-zA-Z_] is the set of characters allowed in C identifiers.
+ * Any other character in the pattern must be matched exactly.
+ *
+ * To suppress the special syntactic significance of any of `[]*?!-\',
+ * and match the character exactly, precede it with a `\'.
+ */
+_Bool GlobiMatch(const char *pattern, const char *text) {
+	return GlobMatch_(pattern, text, false);
+}
```
jdolan: Nice docs!

jdolan: My main critique of these functions (`Cmd_Get_`, `Cvar_Get_`, and `GlobMatch_`) is that taking in a _Bool in c has become a bit of a smell for me. I'm actually in the process of refactoring `Fs_Init` for this same smell, because I now have an additional flag to pass in.

Might it be better to define a bitmask of flags, and take in `uint32_t flags` instead? You could then `#define GLOB_CASE_INSENSITIVE` and potentially (in the future) some others (`GLOB_REGEX`?). What do you think?

src/shared.c
```diff
@@ -1421,3 +1473,31 @@ color_t ColorFromHSV(const vec3_t hsv) {
...
+g_stri_hash (gconstpointer v)
+{
+  const signed char *p;
+  guint32 h = 5381;
+
+  for (p = v; *p != '\0'; p++)
+    h = (h << 5) + h + tolower(*p);
+
+  return h;
+}
```
jdolan: Missing newline, braces missing / on wrong lines. `p` could be reduced in scope.

src/shared.c
Paril: ideally we should get rid of glob and use proper regexes, like from objectively or something, but yeah I can do that

src/shared.c
Paril: These were copied directly from glib so I didn't change their actual code other than tolower, but I can do that.

PR #404, opened 2017-02-09
Paril/teams o plenty

Paril:
NEEDS REVIEW AND TESTING

This is the branch that supports a variable number of (game-capped) teams, currently up to 4. The game is hardcoded to support Red, Blue, Green and Orange teams, and now also supports 4-team CTF (3tctf maps from 3tctf mod work, but ofc we have no 3/4 team maps ourselves).

Because of the limited amount of spawn points in the maps we do have, converting them to support 3/4 teams will be nearly impossible, so the automated creation still only works with 2 teams.

The g_num_teams cvar controls number of teams - default is "default", which picks the proper number based on the map's spawn points, or 2 as a fallback which then auto-generates flags/points.

This will also include a branch in quetoo-data to handle the transition of team1<->team2 (red/blue issue) and adding green/orange versions.

We need pants/shirt colors for proper coloring down the road, since there's no orange skin for q2guard.

Branch also includes other minor fixes, like adding weapon_chaingun support.

TODO: EF_CTF_x needs better effects, something distinct to a shell
Tweak spawn point particle to not be as annoying


kaadmy:
Does this also happen to fix the teams/spectator issue with the scoreboard?

Paril:
Yes @kaadmy

Paril:
Should be finished now, @jdolan final blessing for merge time

src/cgame/cgame.h
```diff
@@ -595,7 +595,7 @@ typedef struct cg_import_s {
 	 * @return The linked entity.
 	 * @remarks The parent entity must have been previously returned by `AddEntity`.
 	 */
-	const r_entity_t *(*AddLinkedEntity)(const r_entity_t *parent, const r_model_t *model, const char *tag_name);
+	r_entity_t *(*AddLinkedEntity)(const r_entity_t *parent, const r_model_t *model, const char *tag_name);
```
jdolan: So the cgame can mess with an entity that's already been copied into the renderer? Is this to allow tag alignment in cgame?

src/cgame/default/cg_entity.c
```diff
@@ -200,12 +205,21 @@ static void Cg_AddClientEntity(cl_entity_t *ent, r_entity_t *e) {
 
 	if (s->model2) {
 		r_model_t *model = cgi.client->model_precache[s->model2];
-		cgi.AddLinkedEntity(head.parent, model, "tag_weapon");
+		r_entity_t *m2 = cgi.AddLinkedEntity(head.parent, model, "tag_weapon");
+
+		if (self_draw) {
+			m2->effects |= EF_NO_DRAW;
+		}
```
jdolan: Hah, good catch!

src/cgame/default/cg_entity_effect.c
```diff
@@ -109,8 +118,17 @@ void Cg_EntityEffects(cl_entity_t *ent, r_entity_t *e) {
...
+		r_light_t l = { { 0.0, 0.0, 0.0 }, { 0.3, 1.0, 0.3 }, 80.0 };
+
+		VectorCopy(e->origin, l.origin);
+		cgi.AddLight(&l);
+
+		VectorMA(e->shell, 0.5, l.color, e->shell);
+	}
+
+	if (e->effects & EF_CTF_ORANGE) {
+		r_light_t l = { { 0.0, 0.0, 0.0 }, { 1.0, 0.7, 0.1 }, 80.0 };
```
jdolan: Nice.

src/cgame/default/cg_entity_trail.c
```diff
@@ -921,16 +951,10 @@ void Cg_EntityTrail(cl_entity_t *ent) {
...
-		case TRAIL_GOOD_SPAWN:
-			Cg_TeleporterTrail(ent, EFFECT_COLOR_BLUE, false);
-			break;
-		case TRAIL_EVIL_SPAWN:
-			Cg_TeleporterTrail(ent, EFFECT_COLOR_RED, false);
+			Cg_TeleporterTrail(ent, ColorFromRGB(255, 255, 211));
 			break;
-		case TRAIL_NEUTRAL_SPAWN:
-			Cg_TeleporterTrail(ent, EFFECT_COLOR_WHITE, false);
+		case TRAIL_SPAWN_POINT:
```
jdolan: `TRAIL_PLAYER_SPAWN` maybe?

src/cgame/default/cg_hud.c
```diff
@@ -512,29 +512,22 @@ static void Cg_DrawReady(const player_state_t *ps) {
...
-		return;
-	}
+	color_t color = cg_team_info[team].color;
+	color.a = 38;
 
 	x = cgi.view->viewport.x;
 	y = cgi.view->viewport.y + cgi.view->viewport.h - 64;
 
-	cgi.DrawFill(x, y, cgi.view->viewport.w, 64, color, 0.15);
+	cgi.DrawFill(x, y, cgi.view->viewport.w, 64, color.u32, -1.0);
```
jdolan: Nice.

src/cgame/default/cg_main.c
```diff
@@ -305,6 +305,34 @@ static void Cg_ClearState(void) {
...
+		
+		g_strlcpy(team->team_name, info[i], sizeof(team->team_name));
+		team->hue = (int16_t) atoi(info[i + 1]);
+		team->color = ColorFromHSV((const vec3_t) {
+			team->hue,
+			1.0,
+			1.0
+		});
+	}
+}
```
jdolan: Nice!

src/cgame/default/cg_main.c
```diff
@@ -320,6 +348,9 @@ static void Cg_UpdateConfigString(uint16_t i) {
 		case CS_HOOK_PULL_SPEED:
 			cg_state.hook_pull_speed = strtof(s, NULL);
 			return;
+		case CS_TEAM_INFO:
+			Cg_ResolveTeamInfo(s);
+			return;
```
jdolan: I dig this.

src/cgame/default/cg_types.h
```diff
@@ -96,4 +96,15 @@ typedef struct cg_particles_s {
...
+/**
+ * @brief Stores info related to teams on the server.
+ */
+typedef struct {
+	char team_name[MAX_USER_INFO_KEY];
+	int16_t hue;
+	color_t color;
+} cg_team_info_t;
+
+extern cg_team_info_t cg_team_info[TEAM_TOTAL];
```
jdolan: `MAX_TEAMS` I think would follow the others?

src/client/cl_binds.h
```diff
@@ -127,4 +125,8 @@ static const char *DEFAULT_BINDS =
     "exec quetoo.cfg\n"
 
     // and "default" autoexec.cfg
-    "exec autoexec.cfg\n";
+    "exec autoexec.cfg\n"
+
+	// bind these last in case somebody is using a Quake 2 config
+    "bind ` cl_toggle_console\n"
+    "bind f8 cl_toggle_console\n";
```
jdolan: Oh smart.

src/client/cl_main.c
```diff
@@ -535,6 +536,7 @@ static void Cl_InitLocal(void) {
 	// register our variables
 	cl_chat_sound = Cvar_Add("cl_chat_sound", "misc/chat", 0, NULL);
 	cl_draw_counters = Cvar_Add("cl_draw_counters", "1", CVAR_ARCHIVE, NULL);
+	cl_draw_position = Cvar_Add("cl_draw_position", "0", CVAR_ARCHIVE, NULL);
```
jdolan: Is this a developer tool? `CVAR_LO_ONLY` and not `CVAR_ARCHIVE`?

src/game/default/g_ballistics.c
```diff
@@ -1208,6 +1208,7 @@ static void G_HookTrail_Think(g_entity_t *ent) {
 	}
 
 	ent->locals.next_think = g_level.time + 1;
+	gi.LinkEntity(ent);
```
jdolan: Oh wow, was this a bug?

src/game/default/g_client.c
```diff
@@ -1116,15 +1120,24 @@ void G_ClientUserInfoChanged(g_entity_t *ent, const char *user_info) {
...
+		s = GetUserInfo(user_info, "skin");
+
+		char *p;
+
+		if (strlen(s) && (p = strchr(s, '/'))) {
+			*p = 0;
+			s = va("%s/%s", s, cl->locals.persistent.team->skin);
+		} else {
+			s = va("%s/%s", DEFAULT_USER_MODEL, cl->locals.persistent.team->skin);
+		}
```
jdolan: Sweet! Does this allow teams to set model but force colored skin?

src/game/default/g_client_stats.c
```diff
@@ -278,18 +269,14 @@ void G_ClientStats(g_entity_t *ent) {
...
-		if (client->locals.persistent.team == &g_team_good) {
-			client->ps.stats[STAT_TEAM] = CS_TEAM_GOOD;
-		} else {
-			client->ps.stats[STAT_TEAM] = CS_TEAM_EVIL;
-		}
+	if ((g_level.teams || g_level.ctf) && client->locals.persistent.team) { // send team ID, -1 is no team
+		client->ps.stats[STAT_TEAM] = client->locals.persistent.team->id;
 	} else {
-		client->ps.stats[STAT_TEAM] = 0;
+		client->ps.stats[STAT_TEAM] = -1;
```
jdolan: Okay, so red team is `0` because of the config string index?

src/game/default/g_cmd.c
```diff
@@ -728,6 +722,7 @@ static const char *vote_cmds[] = {
 	"g_rounds",
 	"g_spawn_farthest",
 	"g_teams",
+	"g_num_teams",
```
jdolan: I see no reason to not keep `g_teams` but let folks set it to `1` or `2` for 2 teams, and `3` or `4` for 3 and 4 respectively. `g_num_teams 2` is just less obvious than `g_teams 1` for the 99% use case.

src/game/default/g_entity.c
```diff
@@ -71,6 +133,9 @@ static g_entity_spawn_t g_entity_spawns[] = { // entity class names -> spawn fun
 
 	{ "worldspawn", G_worldspawn },
 
+	// compatibility-only entities
+	{ "weapon_chaingun", G_weapon_chaingun },
+
```
jdolan: Why isn't it adequate to simply map `weapon_chaingun` to `G_weapon_lightning` ? We'll never have a map that contains both. Then the G_weapon_chaingun functions above just go away.

src/game/default/g_entity.c
```diff
@@ -118,7 +183,7 @@ static void G_SpawnEntity(g_entity_t *ent) {
 		}
 	}
 
-	gi.Debug("%s doesn't have a spawn function\n", ent->class_name);
+	gi.Warn("%s doesn't have a spawn function\n", ent->class_name);
```
jdolan: Yea, warn is probably good here.

src/game/default/g_entity.c
```diff
@@ -215,6 +280,7 @@ static const g_field_t fields[] = {
 	{ "gameplay", SOFS(gameplay), F_STRING, FFL_SPAWN_TEMP },
 	{ "hook", SOFS(hook), F_STRING, FFL_SPAWN_TEMP },
 	{ "teams", SOFS(teams), F_STRING, FFL_SPAWN_TEMP },
+	{ "num_teams", SOFS(num_teams), F_STRING, FFL_SPAWN_TEMP },
```
jdolan: If we revert g_num_teams to g_teams, this should be reverted too.

jdolan: Oh, and we should make sure that the `maps.list` parser supports multi-team support too. That way, we can add 3-team CTF maps to a server and specify `teams 3` there.

src/game/default/g_types.h
```diff
@@ -242,10 +241,14 @@ typedef enum {
...
 #define EF_RESPAWN			(EF_GAME << 2) // yellow shell
 #define EF_QUAD				(EF_GAME << 3) // green shell
-#define EF_CTF_BLUE			(EF_GAME << 4) // blue shell
-#define EF_CTF_RED			(EF_GAME << 5) // red shell
-#define EF_DESPAWN			(EF_GAME << 6) // translucent
-#define EF_LIGHT			(EF_GAME << 7) // colored light
+#define EF_CTF_RED			(EF_GAME << 4) // blue shell
+#define EF_CTF_BLUE			(EF_GAME << 5) // red shell
+#define EF_CTF_GREEN		(EF_GAME << 6) // red shell
+#define EF_CTF_ORANGE		(EF_GAME << 7) // red shell
```
jdolan: Stale comments

src/game/default/g_types.h
```diff
@@ -482,8 +477,6 @@ typedef enum {
  * @brief Item types.
  */
 typedef enum {
-	ITEM_NONE,
-
```
jdolan: Hm. Are you sure we don't introduce ambiguity in removing these? Something with "item 0" is now ammo?

src/game/default/g_util.h
```diff
@@ -39,11 +39,11 @@ void G_SetMoveDir(vec3_t angles, vec3_t movedir);
 char *G_GameplayName(int32_t g);
 g_gameplay_t G_GameplayByName(const char *c);
 g_team_t *G_TeamByName(const char *c);
-g_team_t *G_OtherTeam(g_team_t *t);
-g_team_t *G_TeamForFlag(g_entity_t *ent);
-g_entity_t *G_FlagForTeam(g_team_t *t);
-uint32_t G_EffectForTeam(g_team_t *t);
-size_t G_TeamSize(g_team_t *team);
+const g_item_t *G_HoldingFlagTeam(const g_entity_t *ent);
```
jdolan: This name is weird. Can't tell what it does.

src/cgame/cgame.h
Paril: Basically, yes. I can see if I can work around this if you like, but the issue was mainly that AddLinkedEntity didn't give us a chance to set up the matrix properly. If I moved AddLinkedEntity to the cgame it might be possible to return these to const.

src/client/cl_main.c
Paril: I made it archive to fit the other cl_draw_* counters, but yeah I only used it to debug map issues so it can safely be made LO I suppose

src/game/default/g_ballistics.c
Paril: Yep, whoops

src/game/default/g_client.c
Paril: Yep!

src/game/default/g_client_stats.c
Paril: Yeah basically. I used -1 to indicate none, sort of like what pickup item ID does.

src/game/default/g_cmd.c
Paril: g_teams 2 is already in-use for team balancing, which is why I made a new cvar.

src/game/default/g_entity.c
Paril: Because the map won't have any bolts - the function above picks the nearby bullets and turns them into bolts.

Also there isn't a G_weapon_lightning since it's an item which has a different spawn routine.

src/game/default/g_entity.c
Paril: Pretty sure I did fix the map list parser to support it. If the map has either 3 flags or 3 team spawns, it'll just work automatically anyways, but yeah it does pick it up (g_map_list.c line 128)

src/game/default/g_types.h
Paril: I ran into a little issue and realized that we were never even using the tag 0 for anything special, so it seemed wasteful to waste that index. While I was doing this to the tags, we also weren't using ITEM_NONE so I removed it as well - it would never be able to happen in-game so it seemed weird to have an enum for an impossibility.

If you think it's ambiguous I could change it, but when there's only one path possible (type > 0) then why have an enum entry for 0?

src/game/default/g_util.h
Paril: Changed to "G_IsFlagBearer".

src/client/cl_main.c
jdolan: Honestly, we could include this in `cl_show_render_stats` or something?

src/cgame/cgame.h
jdolan: Oh right :)

src/game/default/g_cmd.c
jdolan: Ohh gotcha. Okay then.

src/game/default/g_entity.c
jdolan: I see.

jdolan: Good call on bullets -> bolts.

src/game/default/g_types.h
jdolan: So that uninitialized members don't appear to be real things. But if there are no side effects from this, I'm okay with that.

src/game/default/g_entity.c
Paril: I think we should make like a g_q2_compat.c which exists solely to satisfy compatibility with Quake 2 maps, in the case that we need to do so. That way this stuff isn't just cluttering other Quetoo-related files.

src/game/default/g_types.h
Paril: Gotcha. So here's my rationale:

g_item_t only exists, to both the people consuming them -and- developing them, as const members. The only instances of g_item_t are in g_itemlist, which is local to g_item.c and no instances ever escape that file, only *pointers* to instance. Unless a developer purposefully sets type to uninitialized (which in this case would be ammo, yeah), which would be really hard to do since they'd likely start by copypastaing other instances in the item list, there's absolutely no way that ITEM_NONE could ever be seen in regular gameplay *unless* there was a bug causing a g_item_t pointer to point outside of bounds, which again is.. extremely hard to do given the tools we have.

If you think we should have ITEM_NONE just for safety I can still do that, but there's like 8 other members there that will have to be filled with valid info before an item is "valid" anyways (name or classname for instance), so I don't think we're losing much by having an extra :)

PR #409, opened 2017-02-11
Paril/weapon select

Paril:
[pics.zip](https://github.com/jdolan/quetoo/files/768907/pics.zip)

Two new pics for this ^^^^^^^^^^^^^

### DIFF
--- a/src/cgame/default/viewcontrollers/KeysViewController.c
+++ b/src/cgame/default/viewcontrollers/KeysViewController.c
@@ -104,8 +104,8 @@ static void loadView(ViewController *self) {
 
 			Cg_BindInput((View *) stackView, "Attack", "+attack");
 			Cg_BindInput((View *) stackView, "Grapple Hook", "+hook");
-			Cg_BindInput((View *) stackView, "Next weapon", "weapon_next");
-			Cg_BindInput((View *) stackView, "Previous weapon", "weapon_previous");
+			Cg_BindInput((View *) stackView, "Next weapon", "cg_weapon_next");
+			Cg_BindInput((View *) stackView, "Previous weapon", "cg_weapon_previous");
 			Cg_BindInput((View *) stackView, "Zoom", "+ZOOM");
 
 			Cg_BindInput((View *) stackView, "Blaster", "use blaster");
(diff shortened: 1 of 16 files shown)


src/cgame/default/viewcontrollers/KeysViewController.c
```diff
@@ -104,8 +104,8 @@ static void loadView(ViewController *self) {
 
 			Cg_BindInput((View *) stackView, "Attack", "+attack");
 			Cg_BindInput((View *) stackView, "Grapple Hook", "+hook");
-			Cg_BindInput((View *) stackView, "Next weapon", "weapon_next");
-			Cg_BindInput((View *) stackView, "Previous weapon", "weapon_previous");
+			Cg_BindInput((View *) stackView, "Next weapon", "cg_weapon_next");
+			Cg_BindInput((View *) stackView, "Previous weapon", "cg_weapon_previous");
```
jdolan: So you no longer need to cycle through on the server.. this means the delay is enforced client-side, right? An easy cheat would then be to remove the delay client-side, yea?

Not a show-stopper for me, just want to fully understand.

src/cgame/default/views/PlayerModelView.c
```diff
@@ -274,6 +274,8 @@ static void animate(PlayerModelView *self) {
 	VectorClear(self->head.angles);
 	VectorClear(self->weapon.angles);
 
+	self->legs.scale = self->torso.scale = self->head.scale = self->weapon.scale = 1.0;
+
```
jdolan: LOL

src/quetoo.h
```diff
@@ -421,17 +421,19 @@ typedef struct {
...
+	BUTTON_STATE_DOWN	= (1 << 1),
+	BUTTON_STATE_UP		= (1 << 2)
+} button_state_t;
+
 typedef struct {
 	uint32_t keys[2]; // keys holding it down
 	uint32_t down_time; // msec timestamp
 	uint32_t msec; // msec down this frame
-	byte state;
+	button_state_t state;
```
jdolan: 👍

src/cgame/default/viewcontrollers/KeysViewController.c
Paril: Delay is still there because it makes switching to weapons faster, actually. The reason is simple: you don't want to switch between 6 different weapons to get the one you need. That's why there's a short delay between scrolling and when it actually switches. However, you can bypass this: clicking will force the weapon you're pointing at to switch to it immediately.

PR #415, opened 2017-02-12
Issue/63 quemap materials

jdolan:
This adds full material awareness to Quemap.

 * Materials are loaded from the map source, rather than the bsp result. This enables materials to influence the BSP compilation phase, and allows mappers to generate materials files without first compiling the map.

 * `quemap -mat` now authors a complete, correct materials file, preserving any previously defined materials and stages as well. Mappers don't have to worry about losing their materials by re-running the `-mat` phase. This will come in very handy for the in-game materials editor.

 * Materials can inform the BSP compilation phase with contents, surface and light parameters. As an example:

```
    {
      material tokays/water
      contents water
      surface "blend_33 light warp"
      light 400
    }
    {
      material torn/floor_fx
      contents mist
      surface material
      {
        texture torn/floorfx_fx
        blend GL_SRC_ALPHA GL_ONE
        stretch 0.5 1.0
        pulse 1.0
        color 0.3 0.3 0.3
      }
    }
```

### DIFF
--- a/src/tools/quemap/map.c
+++ b/src/tools/quemap/map.c
@@ -20,6 +20,7 @@
  */
 
 #include "qbsp.h"
+#include "qmat.h"
 #include "scriptlib.h"
 
 int32_t num_map_brushes;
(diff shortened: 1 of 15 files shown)


src/tools/quemap/map.c
```diff
@@ -468,29 +469,49 @@ static _Bool MakeBrushWindings(map_brush_t *ob) {
 /**
  * @brief
  */
-static void SetImpliedFlags(side_t *side, const char *tex) {
+static void SetMaterialFlags(side_t *side, map_brush_texture_t *td) {
 
-	if (!g_strcmp0(tex, "common/areaportal")) {
+	const cm_material_t *material = LoadMaterial(td->name);
```
jdolan: Here's where the magic happens.

src/tools/quemap/map.c
```diff
@@ -869,6 +889,8 @@ void LoadMapFile(const char *file_name) {
 	memset(map_planes, 0, sizeof(map_plane_t) * MAX_BSP_PLANES);
 	num_map_planes = 0;
 
+	memset(plane_hash, 0, sizeof(plane_hash));
```
jdolan: This fixes multiple invocations of `LoadMapFile` hanging infinitely.

src/tools/quemap/qbsp.c
```diff
@@ -354,12 +348,23 @@ int32_t BSP_Main(void) {
 		WriteBSPFile(bsp_name, version);
 	} else {
 
+		// delete portal and line files
```
jdolan: Only nuke the `.prt` and `.lin` when we are going to re-create them.

src/tools/quemap/qbsp.c
```diff
@@ -354,12 +348,23 @@ int32_t BSP_Main(void) {
...
+
 		CreateBSPFile();
-		// start from scratch
+
 		LoadMapFile(map_name);
 		SetModelNumbers();
 
 		ProcessModels();
+
+		FreeMaterials();
```
jdolan: Load and free the materials so that they're available to `LoadMapFile`.

src/tools/quemap/qmat.c
```diff
@@ -21,77 +21,32 @@
...
 
 	for (int32_t i = 0; i < bsp_file.num_texinfo; i++) {
 		LoadMaterial(bsp_file.texinfo[i].texture);
 	}
 
-	Cm_WriteMaterials(mat_name, (const cm_material_t **) materials->pdata, materials->len);
+	UnloadScriptFiles();
 
-	Com_Print("Generated %d materials\n", materials->len);
+	WriteMaterialsFile(mat_name);
```
jdolan: This is now nice and clean.

kaadmy:
I assume the `contents` can be repeated like so:
```
contents mist
contents lava
```

jdolan:
@kaadmy No, `contents` should be combined into a single token (quoted string), like how `surface` is denoted in the PR comment above.

src/tools/quemap/map.c
Paril: In a way it kind of sucks that we can't also have explicit .mat files for single textures like Q3 does, it'd make the top bit cleaner if it could just load them from, say, common/origin.mat for instance. Oh well.

PR #418, opened 2017-02-14
The better version of my terrible commit.

Paril:
- Autocompletes now have two components: "name" and "description". "name" is the keyname of the object to complete, and "description" is what can be printed to the console in the case where there is enough room for a lot of info to be displayed.
- A few helper functions to help write autocomplete functions (Com_MatchCompare and Com_AllocMatch).
- All autocomplete matches are sorted alphabetically now.
- Autocompletes now only print in the console function, instead of Com_Printing from cvar/cmd/etc. This allows autocomplete to columnize if possible. Columns only appear if there's more than 1 row (and there are no descriptions available) and more than 1 match.
- File system autocomplete compare uses stricmp on Windows and strcmp on Unix. This fixes issue of differently-cased files appearing on Windows even though Windows doesn't care.
- CommonPrefix moved to console since it was the only implementor and to use new match struct

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -163,6 +163,24 @@ void Com_Warnv_(const char *func, const char *fmt, va_list args) __attribute__((
 #define Com_Warn(...) Com_Warn_(__func__, __VA_ARGS__)
 #define Com_Warnv(fmt, args) Com_Warnv_(__func__, fmt, args)
 
+/**
+ * @brief The structure used for autocomplete values.
+ */
+typedef struct {
+	/**
+	 * @brief The match itself
+	 */
+	char *name;
+
+	/**
+	 * @brief The value printed to the screen. If null, name isused.
+	 */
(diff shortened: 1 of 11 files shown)


src/common.h
```diff
@@ -163,6 +163,24 @@ void Com_Warnv_(const char *func, const char *fmt, va_list args) __attribute__((
...
+	/**
+	 * @brief The match itself
+	 */
+	char *name;
+
+	/**
+	 * @brief The value printed to the screen. If null, name isused.
+	 */
+	char *description;
+} com_autocomplete_match_t;
```
jdolan: You can probably remove the `com_` prefixes here if you want.

Paril: Aren't structs generally named to match their subsystem?

jdolan: They are, .. except common and shared, lol.

PR #426, opened 2017-02-17
In-game materials editor (code review)

jdolan:


src/client/ui/viewcontrollers/EditorViewController.c
```diff
@@ -0,0 +1,329 @@
...
+#pragma mark - ViewController
+
+/**
+ * @see View::updateBindings(View *)
+ */
+static void updateBindings(View *self) {
+
+	EditorViewController *this = (EditorViewController *) self;
+
+	if (!this->bumpSlider) { // we probably haven't run loadView yet
```
jdolan: Shouldn't be possible. If this *is* happening, there might be another lifecycle thing going on.

src/client/ui/viewcontrollers/EditorViewController.c
```diff
@@ -0,0 +1,329 @@
...
+	} else {
+		this->material = NULL;
+	}
+
+	if (this->material) {
+		$(this->bumpSlider, setValue, (double) this->material->cm->bump);
+		$(this->hardnessSlider, setValue, (double) this->material->cm->hardness);
+		$(this->specularSlider, setValue, (double) this->material->cm->specular);
+		$(this->parallaxSlider, setValue, (double) this->material->cm->parallax);
+	}
```
jdolan: 👍

src/client/ui/viewcontrollers/EditorViewController.c
jdolan: Oh. I just noticed you're trying to implement `updateBindings` on your ViewController. `updateBindings` is a View method.

jdolan: Perhaps a better way to handle this would be by overriding `respondToEvent` -- or maybe we should modify ViewController so that it also has some lifecycle management. I'll think on this a little today.

src/client/ui/viewcontrollers/EditorViewController.c
```diff
@@ -0,0 +1,329 @@
...
+			Box *box = $(alloc(Box), initWithFrame, NULL);
+			$(box->label, setText, "MATERIAL");
+
+			StackView *stackView = $(alloc(StackView), initWithFrame, NULL);
+
+			// bump
+
+			this->bumpSlider = $(alloc(Slider), initWithFrame, NULL, ControlStyleDefault);
+
+			this->bumpSlider->delegate.self = (ident *) this;
```
jdolan: Cast shouldn't be necessary.

src/client/ui/viewcontrollers/EditorViewController.c
```diff
@@ -0,0 +1,329 @@
...
+
+			$((View *) box, addSubview, (View *) stackView);
+			release(stackView);
+
+			$((View *) column, addSubview, (View *) box);
+			release(box);
+		}
+
+		$((View *) columns, addSubview, (View *) column);
+		release(column);
```
jdolan: 👍  Nicely done!

src/client/ui/viewcontrollers/EditorViewController.c
```diff
@@ -0,0 +1,329 @@
...
+#pragma mark - Class lifecycle
+
+/**
+ * @see Class::initialize(Class *)
+ */
+static void initialize(Class *clazz) {
+
+	((ViewControllerInterface *) clazz->def->interface)->loadView = loadView;
+
+	((ViewInterface *) clazz->def->interface)->updateBindings = updateBindings;
```
jdolan: Yea, VC is not a View. This is going to explode :) No idea what method you're actually trodding on by casting to ViewInterface here, but it's not good.

PR #433, opened 2017-02-19
Paril/parse

Paril:


src/tools/quemap/qzip.c
```diff
@@ -161,91 +161,117 @@ static void AddMaterials(const char *path) {
...
-			if (*c == '#') {
-				g_strlcpy(texture, ++c, sizeof(texture));
+		if (!g_strcmp0(token, "texture") || !g_strcmp0(token, "diffuse")) {
+
+			if (!Parse_Token(&parser, PARSE_DEFAULT | PARSE_NO_WRAP, token, sizeof(token))) {
+				return;
+			}
+
+			if (*token == '#') {
+				g_strlcpy(texture, token + 1, sizeof(texture));
```
jdolan: Heh, I have another ticket pending to completely rewrite this to use Cm_LoadMaterials.. but this is fine.

Paril:
More options often means more verbose, especially now that it stores some state instead of taking a raw pointer :)

src/tools/quemap/qzip.c
Paril: I couldn't leave any of them untouched since ParseToken is gone

Issue #419, opened 2017-02-15
Chasecam switches sometimes cause item pickup flashes


Paril:

jdolan:
Tagging to 1.0 because, while minor, this seems like a quick fix.

jdolan:
Uh, is it possible that the target you just switched to did in fact just pickup an item?

Paril:
Yes and no. They have to pick up an item, then you have to switch away then back. For a few seconds after they pick up an item, every switch to them will cause a screen flash.

kaadmy:
Isn't this fixed?

PR #439, opened 2017-02-21
Issue/419 chasecam target change

jdolan:
This fixes odd lerping when changing chase targets, and probably other
bugs.

### DIFF
--- a/src/cgame/default/cg_hud.c
+++ b/src/cgame/default/cg_hud.c
@@ -63,31 +63,39 @@ typedef struct {
 	int16_t item_index;
 } cg_hud_weapon_t;
 
-typedef struct {
-	// pickup
-	uint32_t last_pulse_time;
-	int16_t pickup_pulse;
-
-	// damage
-	uint32_t last_hit_sound_time;
-
-	// blend
-	uint32_t last_pickup_time;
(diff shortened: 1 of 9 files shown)


src/cgame/default/cg_hud.c
```diff
@@ -446,6 +454,12 @@ static void Cg_DrawChase(const player_state_t *ps) {
 	r_pixel_t x, y, ch;
 	char string[MAX_USER_INFO_VALUE * 2], *s;
 
+	// if we've changed chase targets, reset all locals
+	if (ps->stats[STAT_CHASE] != cg_hud_locals.chase_target) {
```
jdolan: Here's the money shot for #419

src/game/default/g_client_view.c
```diff
@@ -417,4 +417,19 @@ void G_EndClientFrames(void) {
...
+
+		if (!ent->in_use || !ent->client) {
+			continue;
+		}
+
+		if (ent->client->locals.chase_target) {
+			G_ClientChaseThink(ent);
+			G_ClientSpectatorStats(ent);
+		}
+	}
```
jdolan: And here's the fix for weird lerping when changing chase targets.

PR #449, opened 2017-03-09
Indirect static lighting

jdolan:


src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,89 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+	// get surface color
+
+	if (trace.surface) {
+		reflectiveness = Clamp(trace.surface->material->hardness / 5.0, 0.0, 1.0); // 5:1
+
+		GetTextureReflectivity(trace.surface->material->base, color);
+	} else {
+		reflectiveness = 0.2;
+
+		VectorClear(color);
```
jdolan: I think you want to reflect white here (1.0, 1.0, 1.0).

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,89 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+	vec_t bounce_dist;
+	cm_trace_t trace;
+
+	// get a trace
+
+	VectorScale(normal, dist, delta);
+
+	Light_Trace(&trace, delta, pos, CONTENTS_SOLID);
+
+	if (trace.fraction >= 1.0) {
```
jdolan: `== 1.0` would be more standard.

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,89 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+	VectorMA(bounce_pos, SAMPLE_NUDGE, trace.plane.normal, bounce_pos);
+
+	/// length of the next bounce
+
+	bounce_dist = Clamp(reflectiveness * sqdist * INDIRECT_BOUNCE_DIST, INDIRECT_BOUNCE_MIN_DIST, dist);
+
+	// bounce again
+
+	GatherSampleBounceTrace(bounce, center, bounce_pos, bounce_dir, bounce_dist, sample);
+}
```
jdolan: Love it!

src/tools/quemap/qlight.c
```diff
@@ -83,6 +85,34 @@ _Bool Light_PointPVS(const vec3_t org, byte *pvs) {
...
+	const int32_t cluster1 = Cm_LeafCluster(leaf1);
+	const int32_t cluster2 = Cm_LeafCluster(leaf2);
+
+	Cm_ClusterPVS(cluster1, pvs);
+
+	if ((pvs[cluster2 >> 3] & (1 << (cluster2 & 7))) == 0) {
+		return false;
+	}
+
+	return true;
```
jdolan: Does `-light` in general work on non-VIS'ed maps? If so, it would be great if this worked on non-VIS'ed maps, too. It'd be slow, but I think it should still work.

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,87 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+		return;
+	}
+
+	if (trace.start_solid) {
+		return;
+	}
+
+	// distance squared and trace hit normal dot product
+
+	sqdist = trace.fraction * trace.fraction;
```
jdolan: This isn't the distance squared, it's the trace fraction squared. Subtract the trace start and end points, and take `VectorLengthSquared` of that if you want distance squared.

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,87 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+		hardness = 1.0;
+		VectorSet(color, 1.0, 1.0, 1.0);
+	}
+
+	const vec_t reflect = hardness / 5.0;
+	VectorScale(color, reflect * sqdist, color);
+
+	if ((trace.fraction * dist) < AMBIENT_OCCLUSION_DIST) {
+		vec_t ao = (1.0 - reflect) * (trace.fraction * (dist / AMBIENT_OCCLUSION_DIST)) * Max(0.001, dot);
+		color[0] -= ao;
```
jdolan: Is there a reference for this equation you could add in a comment?

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,87 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+	}
+
+	// multiplicative color
+
+	sample[0] *= color[0];
+	sample[1] *= color[1];
+	sample[2] *= color[2];
```
jdolan: This doesn't seem right to me. We should be adding reflective light, not multiplying (modulating) it.

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,87 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+	bounce++;
+
+	if (bounce >= indirect_bounces || bounce >= MAX_INDIRECT_BOUNCES) {
+		return; // don't bounce anymore
+	}
+
+	// direction of the next bounce
+
+	// FIXME: Is this right? Shouldn't we use a DotProduct here ala Pm_ClipVelocity?
+	CrossProduct(normal, trace.plane.normal, bounce_dir);
```
jdolan: See comment. CrossProduct of the normals is almost certainly not what you want here:
https://en.wikipedia.org/wiki/Cross_product

I think we want to bounce the trace direction, just like how we clip to planes in Pm_ClipVelocity / G_ClipVelocity.

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,87 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+
+	// where the next bounce starts
+
+	VectorMA(trace.end, SAMPLE_NUDGE, bounce_dir, bounce_pos);
```
jdolan: I ripped out an attempt to emulate NudgeSamplePosition here. It was in correct. It overwrote trace.plane.normal, but then read from that vector, expecting it to be the original value. It also nudged the new start position towards the center of the previous surface, which doesn't make sense. I think nudging out along the new surface normal should be adequate. If that puts us into a solid, the next trace will fail fast.

src/tools/quemap/lightmap.c
```diff
@@ -620,6 +620,87 @@ static _Bool NudgeSamplePosition(const vec3_t in, const vec3_t normal, const vec
...
+
+	// length of the next bounce
+
+	bounce_dist = Clamp(reflect * sqdist * INDIRECT_BOUNCE_DIST, INDIRECT_BOUNCE_MIN_DIST, dist);
```
jdolan: This isn't right, either. You need to factor in the incoming distance (which is really the amount of light we have).

src/tools/quemap/qlight.c
kaadmy: It doesn't even work since it requires a vis to exist otherwise the `-light` stage crashes.

src/tools/quemap/lightmap.c
kaadmy: I don't know why I did this.

src/tools/quemap/lightmap.c
kaadmy: I do, this modifies a fullbright sample (255, 255, 255) then afterwards adds it to the pixel result.

src/tools/quemap/qlight.c
jdolan: Hah! Well we should catch that ;)

src/tools/quemap/lightmap.c
jdolan: I'm still not sure that's "correct." To do this accurately, we want to keep track of the amount of light that reaches the face. Then mix (multiply?) the light color with the impacted surface color, scale it by the hardness, and reflect accordingly. If we do it right, green surfaces will reflect green light really well, but will absorb red light. That's realistic. That way, the light will naturally decay with each successive bounce, too.

jdolan: Btw the amount of light that reaches the face is the max trace distance times the trace fraction. In other words, if you had 128 units of light to trace, and you impacted a surface at 96 units away, you'd have 32 units of light at the impacted point.

jdolan: You can see how GatherSampleLight accumulates light additively, as integers.

src/tools/quemap/lightmap.c
kaadmy: Basically interpolates from `0` to `1`, where zero is longest distance, least material hardness, and flattest angle, and one is shortest distance, highest material hardness, and trace angle perpendicular to the face normal it hit.

src/tools/quemap/lightmap.c
kaadmy: I'm using this method mainly because for each sample (Think -light -extra = 5 samples per lxuel) has X traces (8 right now I think), they have to be divided down so they only affect the final result by the same amount as 1 sample and 1 trace, ints might not work too well here.

jdolan:
@kaadmy Are you interested in finishing this feature, or are you tapping out? :)

kaadmy:
I want to finish it, but I also want to get all the 1.0 issues ironed out first ;)

kaadmy:
What's the current state of this branch?

jdolan:
The state is.. it's in a very good place to actually do indirect lighting as a separate pass, but that separate pass is broken / incorrect at the moment. So some work is needed to fix the indirect lighting sample routine, and then we can look at how much this changes our existing maps, which ones need / should use indirect lighting, etc..

kaadmy:
Ideally we'd have to rebake lighting on all of our maps if we add a new light pass, since anybody else later   might rebake it and add in indirect lighting by accident or something. It's probably better to make everything consistent whenever possible.
I can look into this a bit later when I have time.

kaadmy:
I really like the conflict in this branch :P

jdolan:
Heh. Feel free to take either side.

PR #450, opened 2017-03-10
Add rim lighting

kaadmy:
Adds rim lighting as an effect (`g_entity_t->s.effects & EF_RIM` and `g_entity_t->s.rim_color[vec4]`), can be configured in game or cgame fairly easily, this is also a shader effect meaning it can be tuned very easily.
The CVar `r_rim` toggles this effect.

### DIFF
--- a/src/cgame/default/cg_client.c
+++ b/src/cgame/default/cg_client.c
@@ -226,10 +226,14 @@ void Cg_LoadClient(cl_client_info_t *ci, const char *s) {
 
 			// load the models
 			if (!Cg_LoadClientModel(ci, info[1], v + 1)) {
+				cgi.Warn("Failed to load client skin %s/%s\n",
+					info[1], v + 1);
 				if (!Cg_LoadClientModel(ci, info[1], DEFAULT_CLIENT_SKIN)) {
+					cgi.Warn("Failed to load client model %s/%s\n",
+						info[1], DEFAULT_CLIENT_SKIN);
 					if (!Cg_LoadClientModel(ci, DEFAULT_CLIENT_MODEL, DEFAULT_CLIENT_SKIN)) {
-						cgi.Error("Failed to load default client model " \
-							DEFAULT_CLIENT_MODEL "/" DEFAULT_CLIENT_SKIN "\n");
+						cgi.Error("Failed to load default client skin %s/%s\n",
+							DEFAULT_CLIENT_MODEL, DEFAULT_CLIENT_SKIN);
 					}
 				}
 			}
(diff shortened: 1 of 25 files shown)


jdolan:
Two things:

1) What is this? It looks cool, but it's adding a white haze on the items that have it.

quetoo012
quetoo013

2) It needs a better name. Really. Rim?

Got a link to an explanation of the effect?

kaadmy:
1. That's what rim lighting does, but maybe it needs tuning, it doesn't look too good right now.
2. http://lmgtfy.com/?q=rim+lighting

jdolan:
Okay, I think I found some references as to what the effect is supposed to look like.

I think for this to look right, you'd need to sharpen the outline lighting. So more intense effect over a thinner outline. Is that possible?

kaadmy:
Yep, very possible :D
Just multiply `rim` three times instead of two in `default_fs.glsl`, I did that at first but I thought it looked a bit too sharp.

WickedShell:
I'm deeply amused that @KaadmY is adding more intensive graphics effects....

jdolan:
So after playing with the exponent a bit, I'm not sure that's the ticket. Maybe it's more the rim effect color (and alpha) that needs tweaking. I also have some other concerns here:

 * Sending over a vec4 for the rim color is expensive. We already have issues with frame size. Might it be better to have "canned" effect colors that we can send over as an enum / byte?
 * This effect is the result of a strong light source being occluded by an object. While there might be some scenarios where this looks correct / natural in Quetoo, generally it will not.
 * My understanding is that you wanted this to replace the shell effect? Can you attach some screenshots of how this looks for Quad, CTF flag carriers, etc?

jdolan:
@kaadmy Could you cherry-pick the skin loading fix into master?

kaadmy:
Image of an older version with a hardcoded player rim light: https://cdn.discordapp.com/attachments/224935404506316800/289887771932360705/quetoo022.png
Maybe it won't work too well with the shell effects, but with some work we could get almost all effects on the model itself without a shell.

kaadmy:
Using a vec4 over network is pretty expensive and I already thought of that when writing that, using a single byte with a color index in a preselected color palette would be good enough for this.

jdolan:
The red effect looks pretty cool on Dragoon there.

I'm not fully sold on this feature. I think I can say with confidence that I don't really like it on pickup items. I could see it possibly replacing the shell effect for some scenarios, but I'm not sure it's necessary (?) @Panjoo or @Paril do you have an opinion on this?

jdolan:
I'd like to close this PR for now. I don't think we have a clear use case for the effect, and there are now conflicts with master. Seems you're focused on the menu branch as well, which is much higher priority.

PR #454, opened 2017-03-14
Make fog more DRY, move includes into a folder.

jdolan:
I’d like to make a pass over the shaders, pulling things into includes
where possible, and making things more functional as well (using inout
or return values instead of relying on globals). And, where warranted,
we can explore breaking things into separate programs.


jdolan:
Mind reviewing this one?

TenaciousTigerTrumpets:
Any chance you can sneak in a brightness cvar?
e.g.
`fragColor.rgb = diffuse.rgb * (lightmap + light) * BRIGHTNESS;`

r_modulate only really brightens the middle of the spectrum and clips lightmap highlights, giving a washed out look. I would prefer to keep the precision of the lightmaps with r_modulate set to 1 and set an r_brightness of 2 or 3. I've tried it and IMHO it just looks way better, the lightmaps are much more punchy.

Maybe brightness is a confusing term, since it only applies to the world lighting and nothing else. Perhaps  "lightscale" is a better term, since that's what it does.

jdolan:
I like that idea @ghost.

kaadmy:
What does `r_modulate_ do anyway?

jdolan:
r_modulate multiplies the lightmap samples, CPU-side, prior to uploading them to the video card. It also scales the BSP light sources, used for static lighting. I'm sure that all needs revisiting with the change @Paril added for `r_lightscale` -- which is fine. The proof is in the pudding, and the screenshots with `r_lightscale` look much better.

PR #455, opened 2017-03-14
Paril/stain changes

jdolan:


src/cgame/default/cg_temp_entity.c
```diff
@@ -298,8 +301,9 @@ void Cg_GibEffect(const vec3_t org, int32_t count) {
 
 	cgi.AddStain(&(const r_stain_t) {
 		.origin = { org[0], org[1], org[2] },
-		 .color = { 0.6 + 0.1 * Randomc(), 0.0, 0.0, 0.9 + Randomf() * 0.1 },
-		  .radius = count * 4.0
+		.radius = count * 4.0,
+		.image = cg_particles_normal->image,
+		.color = { 0.6 + 0.1 * Randomc(), 0.0, 0.0, 0.9 + Randomf() * 0.1 },
```
jdolan: Yea, `astyle` does this weird incremental indent thing on these initializers. Very strange.

src/client/cl_screen.c
```diff
@@ -387,6 +387,10 @@ void Cl_UpdateScreen(void) {
 		}
 	}
 
+	if (cls.state == CL_ACTIVE) {
+		R_AddStains();
+	}
+
```
jdolan: Eeeeeh. This is really ugly here. Is there a specific reason this can't be encapsulated in R_DrawView?

src/client/renderer/r_array.c
```diff
@@ -180,6 +182,12 @@ void R_UploadToSubBuffer(r_buffer_t *buffer, const size_t start, const size_t si
 		return;
 	}
 
+	// Don't allow null ptrs since bufferSubData does not allow it.
+	if (!data) {
+		Com_Warn("Attempted to upload null to GPU\n");
```
jdolan: Because users (and even modders) can't generate this warning (it's a programmer error), I think `Com_Warn` is awkward here. Maybe `Com_Debug` or even `Com_Error(ERROR_DROP, ..`?

src/client/renderer/r_framebuffer.c
```diff
@@ -0,0 +1,148 @@
...
+	R_GetError(NULL);
+}
+
+/**
+ * @brief See if the framebuffer is ready to be used.
+ */
+_Bool R_FramebufferReady(const r_framebuffer_t *fb) {
+	R_BindFramebuffer(fb);
+	return glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE;
+}
```
jdolan: 👍

src/client/renderer/r_main.c
```diff
@@ -598,14 +594,18 @@ void R_Init(void) {
 
 	R_InitConfig();
 
+	R_InitMedia();
+
 	R_InitState();
 
-	R_InitPrograms();
+	R_GetError("NULL");
```
jdolan: Something more fitting than "NULL" yea?

src/parse.c
```diff
@@ -412,13 +412,16 @@ static size_t Parse_TypeSize(const parse_type_t type) {
 		return 4;
 	case PARSE_DOUBLE:
 		return 8;
+	default:
+		signal(SIGSEGV, NULL);
```
jdolan: Fuck, really? :D

jdolan: I guess this will trigger a Com_Error and Sys_Backtrace without introducing a circular dependency there.. I dig it.

src/client/renderer/r_array.c
Paril: The problem is that as a programmer I didn't even notice it because of how specific the issue came up in. Since it's a potentially fatal error it should be something that catches their eyes immediately.

src/client/cl_screen.c
Paril: It can be moved up, but no it can't go in R_DrawView because this is not part of the view, this updates the stain framebuffers which are 2D (so Setup2D needs to be called first).

Issue #428, opened 2017-02-18
quemap -zip should use Cm_LoadMaterials


Dracant:
Failed to add _fx file called out in the .mat file

jdolan:
Can you attach the .mat?

jdolan:
A possible fix was pushed. We were previously not picking up materials defined with `diffuse` -- only `texture` was previously supported.

kaadmy:
Is this fixed or something?

Dracant:
Do I need to put the Cm_LoadMaterial in the command line when running the
pak generator?

jdolan:
The bug is fixed, yes @kaadmy. But there's refactoring to do here.

@Dracant No bud, you can just run quemap like you were before :)

kaadmy:
*Bump*

jdolan:
Bump urself pal! ;) I'll try to get to this over the weekend.

jdolan:
This is done.

PR #467, opened 2017-04-09
#428 material resolution

jdolan:
The main goal of this PR is to centralize material asset resolution, so that we have a single code path for gathering all of the file resources that comprise materials. This is important so that quemap and quetoo agree on which assets are required.

Along the way, I decided that mutable materials collections (GList) are really useful, so the generic material loading entry points now use those. The way these functions work now (taking in `GList **`), you can easily merge multiple materials files into a single collection, which might come in handy.

Another fix with this change is that now _every_ texinfo, in collision-land and renderer-land, now has a material, always. This was not the case with the previous implementation. And all of the materials used by a model (all formats) are now available on `r_model_t` directly.

This means that the cgame no longer requires a special entry point to lookup map materials, and the cgame works directly with the renderer's copies of materials, rather than the collision model's copies. This will allow us to edit footstep sounds and other cgame-dependent functionality in the in-game editor.

This is not ready yet, as the actual changes to quemap are not yet pushed. But I wanted to get your input on the changes overall as this shapes up.

### DIFF
--- a/src/cgame/default/cg_media.c
+++ b/src/cgame/default/cg_media.c
@@ -154,9 +154,9 @@ static void Cg_InitFootsteps(void) {
 
 	default_samples = g_array_append_vals(default_samples, (const s_sample_t *[]) {
 		cgi.LoadSample("#players/common/step_1"),
-		               cgi.LoadSample("#players/common/step_2"),
-		               cgi.LoadSample("#players/common/step_3"),
-		               cgi.LoadSample("#players/common/step_4")
+		cgi.LoadSample("#players/common/step_2"),
+		cgi.LoadSample("#players/common/step_3"),
+		cgi.LoadSample("#players/common/step_4")
 	}, 4);
 
 	g_hash_table_insert(cg_footstep_table, "default", default_samples);
--- a/src/cgame/default/cg_media.c
+++ b/src/cgame/default/cg_media.c
@@ -160,17 +160,12 @@ static void Cg_InitFootsteps(void) {
 	}, 4);
 
 	g_hash_table_insert(cg_footstep_table, "default", default_samples);
-	
-	size_t num_materials = 0;
-	const cm_material_t **materials = cgi.MapMaterials(&num_materials);
 
-	if (num_materials) {
-		for (size_t i = 0; i < num_materials; i++) {
-			const cm_material_t *material = materials[i];
+	r_material_t **material = cgi.WorldModel()->materials;
+	for (size_t i = 0; i < cgi.WorldModel()->num_materials; i++, material++) {
 
-			if (*material->footsteps) {
-				Cg_FootstepsTable_Load(material->footsteps);
-			}
+		if (strlen((*material)->cm->footsteps)) {
+			Cg_FootstepsTable_Load((*material)->cm->footsteps);
 		}
 	}
 }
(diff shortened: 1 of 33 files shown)


jdolan:
Oh, and we need to do some testing on this to make sure I'm not leaking memory :)

jdolan:
@Paril @kaadmy I think this is ready for review now.

I didn't modify the `CalcTextureReflectivity` routine in quemap because KaadmY's indirect lighting branch has other changes in there. But, once this is merged, I'd like to get *that* branch up to date and wrapped up, too.

src/cgame/default/cg_media.c
```diff
@@ -154,23 +154,18 @@ static void Cg_InitFootsteps(void) {
...
 
 	g_hash_table_insert(cg_footstep_table, "default", default_samples);
-	
-	size_t num_materials = 0;
-	const cm_material_t **materials = cgi.MapMaterials(&num_materials);
 
-	if (num_materials) {
-		for (size_t i = 0; i < num_materials; i++) {
-			const cm_material_t *material = materials[i];
+	r_material_t **material = cgi.WorldModel()->materials;
```
Paril: Curious, why do we pass the renderer-specific r_material_t to the cgame? It doesn't/shouldn't need to know about the abstraction, it only needs the cm_material_t pointers. Since r_material_t is a 'subclass' it should be fine to pass them back as cm_material_t; then our API won't suddenly be out of date because of r_material_t changing if we do changes in the renderer.

Ideally we'd do something similar for most of our renderer structs, keep them abstract as possible to the cgame.

src/client/renderer/r_types.h
```diff
@@ -398,7 +397,6 @@ typedef struct r_material_s {
 	r_image_t *tintmap;
 
 	uint32_t time;
-	uint32_t flags; // these may differ from cm->flags
```
Paril: I was wondering about this too, my assumption was that this existed so that we can test if a material has x stage flag/is diffusable/etc instead of having to sort through all of the stages to figure that out. I'll have to read into it further but I'm wondering how you got the STAGE_DIFFUSE/LIGHTING, etc working if this is gone now.

EDIT: ok, I see now that it's because cm->flags is used instead now. Good idea, this does seem a bit redundant.

src/collision/cm_material.h
```diff
@@ -73,8 +89,7 @@ typedef enum {
 
 typedef struct cm_stage_s {
 	uint32_t flags;
-	char image[MAX_QPATH];
-	int32_t image_index;
+	cm_asset_t asset;
```
Paril: I like the idea of a common asset class. Might make cgame easier to manage down the road.

src/cgame/default/cg_media.c
jdolan: Well, we expose all of `r_types.h` to the cgame already, and these pointers _live_ on an `r_model_t`. Seems natural that it would point to `r_material_t`, to me.

src/client/renderer/r_types.h
jdolan: Yea, it was these that were fucking me for like 2 days with the animations not working ;)

src/collision/cm_material.h
jdolan: Yep, and now everywhere that works with materials has a single abstraction for resolving images.

PR #469, opened 2017-05-06
OpenAL sound driver

Paril:
(New "suggestions" for Reviewers is cool! Told me to recommend jdolan because he recently edited the files around here)

Alright, so here's the first fully working/functional commit of OpenAL. I also did some general cleanup of music and sound subsystems.

Music is on its own thread, so it functions just like the Mix used to. Changing "threads" to -1 needs testing though; the behavior might be broken if -1 ends all running threads, because the thread runs for the length of the game.

The game will need to have sdl_sound compiled into it with WAV and OGG enabled, as well as OpenAL linked into it. If we intend to compile FLAC or MP3 support, we should do this now.

Because of the removal of SDL_mixer, the Windows version is now several MB smaller (less DLLs!).


src/client/sound/s_sample.c
```diff
@@ -34,41 +34,55 @@ static _Bool S_LoadSampleChunkFromPath(s_sample_t *sample, char *path, const siz
...
 		int64_t len;
 		if ((len = Fs_Load(path, &buf)) == -1) {
 			continue;
 		}
 
-		if (!(rw = SDL_RWFromMem(buf, (int32_t) len))) {
-			Fs_Free(buf);
-			continue;
-		}
+		Sound_AudioInfo desired = {
```
jdolan: Can this be const?

src/client/sound/s_sample.c
```diff
@@ -34,41 +34,55 @@ static _Bool S_LoadSampleChunkFromPath(s_sample_t *sample, char *path, const siz
...
+		};
 
-		if (!(sample->chunk = Mix_LoadWAV_RW(rw, false))) {
-			Com_Warn("%s\n", Mix_GetError());
+		Sound_Sample *snd = Sound_NewSampleFromMem(buf, (Uint32) len, ext, &desired, 2048);
+		
+		if (!snd) {
+			Com_Warn("%s\n", Sound_GetError());
+		} else {
+			Uint32 snd_length = Sound_DecodeAll(snd);
```
jdolan: const? Also is Uint32 the OpenAL return type here? We're using stdint.h, so uint32_t would be preferred.

src/client/sound/s_mix.c
```diff
@@ -120,37 +157,33 @@ void S_MixChannels(void) {
...
+					alSourcei(s_env.sources[i], AL_BUFFER, ch->sample->buffer);
+					alSourcei(s_env.sources[i], AL_LOOPING, (ch->play.flags & S_PLAY_LOOP));
+					alSourcePlay(s_env.sources[i]);
+				} else {
+					ALenum state;
+					alGetSourcei(s_env.sources[i], AL_SOURCE_STATE, &state);
+					
+					if (state != AL_PLAYING) {
+						S_FreeChannel(i);
+						continue;
```
jdolan: Love it!

src/client/sound/s_mix.c
```diff
@@ -105,10 +132,20 @@ static _Bool S_SpatializeChannel(s_channel_t *ch) {
...
-	Mix_Volume(-1, Clamp(s_volume->value, 0.0, 1.0) * MIX_MAX_VOLUME);
+	const vec3_t orientation[] = {
+		{ r_view.forward[0], r_view.forward[1], r_view.forward[2] },
+		{ r_view.up[0], r_view.up[1], r_view.up[2] }
+	};
+
+	alDistanceModel(AL_NONE);
+	alListenerfv(AL_POSITION, r_view.origin);
+	alListenerfv(AL_ORIENTATION, &orientation[0][0]);
+	//alDopplerFactor(0.05);
```
jdolan: No good? :)

src/client/sound/s_mix.c
```diff
@@ -90,12 +106,23 @@ static _Bool S_SpatializeChannel(s_channel_t *ch) {
...
+	ch->gain = 1.0 - frac;
 
-	const vec_t dot = DotProduct(r_view.right, delta);
-	const vec_t angle = acos(dot) * 180.0 / M_PI - 90.0;
+	// fade out frame sounds if they are no longer in the frame
+	if (ch->start_time) {
+		if (ch->play.flags & S_PLAY_FRAME) {
+			if (ch->frame != cl.frame.frame_num) {
+				const int32_t frame_diff = cl.frame.frame_num - ch->frame;
+				const uint32_t ms_diff = frame_diff * QUETOO_TICK_MILLIS;
```
jdolan: To match the rest of the code, I'd probably name these `frames` and `delta` respectively.

src/client/sound/s_mix.c
```diff
@@ -58,17 +60,31 @@ static _Bool S_SpatializeChannel(s_channel_t *ch) {
...
 			const r_model_t *mod = cl.model_precache[ent->current.model1];
 			VectorLerp(mod->mins, mod->maxs, 0.5, center);
 			VectorAdd(center, ent->origin, org);
 		} else {
-			VectorCopy(ent->origin, org);
+			if (ent == cl.entity) {
+				VectorCopy(r_view.origin, org);
+			} else {
+				VectorCopy(ent->origin, org);
+			}
```
jdolan: This is a really good catch. Keeps pain sounds centered on the player, etc.

Quetoo.vs15/libs/sdl_sound/SDL2/SDL_sound.h
```diff
@@ -0,0 +1,737 @@
...
+#  define SNDDECLSPEC __declspec(dllexport)
+#elif (__GNUC__ >= 3)
+#  define SNDDECLSPEC __attribute__((visibility("default")))
+#else
+#  define SNDDECLSPEC
+#endif
+
+#define SOUND_VER_MAJOR 1
+#define SOUND_VER_MINOR 0
+#define SOUND_VER_PATCH 1
```
jdolan: This is an older version of SDL_sound. I think you want 1.0.3. In version 1.0.3, `Sample_NewSampleFromMem` is not public. The "right way" to do this would be to create an `SDL_RWops` with `SDL_RWFromMem`, and then pass that to `Sample_NewSample`.

jdolan:
So, I pushed a few changes to get this going on OS X. It compiles and runs, but it hangs on loading any sound files. I'm sure it's something to do with SDL_sound, but I've no idea what yet.

I'm not jazzed about requiring SDL_sound, because that means we're now depending on SDL and SDL2 at the same time. In fact, I have to "trick" SDL_sound into seeing SDL2's `SDL.h`, which I'm not in love with. But.. I dunno what other cross platform audio loading framework we might use, either.

src/client/sound/s_mix.c
Paril: This is how it worked in Mixer, but OpenAL requires this adjustment otherwise the sound will fade away as it's being made.

src/client/sound/s_sample.c
Paril: I usually try to use the same return value as the function I'm calling, not one that we use, just to ensure that we're always getting the right data back in case our own versions change for whatever reason. Personal preference I guess.

src/client/sound/s_mix.c
Paril: Will do.

src/client/sound/s_mix.c
Paril: Just commented so you guys can play with it. It's a funny effect for sure.

Paril:
What about http://www.mega-nerd.com/libsndfile/ ? It's not SDL-dependent at all.

src/client/sound/s_main.c
```diff
@@ -77,7 +139,8 @@ void S_Frame(void) {
 	}
 
 	if (s_reverse->modified) {
-		Mix_SetReverseStereo(MIX_CHANNEL_POST, s_reverse->integer);
+		// TODO
+		//Mix_SetReverseStereo(MIX_CHANNEL_POST, s_reverse->integer);
```
jdolan: Honestly, this is a stupid option. I don't know anyone who needs it. We can remove this if it's going to take more than 5 seconds to fix.

src/client/sound/s_main.c
```diff
@@ -255,13 +312,22 @@ void S_Shutdown(void) {
...
+		alDeleteSources(MAX_CHANNELS, s_env.sources);
+		S_CheckALError();
+	}
 
 	S_ShutdownMusic();
 
 	S_ShutdownMedia();
+	
+	if (s_env.context) {
+		alcDestroyContext(s_env.context);
```
jdolan: I'm getting a 100% reproducible crash on this line. No idea.

screen shot 2017-05-07 at 9 09 19 am

src/client/sound/s_sample.c
```diff
@@ -28,47 +28,109 @@ static const char *SAMPLE_TYPES[] = { ".ogg", ".wav", NULL };
...
+	const vec_t fracstep = stepscale * 256.0;
+
+	for (size_t i = 0; i < outcount; i++) {
+		const int32_t srcsample = samplefrac >> 8;
+
+		samplefrac += fracstep;
+		(*outdata)[i] = LittleShort(indata[srcsample]);
+	}
+
+	return outcount;
```
jdolan: What'd I tell you? :)

src/client/sound/s_sample.c
```diff
@@ -28,47 +28,109 @@ static const char *SAMPLE_TYPES[] = { ".ogg", ".wav", NULL };
...
+	const size_t outcount = incount / 2;
+
+	*outdata = Mem_Malloc(outcount * sizeof(int16_t));
+
+	for (size_t i = 0, x = 0; x < outcount; i += 2, x++) {
+		(*outdata)[x] = (((int32_t) indata[i]) + ((int32_t) indata[i + 1])) / 2;
+	}
+
+	return outcount;
+}*/
```
jdolan: If you do put this back in, it should take the sndfile handle in so that it can average based on the number of channels in the source sample. That way, it'll be safe to call on mono samples too -- just in case.

src/client/sound/s_main.c
```diff
@@ -35,23 +34,86 @@ extern cl_client_t cl;
...
+	return rwops->read(rwops, ptr, 1, count);
+}
+
+/**
+ * @brief
+ */
+static sf_count_t S_RWops_write(const void *ptr, sf_count_t count, void *user_data) {
+	SDL_RWops *rwops = (SDL_RWops *) user_data;
+	return rwops->write(rwops, ptr, 1, count);
+}
```
jdolan: Clever to use SDL_RWops as a shortcut here.

src/client/sound/s_sample.c
Paril: Ideally it'd only be called on channels == 2, and I don't think we'd ever support > 2

Paril:
Should be ready to go now. Feel free to look over and then merge when ready.

src/client/sound/s_main.c
jdolan: Are we good with just killing this?

PR #476, opened 2017-05-16
Paril/model cleanup

Paril:
Simplify model structures so that the public-facing component is one type (no more split between obj and md3) - this makes a lot of the code paths much simpler
Speed up obj loading
Support groups in obj files
Moved material from model to mesh, updated rendering paths to account for this
Some slight cleanup/improvement of rendering pipeline (mesh and lights)
Support shaders in md3s and materials in obj files, which are (attempted to be) loaded as the material used by a particular mesh. This allows @Panjoo and friends to use multiple texture files for a single model, and will allow us to reduce the rendering complexity of some of our models drastically (ie, moving the small blinking lights to separate textures & meshes so that they don't have to literally re-render the *whole* gun just to render a tiny blinking light)

### DIFF
--- a/Quetoo.vs15/quetoo_all.sln
+++ b/Quetoo.vs15/quetoo_all.sln
@@ -426,4 +426,19 @@ Global
 		{6EBF61D4-CA15-4D8A-A957-7C9CF8A3E815} = {39B7FD8A-D56D-480C-BFF3-E0DCB1B151A7}
 		{FEF51ACC-31F2-41AC-B562-41F5C5988088} = {12075155-7B18-4525-885E-A2CCD24EFE04}
 	EndGlobalSection
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
+	GlobalSection(Performance) = preSolution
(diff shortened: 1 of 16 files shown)


jdolan:
This is great. The only part I don't love is R_FindMaterial. I don't think the caller should have to worry about whether or not we may have already loaded something. It's making the rest of the code work around a suboptimal R_LoadMaterial API imo.

Paril:
@jdolan The reason for the implementation is thusly: R_LoadMaterial either returns an existing material or creates a dummy one with a diffuse set to the input. In the case of supporting mesh-based materials, I need to check if materials explicitly set from the usemtl/md3 skin exists or if a material implicitly created from the mesh name exists already before calling back to just loading the old "skin" material (which already exists, so I don't need to call LoadMaterial again). The only way I can do that without creating 2 dummy materials along the way is by splitting the logic of LoadMaterial into Find or Load, where Find just returns one that exists and Load checks Find and returns dummy if not found.

I don't want this resolution function to create 2 dummy materials for every single model just because it doesn't exist. I have another idea for implementing this (searching through the model::materials array), but this version is technically the fastest because it utilizes our existing media subsystem and hash table.

If you got an idea I'm all ears, though.

jdolan:
I figured that was the rationale. We could always try loading each, and if they fail to resolve, simply free them? Either way, I don't want to hold this PR up. Merge away :)

Paril:
We have no method of freeing media, though. Media is freed at the discretion of the seed, and I think we'd be violating that concept of we're loading materials just to be unloaded because they don't exist. It seems a little overkill when we could just see if she's already loaded.

I think the better option, if you really want to not have FindMaterial exposed, is to just look through model::materials and strcmp the material names looking for a match. Models are unlikely to have more than three/four materials per file so it shouldn't be a huge hit compared to the hash table.

PR #483, opened 2017-05-22
Paril/geometry shaders

Paril:
Initial attempt at geometry shaders. Worked out quite nicely. GPU handles vertex calculations! This gives me a 10 FPS increase on Fractures, may do wonders for other people who are more CPU-bound than I am.

* Since we're actually doing core GL 3.3 since many commits ago, I updated our GLAD output to be core 3.3 + anisotropic extension. This gives us the ability to do some interesting new things:
  * ARB_vertex_type_2_10_10_10_rev is officially in this version, so I may re-look into getting that working (it's possible the extension was broken for @jdolan because it's no longer an extension, it's actually part of core 3.3)
  * ARB_explicit_attrib_location (which may help make shaders much easier to interact with; instead of having the locations be implicit and deducing them in the code, we can just have the attribute IDs be explicit locations that you assign in the shader which can be pulled in order, this way they can have any name).
  * ARB_geometry_shader4 (geometry shaders, wew)
  * ARB_draw_elements_base_vertex (offset to DrawElements that is added to indexes for vertex lookup; not sure how we can use this yet but I'll have to think about it)
  * NV_primitive_restart (the ability to restart primitives like LINE_LOOP, TRIANGLE_STRIP/LOOP without another call to DrawElements. Need to test if drawing loops + restart this way is quicker than doing 6 indices for something like particles or console characters)
  * ARB_texture_rectangle (rectangular textures, where you fetch texels from [0,w]-[0,h] instead of floating points - might be handy for atlasses or fonts),
  * ARB_texture_buffer_object (this one interests me quite a bit. We might be able to speed up lightmap and stainmap generation + usage if we use texture buffer objects for them. It might also let us just use the native lightmap format, since TBOs are 1 dimensional, as are lightmaps),
  * ARB_uniform_buffer_object (a bit of a complicated feature, but allows us to basically share one giant block of uniforms across all programs. We can use this for matrices and the like, which may help reduce the amount of uniform calls per frame).
  * Not a new extension, but we should also look into using ARB_pixel_buffer_object for stainmaps.
* The ability to create integral shader attributes has been added.
* Offset is now calculated in the buffer creation and is not specified by hand
* Program code split up into three sections: create, attach, link. A convenience function was added which does the old routine, but you can do them by hand if you need to (which I do for geometry-enabled shaders).
* Added geometry shader for particle + corona, as well as custom vertex shaders to handle the input/output of the new buffers. The fragment shaders remain the same, minus some adjustments in particle to remove unnecessary variables (tintmaps).
* All shaders were adjusted to use interface blocks for input/outputs. Makes them much easier to read. Ideally I'd like to have them be a bit more uniform in this usage but I haven't gotten around to it.
* Added new buffer code, linkage, uploading, etc for the new geometry shaders in r_particle. 
* Particle indexing upgraded to uint32. I wondered why sometimes with extreme particles you'd end up with shit covering the screen, it's because MAX_PARTICLES * 6 is way beyond USHRT_MAX. Derp.
* Four new attributes for the programs are now available. Honestly I'm not happy at all with the way programs/attributes are interacted with, I need to figure out something better.


src/client/renderer/r_array.c
```diff
@@ -350,10 +351,10 @@ void R_CreateInterleaveBuffer_(r_buffer_t *buffer, const GLubyte struct_size, co
 		Com_Warn("Buffer struct not aligned to 4, might be an error\n");
 	}
 
-	R_CreateBuffer(buffer, R_ATTRIB_TOTAL_TYPES, struct_size, false, hint, R_BUFFER_DATA | R_BUFFER_INTERLEAVE, size, data,
+	R_CreateBuffer(buffer, R_ATTRIB_TOTAL_TYPES, struct_size, false, false, hint, R_BUFFER_DATA | R_BUFFER_INTERLEAVE, size, data,
```
jdolan: When things get over 3-4 parameters, I think we should consider refactoring to take in a pointer to a struct, as we do for a lot of the cgame-visible entry points. That way, it's easy to keep track of the parameters, it's easy to pass `NULL` and `false` because you can just omit them, etc. So this could look something like:

    R_CreateBuffer(&(const r_create_buffer_t) {
        .id = id,
        .types = R_ATTRIB_TOTAL_TYPES,
        .size = size,
        .hint = hint,
        .flags = R_BUFFER_DATA | R_BUFFER_INTERLEAVE,
        .stride = struct_size,
        .data = data
    });

Thoughts?

src/client/renderer/r_material.c
```diff
@@ -49,36 +49,31 @@ static r_buffer_layout_t r_material_buffer_layout[] = {
...
-		.size = sizeof(vec2_t),
-		.offset = 44
+		.size = sizeof(vec2_t)
 	},
 	{
 		.attribute = R_ARRAY_LIGHTMAP,
 		.type = R_ATTRIB_UNSIGNED_SHORT,
 		.count = 2,
 		.size = sizeof(u16vec2_t),
-		.offset = 52,
```
jdolan: Nice. You're able to use `sizeof` or `offsetof` here now?

src/client/renderer/r_particle.c
```diff
@@ -50,8 +50,31 @@ typedef struct {
...
+typedef struct {
+	vec3_t start;
+	vec2_t texcoord0;
+	vec2_t texcoord1;
+	u8vec4_t color;
+	vec_t scale;
+	vec_t roll;
+	vec3_t end;
+	int32_t type;
+} r_geometry_particle_interleave_vertex_t;
```
jdolan: 👍

src/client/renderer/r_particle.c
```diff
@@ -180,7 +211,7 @@ static void R_ParticleTexcoords(const r_particle_t *p, r_particle_interleave_ver
 	// atlas needs a different pipeline
 	if (is_atlas) {
 		const r_atlas_image_t *atlas_image = (const r_atlas_image_t *) p->image;
-
+
```
jdolan: Whitespace :P

src/client/renderer/r_particle.c
```diff
@@ -247,16 +350,25 @@ void R_UpdateParticles(r_element_t *e, const size_t count) {
...
+			R_ParticleGeometryColor(p, &r_particle_state.geometry_verts[r_particle_state.num_particles]);
+		} else {
+			const uint32_t vertex_start = r_particle_state.num_particles * 4;
 
 			R_ParticleVerts(p, &r_particle_state.verts[vertex_start]);
 			R_ParticleTexcoords(p, &r_particle_state.verts[vertex_start]);
 			R_ParticleColor(p, &r_particle_state.verts[vertex_start]);
 
-			const uint16_t index_start = r_particle_state.num_particles * 6;
+			const uint32_t index_start = r_particle_state.num_particles * 6;
```
jdolan: Yea, I think back when we were using `GL_QUADS`, we used to be okay with `uint16_t` -- or at least we blew it out much less often ;)

src/client/renderer/r_program.c
```diff
@@ -481,15 +488,13 @@ static void R_LoadShader(GLenum type, const char *name, r_shader_t *out_shader)
...
+
 	if (!e) {
 		glGetShaderInfoLog(out_shader->id, sizeof(log) - 1, NULL, log);
 		Com_Warn("%s: %s\n", out_shader->name, log);
 
 		glDeleteShader(out_shader->id);
 		memset(out_shader, 0, sizeof(*out_shader));
-
-		Fs_Free(buf);
-		return;
```
jdolan: You should you don't need to return here on failure?

src/client/renderer/r_program_particle.c
```diff
@@ -0,0 +1,158 @@
...
+	R_ProgramParameter3fv(&p->weather_right, weather_right);
+	R_ProgramParameter3fv(&p->weather_up, weather_up);
+
+	for (int32_t i = 0; i < 2; i++) {
+		R_ProgramParameter3fv(&p->splash_right[i], splash_right[i]);
+		R_ProgramParameter3fv(&p->splash_up[i], splash_up[i]);
+	}
+
+	R_ProgramParameter1f(&p->ticks, r_view.ticks / 1000.0);
+}
```
jdolan: 👍

src/client/renderer/r_program_particle_corona.c
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @brief
+ */
+void R_UseParticleData_particle_corona(vec3_t weather_right, vec3_t weather_up, vec3_t splash_right[2], vec3_t splash_up[2]) {
+	r_particle_corona_program_t *p = &r_particle_corona_program;
+
+	R_ProgramParameter3fv(&p->view_right, r_view.right);
+	R_ProgramParameter3fv(&p->view_up, r_view.up);
+}
```
jdolan: Nice! Glad to see coronas dealt with in a shader too.

src/client/renderer/r_types.h
```diff
@@ -196,8 +210,9 @@ typedef union {
 		uint32_t type : 3;
 		uint32_t count : 3;
 		uint32_t offset : 6;
-		uint32_t normalized : 2;
 		uint32_t stride : 6;
+		uint32_t normalized : 2;
+		uint32_t integer : 2;
```
jdolan: Huh? What's this?

src/client/renderer/shaders/particle_corona_vs.glsl
```diff
@@ -0,0 +1,46 @@
+/**
```
jdolan: I would be in favor of adding some folder structure to our shaders. In fact, I have a (now horribly stale and in need of closing) pull request that moves our includes to "include". So, if we're going to have an assortment of particle shaders, we could put those in "shaders/particle" or something.

jdolan:
Ok, I posted this in Discord, but I wanted to get it down here as well for posterity.

I'm seeing what I think is some sort of GL state leakage in this branch. There's flickering on all geometry (world and mesh), and at times it almost seems to follow the crosshair (?) Could just be the order of PVS data..

Here's a video illustrating the issue:
https://www.dropbox.com/s/eskp4sbfyr5ay3q/flicker.mov?dl=0

I'm also seeing a performance drop on this branch, but hopefully that's just a side effect of the state leaking.

jdolan:
Ok, this problem goes away if I disable `r_shadows`.

Paril:
Hmm, shadows weren't touched.. I don't see this one either. Does it go away
with r_geo_shaders 0? I'll have to see where I changed stuff tonight that
might do this.

On May 22, 2017 9:54 AM, "Jay Dolan" <notifications@github.com> wrote:

Ok, this problem goes away if I disable r_shadows.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jdolan/quetoo/pull/483#issuecomment-303106680>, or mute
the thread

src/client/renderer/r_types.h
Paril: Added integer to the packed type struct, for integral attributes

src/client/renderer/r_array.c
Paril: Yeah that might be easier to handle in the long run

src/client/renderer/r_material.c
Paril: It can be calculated on the fly (itsvonly done once per buffer type), so I got rid of doing it by hand. I can also get rid of "size" since type + count can do it, but haven't done that yet.

src/client/renderer/r_program.c
Paril: Used to have to but now that the functions are split, no. The code after it will still work if a shader fails to load (the program will fail to compile which is what we want).

src/client/renderer/r_program_particle_corona.c
Paril: Coronas have been a shader for a long time, this just allows them to be packed in geometry shader too

src/client/renderer/shaders/shadow_vs.glsl
```diff
@@ -24,26 +24,27 @@ out VertexData {
...
+
+	ShadowVertex();
 	
 	// mvp transform into clip space
 	gl_Position = PROJECTION_MAT * point;
-
-	point = ShadowVertex();
 	    
-	fog = FogShadowVertex();
+    FogShadowVertex();
```
jdolan: Sure did!

PR #492, opened 2017-09-21
Crouchsliding

jdolan:


jdolan:
I generally have no problem with this, other than some naming conventions:

`cg_state_t` is a horrible name for what are really `bg_pmove` parameters. That struct should be renamed to something more appropriate.

Likewise, the getters for it are unnecessarily wordy. `Cg_GetHookPullSpeed` and `Cg_GetCrouchSlideBoost` etc.. I'm not even sure why these exist. There should just be an `extern cg_pm_params_t` or something that contains these things as members directly.

Similarly, I think that, rather than chew through config strings for each of these attributes, we should munge them all into a single config string that is user-info-delimited (`\\`):

 * gravity
 * hook pull speed
 * crouch slide enabled
 * crouch side boost (a value of `-1` could also serve to disable couch sliding?)
 * crouch slide friction
 * etc..

These variables will change so infrequently, that re-transmitting all of them in one string will be fine. That way, we're not burning config strings on things that almost never change. Using the user-info format will allow us to easily marshal and parse these values with existing code.

kaadmy:
I didn't add `cg_state_t`, Paril did that when he added the grapple hook.

I have no ideas why getters existed for the hook speed, CC @Paril

Joining the pmove values into a single userinfo string is a good idea.

I should also note that this I made this branch to experiment with crouch sliding, and I believe there were some bugs with it still, unfortunately I forget what exactly.

Joining the pmove values into a single userinfo string is a good idea.

I should also note that this I made this branch to experiment with crouch sliding, and I believe there were some bugs with it still, unfortunately I forget what exactly.

jdolan:
Yea, I noticed that the struct wasn't net-new. Just pointing out that it can / should be refactored as part of this change imo. 👍

jdolan:
@kaadmy Is this still something you're interested in implementing?

kaadmy:
Since this isn't a very good implementation as well as being incomplete and not work as intended, should I close this?

jdolan:
Up to you. I haven't looked at it very closely, to be honest. If you think it should be revisited with a fresh start some other time, then yes: please close and delete the branch.

PR #494, opened 2017-09-26
PhysicsFS 3.0.0

Paril:
- Minizip now uses miniz just to keep things all-in-one and not require zlib. Makes it easy for Winders. If this isn't what jdolan wants I'll modify it to require zlib like before and figure out what I'll do on the Windows side for that. Since it's the only dep we have that uses zlib anyways...
- Used the non-deprecated versions of the functions for PhysFS.
- Enumerate now allows the return values to stop, error or continue enumeration.


kaadmy:
Clean code, I figured there would be more changes to get Physfs3 working.

src/filesystem.h
```diff
@@ -41,7 +41,13 @@ typedef struct {
...
 } file_t;
 
-typedef void (*Fs_EnumerateFunc)(const char *path, void *data);
+typedef enum {
+	FS_ENUM_ERROR = -1,
+	FS_ENUM_STOP = 0,
+	FS_ENUM_CONTINUE = 1
+} Fs_EnumerateResult;
+
+typedef Fs_EnumerateResult (*Fs_EnumerateFunc)(const char *path, void *data);
```
jdolan: I guess this might be useful someday. Might as well expose it.

src/filesystem.h
```diff
@@ -65,7 +71,7 @@ int64_t Fs_Load(const char *filename, void **buffer);
 void Fs_Free(void *buffer);
 _Bool Fs_Rename(const char *source, const char *dest);
 _Bool Fs_Unlink(const char *filename);
-void Fs_Enumerate(const char *pattern, Fs_EnumerateFunc, void *data);
+int32_t Fs_Enumerate(const char *pattern, Fs_EnumerateFunc, void *data);
```
jdolan: I feel like the return type here could be either FsEnumerateResult or _Bool.

Paril:
Glib comprises like 400-some source files, and SDL comprises like 6 different modules that we use (and is in C++), so those are reasons why I'd say no on those, but PhysFS is fairly small and can easily be added as a submodule (well, it *could*, if it wasn't Mercurial...) or something? Just thinking out loud.
Doesn't have to be bundled, but would be cool if it did.

Paril:
Cou

Paril:
Whoopsies, wrong button.

src/filesystem.h
Paril: _Bool is preferable, yeah. I just picked int32_t because that's what PhysFS returns, but it only returns 0 or 1.

Paril:
Now is lib like before! Needs Unix and Linux build path though @jdolan

PR #496, opened 2017-09-28
PhysicsFS 3.0 support

Paril:
PhysFS 3.0.0 for real now. Not fake and not a virus.
Lib file, not in deps/ now.
Requires Linux/Mac toolsets


Quetoo.vs15/quetoo.vcxproj.user
```diff
@@ -12,9 +12,9 @@
     <LocalDebuggerCommandArguments>+set sv_max_clients 8 +set r_get_error 1 +set g_hook 1 +map edge</LocalDebuggerCommandArguments>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <LocalDebuggerCommandArguments>-p G:\quake2\baseq2 -p G:\quake2\ctf -p G:\quake2\lmctf -p G:\quake2\3tctf</LocalDebuggerCommandArguments>
+    <LocalDebuggerCommandArguments>-p G:\quake2\baseq2 -p G:\quake2\ctf -p G:\quake2\lmctf -p G:\quake2\3tctf +map fractures</LocalDebuggerCommandArguments>
```
jdolan: Dem maps tho!

Paril: Yeah was for testing ctf/3tctf/Q2 BSPs

src/filesystem.c
```diff
@@ -263,16 +263,13 @@ int64_t Fs_Load(const char *filename, void **buffer) {
 	} fs_block_t;
 
 	file_t *file;
-	if ((file = Fs_OpenRead(filename))) {
-
-		if (!PHYSFS_setBuffer((PHYSFS_File *) file, FS_FILE_BUFFER)) {
-			Com_Warn("%s: %s\n", filename, Fs_LastError());
-		}
```
jdolan: This is a great find, @Paril. I must have hastily read the docs, and assumed that you must always initialize a buffer. My loading times (which were already quite small) are now down to 2 seconds on the first map, and under 1 second on subsequent maps (with exceptions like Wake).

kaadmy:
FWIW, if the `share/quetoo/default` data dir is a link PhysFS 3.0 will raise an error and data will not be found.

jdolan:
Again, I'd like to merge the PhysicsFS 3.0 stuff, but not bother with the change in the enumeration API. We never use the break-iteration path, so imposing the return value everywhere is kind of annoying imo. Other than that, let's merge this!

jdolan:
I was going to address this (removing the break-iteration bits, but keeping the minor changes needed for PhysicsFS 3.0). But I can't correctly resolve the conflicts in the VS15 project files. Could you get around to merging this bad boy? Or, even if you just take care of the conflicts in your project files, I'm happy to remove the break-iteration PhysFS bits.

PR #499, opened 2017-09-30
Paril/windowscleanup

Paril:


jdolan:
Can we merge this too?

PR #503, opened 2017-10-26
Paril/lmcache

Paril:
This is more of a test concept, but it's cool to see in action. This adds a lightmap cache, which caches lightmap atlas outputs in your local game folder. It increases map load times by approximately 14x on subsequent loads (Wake down from 4300 ms to 340 ms, for instance). Fairly simple code for these returns, just not sure how you feel about the idea of a cache for this kind of thing. Let me know how you feel about it.

### DIFF
--- a/src/client/renderer/r_atlas.h
+++ b/src/client/renderer/r_atlas.h
@@ -41,15 +41,15 @@ typedef struct {
 } r_packer_node_t;
 
 typedef struct {
-	uint32_t max_width;
-	uint32_t max_height;
-
+	// required if only reading
+	uint32_t num_nodes;
 	r_packer_node_t *nodes;
-	uint32_t num_nodes, num_alloc_nodes;
-
 	uint32_t root;
 
+	// required for writing
+	uint32_t num_alloc_nodes;
 	_Bool keep_square;
+	uint32_t max_width, max_height;
 } r_packer_t;
 
 void R_AtlasPacker_InitPacker(r_packer_t *packer, const uint32_t max_width, const uint32_t max_height,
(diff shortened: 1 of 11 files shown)


src/client/renderer/r_atlas.h
```diff
@@ -31,35 +31,37 @@ void R_CompileAtlas(r_atlas_t *atlas);
...
-	uint32_t x;
-	uint32_t y;
+	uint16_t width;
+	uint16_t height;
+	uint16_t x;
+	uint16_t y;
 
 	uint32_t right;
 	uint32_t down; // nodes to the right/below this one
 } r_packer_node_t;
```
jdolan: Wouldn't mind seeing these structs named `r_atlas_packer_t`, `r_atlas_packet_node_t`, etc..

Paril:
Negligible amount of time. :)

src/client/renderer/r_atlas.h
Paril: I was also wondering if maybe we should just move the atlas packer to libimage or something, and call them Img_Atlas*. We only ever use the packer for images.

jdolan: I guess the question is: would we ever build an atlas outside of the renderer? Would Quemap potentially bake these lightmap atlases for us? Probably not, because they depend on your lightmap block size cvar, right? So, assuming this is a renderer-only concern, it should stay in the renderer imo.

Paril: Cgame uses it to compile an atlas, but that's about it.

Paril:
@jdolan made some minor changes, confirm that these are OK and we'll merge

jdolan:
@Paril The code looks great. You know me: I'm picky about names:

 * r_lmcache -> r_lightmap_cache
 * r_packer_t -> r_atlas_packer_t

Other than that, I love it.

PR #506, opened 2017-11-01
Lightmaps workbench

void-995:


src/client/renderer/r_bsp_model.c
```diff
@@ -620,7 +620,8 @@ static gboolean R_UniqueVerts_EqualFunc(gconstpointer a, gconstpointer b) {
 	        memcmp(bsp->normals[va], bsp->normals[vb], sizeof(vec3_t)) == 0 &&
 	        memcmp(bsp->texcoords[va], bsp->texcoords[vb], sizeof(vec2_t)) == 0 &&
 	        memcmp(bsp->lightmap_texcoords[va], bsp->lightmap_texcoords[vb], sizeof(vec2_t)) == 0 &&
-	        memcmp(bsp->tangents[va], bsp->tangents[vb], sizeof(vec4_t)) == 0;
+	        memcmp(bsp->tangents[va], bsp->tangents[vb], sizeof(vec3_t)) == 0 &&
+			memcmp(bsp->bitangents[va], bsp->bitangents[vb], sizeof(vec3_t)) == 0;
```
jdolan: I'm honestly surprised we don't need an epsilon / delta check here, but if it works..

src/client/renderer/r_types.h
```diff
@@ -155,24 +157,26 @@ typedef enum {
...
-	R_ATTRIB_MASK_TYPE			= (1 << R_ATTRIB_TYPE),
+	R_ATTRIB_MASK_SCALE				= (1 << R_ATTRIB_SCALE),
+	R_ATTRIB_MASK_ROLL				= (1 << R_ATTRIB_ROLL),
+	R_ATTRIB_MASK_END				= (1 << R_ATTRIB_END),
+	R_ATTRIB_MASK_TYPE				= (1 << R_ATTRIB_TYPE),
 
-	R_ATTRIB_MASK_ALL			= (1 << R_ATTRIB_ALL) - 1,
-	R_ATTRIB_GEOMETRY_MASK		= R_ATTRIB_MASK_SCALE | R_ATTRIB_MASK_ROLL | R_ATTRIB_MASK_END | R_ATTRIB_MASK_TYPE
+	R_ATTRIB_MASK_ALL				= (1 << R_ATTRIB_ALL) - 1,
+	R_ATTRIB_GEOMETRY_MASK			= R_ATTRIB_MASK_SCALE | R_ATTRIB_MASK_ROLL | R_ATTRIB_MASK_END | R_ATTRIB_MASK_TYPE
```
jdolan: Watch the whitespace diffs here.

src/shared.c
```diff
@@ -200,32 +200,15 @@ void PerpendicularVector(const vec3_t in, vec3_t out) {
...
-
-	if (DotProduct(t, bitangent) < 0.0) {
-		tangent[3] = -1.0;
-	} else {
-		tangent[3] = 1.0;
-	}
-
-	VectorScale(bitangent, tangent[3], bitangent);
+	VectorCopy(tdir, bitangent);
+	VectorNormalize(bitangent);
```
jdolan: I think we have a VectorNormalize2 that could replace the VectorNormalize && VectorCopy. I also wonder if it's time to just kill this function, if that's all it does.

src/tools/quemap/lightmap.c
```diff
@@ -753,12 +753,13 @@ static void SampleNormal(const light_info_t *l, const vec3_t pos, vec3_t normal)
...
 	{ 0.0, 0.0 },
-	{ -0.125, -0.125 },
-	{ 0.125, -0.125 },
-	{ 0.125, 0.125 },
-	{ -0.125, 0.125 }
+	{ -0.5, -0.5 },
+	{ 0.5, -0.5 },
+	{ 0.5, 0.5 },
+	{ -0.5, 0.5 }
 };
```
jdolan: lulz. Really? That's it? So the initial offsets were not wide enough to actually grab different sample points. We were effectively running "the same" sample 5 times?

void-995: Exactly, because you need to take _whole_ luxel area to do it correctly.

src/client/renderer/r_bsp_model.c
void-995: Would be great to have that, but as for just adding bitangent attribute support it's fine. Also, white-space again.

src/client/renderer/r_types.h
void-995: For a reasons it looked fine in Visual Studio. Mine tab size is 4. Will check it in another editor as well.

jdolan: We use hard tabs, maybe that's the issue.

void-995:
As for indirect I would better to concentrate on my tool. I guess, I fixed everything that was the real showstopper as for now and it looks way better. 

Yes, you would need to re-bake maps. As I recommend: _-extra -surface 2 -patch 4_

P.S. Will try to investigate whitespace and then merge.

PR #522, opened 2018-02-09
Texture Arrays, HDR

void-995:


src/client/renderer/r_lightmap.c
```diff
@@ -113,35 +114,40 @@ static void R_AllocStainmap_(const uint32_t width, const uint32_t height, r_stai
 	R_BindFramebuffer(FRAMEBUFFER_DEFAULT);
 }
 
-#define R_AllocLightmap(w, h) R_AllocLightmap_(IT_LIGHTMAP, w, h)
+#define R_AllocLightmap(w, h) R_AllocLightmap_(IT_LIGHTMAP, w, h, MAX_LIGHTMAP_PAGES)
```
jdolan: Is it pages, or layers? Or are those different things? Just for sake of consistency..

src/client/renderer/r_main.c
```diff
@@ -474,7 +475,7 @@ static void R_InitLocal(void) {
 	r_no_vis = Cvar_Add("r_no_vis", "0", CVAR_DEVELOPER, "Disables PVS refresh and lookup for world surfaces (developer tool)");
 	r_draw_bsp_leafs = Cvar_Add("r_draw_bsp_leafs", "0", CVAR_DEVELOPER, "Controls the rendering of BSP leafs (developer tool)");
 	r_draw_bsp_lights = Cvar_Add("r_draw_bsp_lights", "0", CVAR_DEVELOPER, "Controls the rendering of static BSP light sources (developer tool)");
-	r_draw_bsp_lightmaps = Cvar_Add("r_draw_bsp_lightmaps", "0", CVAR_DEVELOPER, "Controls the rendering of BSP lightmap textures (developer tool)");
+	r_draw_bsp_lightmaps = Cvar_Add("r_draw_bsp_lightmaps", "0", CVAR_DEVELOPER | CVAR_R_CONTEXT, "Controls the rendering of BSP lightmap textures (developer tool)");
```
jdolan: This requires an `r_restart`?

src/client/renderer/r_media.c
```diff
@@ -72,18 +74,27 @@ void R_DumpImage(const r_image_t *image, const char *output) {
 	glBindBuffer(GL_PIXEL_PACK_BUFFER, 0);
 	glBindBuffer(GL_PIXEL_UNPACK_BUFFER, 0);
 
-	R_BindDiffuseTexture(image->texnum);
+	// try GL_TEXTURE_2D first
+	target = GL_TEXTURE_2D;
+	R_BindUnitTexture(texunit_diffuse, image->texnum, target);
+
+	// and if GL_TEXTURE_2D failed, then it had to be GL_TEXTURE_2D_ARRAY
+	if (glGetError() == GL_INVALID_OPERATION) {
```
jdolan: Gross. There's no way to infer the correct target from the image itself? Seems like a miss.

src/client/renderer/r_state.c
```diff
@@ -868,9 +868,10 @@ void R_InitSupersample(void) {
 
 	r_state.supersample_image->width = r_context.render_width;
 	r_state.supersample_image->height = r_context.render_height;
+	r_state.supersample_image->layers = 0;
 	r_state.supersample_image->type = IT_NULL;
 
-	R_UploadImage(r_state.supersample_image, GL_RGBA, NULL);
+	R_UploadImage(r_state.supersample_image, r_hdr->integer ? GL_RGBA16 : GL_RGBA8, NULL);
```
jdolan: Should this be `r_hdr_enabled` .. err `r_using_hdr` ?

src/client/renderer/r_state.h
```diff
@@ -195,16 +195,16 @@ extern r_state_t r_state;
...
+void R_BindUnitTexture(r_texunit_t *texunit, GLuint texnum, GLenum target);
+
+#define R_BindDiffuseTexture(texnum)		R_BindUnitTexture(texunit_diffuse, texnum, GL_TEXTURE_2D)
+#define R_BindLightmapTexture(texnum)		R_BindUnitTexture(texunit_lightmap, texnum, GL_TEXTURE_2D_ARRAY)
+#define R_BindDeluxemapTexture(texnum)		R_BindUnitTexture(texunit_deluxemap, texnum, GL_TEXTURE_2D)
+#define R_BindNormalmapTexture(texnum)		R_BindUnitTexture(texunit_normalmap, texnum, GL_TEXTURE_2D)
+#define R_BindSpecularmapTexture(texnum)	R_BindUnitTexture(texunit_specularmap, texnum, GL_TEXTURE_2D)
+#define R_BindWarpTexture(texnum)			R_BindUnitTexture(texunit_warp, texnum, GL_TEXTURE_2D)
+#define R_BindTintTexture(texnum)			R_BindUnitTexture(texunit_tint, texnum, GL_TEXTURE_2D)
+#define R_BindStainmapTexture(texnum)		R_BindUnitTexture(texunit_stainmap, texnum, GL_TEXTURE_2D)
```
jdolan: Should we go ahead and merge normalmaps and specularmaps into a texture array along with diffuse? Wouldn't that save an awful lot of binds?

src/client/renderer/shaders/default_fs.glsl
```diff
@@ -123,8 +124,6 @@ void LightFragment(in vec4 diffuse, in vec3 lightmap, in vec3 normalmap, in floa
 			}
 		}
 	}
-
-	light *= MODULATE_SCALE;
```
jdolan: I think it's actually cleaner to keep this here, rather than munged in the line below.

src/client/renderer/shaders/default_fs.glsl
```diff
@@ -185,7 +184,12 @@ void main(void) {
 	vec3 deluxemap = vec3(0.0, 0.0, 1.0);
 
 	if (LIGHTMAP) {
-		lightmap = texture(SAMPLER1, texcoords[1]).rgb;
+#if DEBUG_LIGHTMAP_LAYER_INDEX > 0
+		vec4 lightmapColorHDR = texture(SAMPLER1, vec3(texcoords[1], DEBUG_LIGHTMAP_LAYER_INDEX - 1));
+#else
+		vec4 lightmapColorHDR = texture(SAMPLER1, vec3(texcoords[1], 0));
+#endif // DEBUG_LIGHTMAP_LAYER_INDEX
+		lightmap = lightmapColorHDR.rgb * lightmapColorHDR.a;
```
jdolan: Oh, ok. This is why `r_draw_bsp_lightmaps` requires `r_restart`.

src/shared.c
```diff
@@ -654,6 +666,50 @@ void ColorDecompose3(const vec3_t in, u8vec3_t out) {
 	}
 }
 
+/**
+* @brief H-basis manipulating.
+*/
+void ProjectOntoH4(const vec3_t direction, h4_t result) {
```
jdolan: Jumping the gun here? :)

src/tools/quemap/lightmap.c
```diff
@@ -1015,50 +1042,42 @@ void FinalizeLighting(int32_t face_num) {
...
-
-				if (tex->flags & SURF_PHONG) { // interpolated normal
-					SampleNormal(f, fl->origins + i * 3, normal);
-				} else {
-					if (f->side) {
-						VectorNegate(bsp_file.planes[f->plane_num].normal, normal);
-					} else {
-						VectorCopy(bsp_file.planes[f->plane_num].normal, normal);
-					}
-				}
```
jdolan: Ok, this is what I was asking about over Discord. So you basically tackle this when the sample is gathered, rather than when it is finalized. Works for me 👍

src/client/renderer/shaders/default_fs.glsl
void-995: It's here because tonemap breaks it. So it doesn't break an equation if I move it to the end of file.

src/client/renderer/r_state.c
void-995: Can be, it's not that much required here, but it's logical to change this.

src/client/renderer/r_state.h
void-995: They can be different size. What if diffuse 2 times bigger than normal map?

src/client/renderer/r_media.c
void-995: It should be possible by using it from image type, but there are some special virtual images that has most of it's fields filled with some kind of garbage.

src/client/renderer/r_lightmap.c
void-995: PAGES = LAYERS, may rename it later

src/client/renderer/shaders/default_fs.glsl
jdolan: 👍

src/client/renderer/r_state.h
jdolan: I guess that happens sometimes. I'd say it's rare, but we currently support it.

src/client/renderer/r_media.c
jdolan: Ruh oh. Can you call any of them out? @Paril and I can probably get those fixed.

src/client/renderer/r_lightmap.c
jdolan: Let's use LAYERS everywhere for consistency.

src/client/renderer/r_image.c
```diff
@@ -75,6 +75,8 @@ static void R_TextureMode(void) {
 typedef struct {
 	uint32_t width;
 	uint32_t height;
+	uint32_t using_rgb;
+	uint32_t using_hdr;
```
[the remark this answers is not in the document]
void-995: Fair enough

src/client/renderer/r_media.c
void-995: I may do warning and skip with them

src/shared.c
void-995: A bit

src/client/renderer/r_image.c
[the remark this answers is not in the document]
SpineyPete: I don't like it when cvars use the imperative form because I always forget which verb (show/draw/use/enable/...) is used, and if it's in front or behind, etc. Also find it easier for autocomplete lists to sort by I-want-all-settings-for-this-component rather than I-want-to-enable-something-and-maybe-you-have-it. *shrug*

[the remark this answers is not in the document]
jdolan: That's a very fair point and we could probably act on it with a bunch of our cvars. I agree. But this one in particular is not one that the user should mess with. It's CVAR_NO_SET, and he's really only using a cvar as a means to inject branching logic into the shader. That said, I won't push on this rename any further.

src/client/renderer/r_media.c
Paril: None of them should be garbage, but, I think I recall what they're talking about. Atlasses are "virtual" images/they extend r_image_t, so that's probably what you're seeing there. Agree that target should be stored in r_image_t though.

void-995: I'm looking for target by `layers` property which is missing for atlases. If that will be fixed then I might change this part to work correctly.

Paril: Both r_atlas_t and r_atlas_image_t extend r_image_t, but I don't think they memset to zero. Either way you'll have to assign the target/layers in R_CompileAtlas and R_StitchAtlas, respectively. You can throw them under the "image.type = IT_ATLAS_" lines.

src/tools/quemap/lightmap.c
```diff
@@ -230,6 +230,21 @@ static void CalcSamplePoints(const light_info_t *l, vec_t sofs, vec_t tofs, vec_
...
+	// temporary storage in case if direction and result are the same pointer
+	vec3_t temporary_result;
+
+	// transform it into tangent space
+	temporary_result[0] = DotProduct(direction, tangent);
+	temporary_result[1] = DotProduct(direction, bitangent);
+	temporary_result[2] = DotProduct(direction, normal);
+
+	VectorCopy(temporary_result, result);
+}
```
jdolan: This makes sense. I dig it. To bounce indirect light, you have to transform the other way, tho, right?

src/tools/quemap/lightmap.c
```diff
@@ -734,14 +753,18 @@ static void SampleNormal(const bsp_face_t *face, const vec3_t pos, vec3_t normal
...
 static const vec_t sampleofs[MAX_SAMPLES][2] = {
-	{ 0.0, 0.0 },
-	{ -0.5, -0.5 },
-	{ 0.5, -0.5 },
-	{ 0.5, 0.5 },
-	{ -0.5, 0.5 }
+	{ +0.0, +0.0 }, { -0.5, -0.5 }, { +0.0, -0.5 },
+	{ +0.5, -0.5 },	{ -0.5, +0.0 }, { +0.5, +0.0 },
+	{ -0.5, +0.5 }, { +0.0, +0.5 }, { +0.5, +0.5 }
+};
```
jdolan: Oh damn. So -extra is going to do a 9 samples per luxel now?

src/tools/quemap/lightmap.c
```diff
@@ -810,6 +833,10 @@ void DirectLighting(int32_t face_num) {
 	fl->directions = Mem_TagMalloc(fl->num_samples * sizeof(vec3_t), MEM_TAG_FACE_LIGHTING);
 	fl->indirect = Mem_TagMalloc(fl->num_samples * sizeof(vec3_t), MEM_TAG_FACE_LIGHTING);
 
+	memset(fl->direct, 0, fl->num_samples * sizeof(vec3_t));
+	memset(fl->directions, 0, fl->num_samples * sizeof(vec3_t));
+	memset(fl->indirect, 0, fl->num_samples * sizeof(vec3_t));
+
```
jdolan: Mem_TagMalloc guarantees 0'd memory.

src/tools/quemap/lightmap.c
```diff
@@ -903,6 +935,87 @@ void IndirectLightingImpact(const cm_trace_t *trace, const vec3_t color) {
...
+	num_lights = 0;
+	memset(lights, 0, sizeof(lights));
+
+	for (face_num = 0; face_num < bsp_file.num_faces; face_num++) {
+		const bsp_face_t *face = &bsp_file.faces[face_num];
+		const bsp_texinfo_t *texinfo = &bsp_file.texinfo[face->texinfo];
+		const bsp_plane_t *plane = &bsp_file.planes[face->plane_num];
+
+		if (texinfo->flags & (SURF_SKY | SURF_WARP | SURF_LIGHT | SURF_HINT | SURF_NO_DRAW)) {
+			continue; // we have no light to reflect
```
jdolan: Not sure about `SURF_WARP` here. Liquids certainly reflect some light, right?

src/tools/quemap/lightmap.c
```diff
@@ -903,6 +935,87 @@ void IndirectLightingImpact(const cm_trace_t *trace, const vec3_t color) {
...
+			light->next = lights[cluster];
+			lights[cluster] = light;
+
+			light->type = LIGHT_SPOT;
+			light->stopdot = 1.0;
+
+			VectorCopy(normal, light->normal);
+
+			VectorScale(color, 1.0 / 255.0 / 256.0 / (float)current_bounce, light->color);
+			light->intensity = 256;
```
jdolan: So they all uniformly bounce light 256 units away, but the light is scaled by the direct light color?

Shouldn't we factor the material hardness in here? We have easy access to it..

src/tools/quemap/lightmap.c
```diff
@@ -914,11 +1027,10 @@ void IndirectLighting(int32_t face_num) {
 	const bsp_face_t *face = &bsp_file.faces[face_num];
 	const bsp_texinfo_t *texinfo = &bsp_file.texinfo[face->texinfo];
 
-	if (texinfo->flags & (SURF_SKY | SURF_WARP)) {
+	if (texinfo->flags & (SURF_SKY | SURF_WARP | SURF_HINT | SURF_NO_DRAW)) {
```
jdolan: Again, SURF_WARP can maybe go here (I know you didn't add it)

src/tools/quemap/lightmap.c
```diff
@@ -928,37 +1040,37 @@ void IndirectLighting(int32_t face_num) {
...
+		byte pvs[(MAX_BSP_LEAFS + 7) / 8];
 
-		vec3_t reflect;
-		Reflect(direction, normal, reflect);
+		const vec_t *point = fl->origins + i * 3;
 
-		vec3_t end;
-		VectorMA(org, light, reflect, end);
+		if (texinfo->flags & SURF_PHONG) { // interpolated normal
+			SampleNormal(face, point, normal);
```
jdolan: I'm surprised you haven't torn SampleNormal a new one by now. It's clearly not correct.

src/tools/quemap/lightmap.c
void-995: That's nice to know, will clean that up

src/tools/quemap/lightmap.c
jdolan: I'm not suggesting we get all fancy with it. Just multiply the reflected light by the material hardness 🤓 That's literally what it does in the shader.

SpineyPete: Hardness only scales the specular light, not the diffuse :)

jdolan: Touché.

jdolan:
I noticed that `R_FilterLightmapTexture` is disabled, with a comment that we should move brightness, modulate, saturation and contrast into GLSL. Agreed! Is that something you want to tackle as part of this PR?

jdolan:
I re-baked Torn Glory locally:

    quemap -bsp -vis -light -indirect maps/torn.map

The result is very dark:

quetoo012

I tried again without indirect:

    quemap -light maps/torn.bsp

Still dark:

quetoo013

Interestingly, the screenshots are *much paler* than the image on-screen. On-screen, this image was mostly black.

I also noticed a warning about `Unable to resolve SAMPLER2 in default_glsl`, which appears valid. The variable is declared but never used, and thus optimized out. Seems these are holdovers from pre-texture arrays?

src/client/renderer/r_main.c
```diff
@@ -335,7 +337,12 @@ void R_EndFrame(void) {
 		}
 	}
 
-	SDL_GL_SwapWindow(r_context.window);
+	if (r_state.screenshot_pending) {
+		R_Screenshot_f();
+		r_state.screenshot_pending = false;
+	} else {
+		SDL_GL_SwapWindow(r_context.window);
+	}
```
jdolan: @void-995 Why is this necessary now?

PR #528, opened 2018-02-21
Quake3 t-junctions fixes

jdolan:


src/tools/quemap/faces.c
```diff
@@ -441,10 +442,15 @@ static void FixEdges_r(node_t *node) {
...
 
+void FixTJunctionsQ3( node_t *node );
+
 /**
  * @brief
  */
 void FixTjuncs(node_t *head_node) {
+	FixTJunctionsQ3(head_node);
+	FixTJunctionsQ3(head_node);
+
```
jdolan: It looks like we're always calling into the Q3 code, regardless of the `notjunc` parameter?

src/tools/quemap/tjunctions.c
```diff
@@ -0,0 +1,547 @@
...
+
+Quake III Arena source code is distributed in the hope that it will be
+useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Foobar; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+===========================================================================
```
jdolan: If this code works better, I'm happy to adopt it. We should run this through `astyle` or something so that it [mostly] aligns with the rest of Quemap.

src/tools/quemap/faces.c
jdolan: Er, and calling this twice?

PR #536, opened 2018-03-04
Tweak the various crosshair health modes

Calinou:
Green is now used to indicate overheal instead of blue since it is easier to see for the human eye.

The red-yellow-white formula has been adjusted to look more natural.


kaadmy:
Haven't tested ingame, but I'm fine with this.

PR #561, opened 2019-03-26
Some throwaway experiment in making the blood look more realistic. Pr…

jdolan:
…obably broke some shader spaghetti in the process. Some screenshots: https://imgur.com/a/OKVQBb2


jdolan:
These screenshots look dope AF. Can we get this into brave-new-world?

SpineyPete:
This is a total hack, the stainmap adds only blood now. I just hijacked the red channel of the stainmap. Blood would need it's seperate texture to not interfere with the other stainmap effects. Maybe something @Paril could look at?

PR #568, opened 2020-04-06
Materials changes in order to get the renderer right this time

jdolan:
* Glossmap and specularmap are now both properly supported.
 * Normalmap and heightmap are now both properly supported.
 * Specularmap, if present, is uploaded into the glossmap alpha channel.
 * Heightmap, if present, is uploaded into the normalmap alpha channel.
 * All world and mesh materials are array textures with 3 layers: diffusemap, normalmap, glossmap
 * Texture unit management is much simpler: single bind to change materials
 * Image types greatly reduced, as many are now IT_MATERIAL.
 * Refactor image loading API to return an SDL_Surface, finally.
 * Kill off the last dingleberries of the color palette.


SpineyPete:
Looks good to me.

PR #576, opened 2020-11-29
Volumetric fog

jdolan:
Opening up a pull request for the volumetric fog changes into brave-new-world. Big items here are:

1. BSP format changes: brushes now reference their original entity, and the light grid now includes an additional four channel 3D texture for fog color and density.
1. `misc_fog` entities will now create fog volumes. Their brushes should be `CONTENTS_MIST` and textured with _common/fog_.
1. `misc_dust` entities for sprite emission also received some scaffolding, but are not complete.
1. Fog is calculated during the `-light` stage, and luxels are either in or out of a fog volume. No attenuation.
1. Global fog is still defined in `worldspawn`, but with new keys, and is baked into the fog grid.
1. The light grid is now part of global shader state, since everything 3D requires fog.
1. I've introduced a UBO for all shared uniforms that includes light sources, light grid, color correction, etc.
1. Some shader consolidation into `common_vs.glsl` and `common_fs.glsl`.

There's still more cleanup and improvement to be done here. The shaders can probably still be made more consistent. @Paril one thing I would love eyes on is Cm_PointInsideBrush, which mostly works, but you can see some fog leakage on Frag Pipe. At any rate, I think this is good enough (and fucking awesome!) to be merged into BNW for further refinement and collaboration.


SpineyPete:
Fog leakage is to be expected, this is an issue common to all sampling of 3d textures. The only way to "fix" it is to increase texture dimension or just edit the brushes so it isn't as apparent. To be honest I was surprised we haven't had obvious cases of light leaking with the lightgrid, but I guess those are just not as noticeable.

jdolan:
> Fog leakage is to be expected, this is an issue common to all sampling of 3d textures. The only way to "fix" it is to increase texture dimension or just edit the brushes so it isn't as apparent. To be honest I was surprised we haven't had obvious cases of light leaking with the lightgrid, but I guess those are just not as noticeable.


It’s not leakage from sampling. If you turn on r_draw_bsp_lightgrid 2 you’ll see some samples clearly fall outside of the red fog in the main atrium of Frag Pipe, like inside one of the lower rooms. Half of one of the lower rooms has red fog. The others look correct. So something about Cm_PointInsideBrush is wrong. Maybe a front vs back plane reference on the brush side or something?

PR #577, opened 2020-12-28
Remove PVS / PHS / VIS entirely. Batch the world by material.

jdolan:
* Removes the `-vis` stage from quemap entirely. Farewell to the buggy code that no human ever fully understood. Farewell to long `-vis` compile times. Farewell to hall-of-mirror errors. Farewell to caring about structural vs detail, hint brushes, or really even caring about triangle counts.
 * The entire world is now batched by material into glDrawElements commands at compile time. Our maps typically use 20-50 materials, so the entire world is now drawn in ~35 draw commands.
 * A depth pre-pass technique is used to quickly render the world without materials. Subsequent rendering passes will take advantage of this through an early-Z test and skip expensive shader executions for occluded samples.
 * To facilitate the depth pre-pass, the client frame is now segmented into UpdateView and PopulateView. The former resolves the view origin and angles so that a minimal scene can be rendered. The latter then populates the scene, and allows the renderer to take advantage of hardware occlusion queries and reject sprites, entities, etc, which are occluded.
 * For further overdraw protection, the game now supports mapper-defined hardware occlusion queries. Place a brush of `CONTENTS_OCCLUSION_QUERY` around entire rooms or areas of the map. For each such brush, an occlusion query will execute once per frame to determine if any alpha blended faces, mesh entities, sprites, etc.. within the occlusion query brush volume can be skipped. This is experimental, and disabled by default for now. `r_occlude 1` to enable.
 * Alpha blended faces are now batched by plane into glDrawElements commands at compile time. These are sorted by BSP recursion each frame. This means that, for example, all water lower on Tokays Towers is drawn in a single command. This dramatically increases performance of blend face rendering, but also blend depth sorting for sprites and mesh entities.
 * Introduced `SURF_DECAL` to flag faces that provide decals. Since sprites and meshes can not appear behind decals (unless the decals are themselves on glass), this allows the renderer to optimize the blend depth sorting algorithm. This is optional, but may provide a performance boost on maps with lots of decals.
 * A good bit of other miscellaneous cleanup.


jdolan:
I tagged you guys in case you wanted to weigh in on this. I'm not asking for an in-depth code review (but would welcome one :)). Are we all good with killing off PVS?

PR #592, opened 2021-01-06
#593 Paril's blob shadows as sprites

jdolan:


src/client/renderer/r_flare.c
```diff
@@ -64,6 +64,7 @@ void R_LoadFlare(r_bsp_model_t *bsp, r_bsp_face_t *face) {
 	}
 
 	face->flare->media = s->media;
+	face->flare->softness = 1.f;
```
jdolan, line 67: Great example of somewhere we might turn this up.

jdolan: Or even allow the material to define it... eh? Eh??

jdolan (commented):
This looks great @Paril!

PR #611, opened 2021-02-02
Paril/quickmaffs

Paril:


src/matrix.h
```diff
@@ -20,96 +21,465 @@
...
 
-typedef struct matrix4x4_s {
+/**
+ * @brief Sixteen-component single precision 4x4 matrix type.
+ */
+typedef union {
+	/**
+	 * @brief Flat array accessor.
+	 */
+	float flat[16];
```
jdolan, line 34: I'd probably call this 'array'

src/matrix.h
```diff
@@ -20,96 +21,465 @@
...
-// generate a perspective transformation matrix from bounds
-void Matrix4x4_FromFrustum (mat4_t *out, double left, double right, double bottom, double top, double nearval,
-                            double farval);
-// generate an orthogonal projection matrix from bounds
-void Matrix4x4_FromOrtho (mat4_t *out, double left, double right, double bottom, double top, double nearval,
-                          double farval);
+/**
+ * @return A `mat4_t` with the specified components.
+*/
+static inline mat4_t __attribute__ ((warn_unused_result)) Mat4(const float elements[16]) {
```
jdolan, line 45: The rest of the code puts __attribute__ after the function parameters (at the end of the line.). Let's pick one place for them and make it consistent.

src/matrix.h
```diff
@@ -20,96 +21,465 @@
...
+	mat4_t matrix;
+	memcpy(matrix.flat, elements, sizeof(matrix.flat));
+	return matrix;
+}
+
+/**
+ * @return A `mat4_t` with the specified rows.
+*/
+static inline mat4_t __attribute__ ((warn_unused_result)) Mat4_FromRows(const float row0[4], const float row1[4], const float row2[4], const float row3[4]) {
```
jdolan, line 54: Should these take vec4_t for rows and cols?

src/matrix.h
```diff
@@ -20,96 +21,465 @@
...
+				0.f,
+				1.f
+			});
+	}
+}
+
+/**
+ * @return The inverse of the input matrix.
+ */
+static inline mat4_t __attribute__ ((warn_unused_result)) Mat4_Invert(const mat4_t a) {
```
jdolan, line 269: Probably name this Mat4_Inverse.

src/matrix.h
```diff
@@ -20,96 +21,465 @@
...
+		0.f,					-2.f * bt,				0.f,						0.f,
+		0.f,					0.f,					2.f * nf,					0.f,
+		(left + right) * lr,	(top + bottom) * bt,	(farval + nearval) * nf,	1.f
+	});
+}
+
+/**
+ * @return A matrix constructed from the specified translation, Euler angles and scale.
+ */
+static inline mat4_t __attribute__ ((warn_unused_result)) Mat4_FromOriginAnglesScale(const vec3_t origin, const vec3_t angles, const float scale) {
```
jdolan, line 166: Most of the functions use 'translate' instead of 'origin'. Mat4_FromTranslationRotationScale? That would be consistent with the functions below that do just one of those things.

src/matrix.h
```diff
@@ -20,96 +21,465 @@
...
+ * @return A scale matrix.
+ */
+static inline mat4_t __attribute__ ((warn_unused_result)) Mat4_FromScale(const float scale) {
+	return Mat4_FromScale3(Vec3(scale, scale, scale));
+}
+
+/**
+ * @brief Fetch the three directional vectors and/or translation from this matrix.
+ */
+static inline void Mat4_ToVectors(const mat4_t in, vec3_t *forward, vec3_t *right, vec3_t *up, vec3_t *translation) {
```
jdolan, line 376: Mat4_Vectors would be consistent with Vec3_Vectors.

src/matrix.h
```diff
@@ -20,96 +21,465 @@
...
+		v.x * m.m[0][0] + v.y * m.m[1][0] + v.z * m.m[2][0] + m.m[3][0],
+		v.x * m.m[0][1] + v.y * m.m[1][1] + v.z * m.m[2][1] + m.m[3][1],
+		v.x * m.m[0][2] + v.y * m.m[1][2] + v.z * m.m[2][2] + m.m[3][2]
+	);
+}
+
+/**
+ * @return The transformed positive distance plane (A*x+B*y+C*z-D=0).
+ */
+static inline vec4_t __attribute__ ((warn_unused_result)) Mat4_TransformQuakePlane(const mat4_t in, const vec3_t n, float d) {
```
jdolan, line 421: We can probably just call this Mat4_TransformPlane.

jdolan: And make it take in a double for the distance, so that Quemap can use it.

src/collision/cm_trace.h
```diff
@@ -26,10 +26,67 @@
...
+
+	if (trace.fraction < 1.0f) { // transform the impacted plane
+		trace.plane = Cm_TransformPlane(matrix, &trace.plane);
+	}
+
+	// and calculate the final end point
+	trace.end = Vec3_Mix(start, end, trace.fraction);
+
+	return trace;
+}
```
jdolan, line 89: 👍

jdolan: This is in the header just to inline the pre/post trace transforms? I'm not sure that's necessary.

src/collision/cm_test.h
```diff
@@ -34,13 +32,51 @@ static inline float Cm_DistanceToPlane(const vec3_t point, const cm_bsp_plane_t
...
+ * @param head_hode The BSP head node to recurse down.
+ * @param inverse_matrix The inverse matrix of the entity to be tested.
+ *
+ * @return The contents mask at the specified point.
+ */
+static inline int32_t Cm_TransformedPointContents(const vec3_t p, int32_t head_node, const mat4_t inverse_matrix) {
+	const vec3_t p0 = Mat4_Transform(inverse_matrix, p);
+	return Cm_PointContents(p0, head_node);
+}
+
```
jdolan, line 79: I'm not sure I like this in the header, either.

jdolan (changes requested):
I'd like to keep the trace functions in the .c files. Otherwise, some minor comments, but this is great. How extensively have you tested this? Have you baked maps with it and checked for lighting issues?

src/matrix.h
Paril: They can't be put at the end if it's a definition, only a declaration, that's why I moved them to the start.

src/matrix.h
Paril: Good point

src/collision/cm_test.h
Paril: I'm just worried these won't inline properly without it, and passing a mat4 by value might be expensive, but it's less ugly/consistent with our vector system where we don't pass them via pointer. I can move them back though, and we can see what happens.

src/matrix.h
jdolan: Oh! Weird!

src/matrix.h
Paril: Quemap doesn't use this function at all, so it seems to be okay to keep to float. Does it have its own version maybe?

jdolan-chwy: Ah, no, I guess it doesn't need this. Makes sense... Quemap lights inline models in their natural (0, 0, 0) positions, so no transforms are required to accurately do lighting traces.

jdolan (approved):

PR #628, opened 2021-03-25
Paril/bounds type

Paril:
Not a fan of the naming of a few of these functions, but here she is.


src/ai/default/ai_main.c
```diff
@@ -714,8 +714,8 @@ static inline float Ai_Wander(g_entity_t *self, pm_cmd_t *cmd) {
 	vec3_t forward;
 	Vec3_Vectors(Vec3(0.f, *angle, 0.f), &forward, NULL, NULL);
 
-	const vec3_t end = Vec3_Fmaf(self->s.origin, (self->maxs.x - self->mins.x) * 2.0f, forward);
-	const cm_trace_t tr = aim.gi->Trace(self->s.origin, end, Vec3_Zero(), Vec3_Zero(), self, CONTENTS_MASK_CLIP_PLAYER);
+	const vec3_t end = Vec3_Fmaf(self->s.origin, Bounds_Width(self->bounds) * 2.0f, forward);
```
jdolan-chwy, line 717: This might be clearer as Bounds_Size().x imo.

src/client/renderer/r_entity.c
```diff
@@ -26,25 +26,7 @@
...
-		Vec3(e->model->maxs.x, e->model->maxs.y, e->model->maxs.z),
-		Vec3(e->model->mins.x, e->model->maxs.y, e->model->maxs.z),
-	};
-
-	for (size_t i = 0; i < lengthof(corners); i++) {
-		const vec3_t corner = Mat4_Transform(e->matrix, corners[i]);
-		e->abs_model_mins = Vec3_Minf(e->abs_model_mins, corner);
-		e->abs_model_maxs = Vec3_Maxf(e->abs_model_maxs, corner);
-	}
+	e->abs_model_bounds = Mat4_TransformBounds(e->matrix, e->model->bounds);
```
jdolan-chwy, line 29: Are you sure this correctly handles rotations? The above code took me quite some time to sort out. The code isn't rocket science, but due to rotations, it's not as easy as "subtract mins and add maxs from / to origin." Getting this wrong will mean tons of regressions around culling, occluding, etc.

src/client/renderer/r_model.c
```diff
@@ -77,8 +77,7 @@ r_model_t *R_LoadModel(const char *name) {
 
 		mod->type = format->type;
 
-		mod->mins = Vec3_Mins();
-		mod->maxs = Vec3_Maxs();
+		mod->bounds = Bounds_Infinity();
```
jdolan-chwy, line 80: This could probably be just Box() imo. It's not really infinity, right? It's +/-MAX_WORLD_COORD in all directions?

jdolan-chwy (approved):
A few comments. I assume that, in general, you'd prefer Box over Bounds? I'm onboard with that. I'm a little leery of Box_Width, Box_Height, etc, mostly because X and Y are easy to mix up. I think Box_Size().x might be good enough and clearer. I also just wanted to make sure that the Matrix_TransformBounds function doesn't cause regression, see my comment there.

Overall, this is great. Really tidies up a lot of our code.

src/client/renderer/r_model.c
Paril: I'm down to rename this to Box(). It *is* infinity, but perhaps Box() makes sense for a "this box literally contains nothing, not even 0,0,0, it's space that cannot physically exist in reality".

Paril:
Did the thingies

src/cgame/cgame.h
```diff
@@ -800,11 +799,11 @@ typedef struct cg_import_s {
 	
 	/**
 	 * @brief Draw a 3D bbox at the given coordinates.
-	 * @param mins Maxs.
-	 * @param maxs Mins.
-	 * @param color Color of lines.
+	 * @param bounds Box.
```
jdolan-chwy, line 802: Param should be `box`

jdolan-chwy (approved):

PR #630, opened 2021-04-07
Hotfix: glFinish for low framerates so that frames are not dropped.

jdolan:


src/client/cl_main.c
```diff
@@ -663,7 +663,7 @@ void Cl_Frame(const uint32_t msec) {
 
 	Cl_UpdateScreen();
 
-	R_EndFrame();
+	R_EndFrame(cl.frame_msec > 8);
```
Paril, line 666: Interesting idea, I worry that this might cause a cascade of forced glFinish if one frame ends up being too slow and all subsequent frames are stuck being finish'd though.

Issue #617, opened 2021-02-12
Unique plane issue with bmodels (see tokays)


Paril:
https://i.imgur.com/nv39IB8.jpeg

jdolan-chwy:
![image](https://user-images.githubusercontent.com/77992833/107793046-dfbe3680-6d23-11eb-8057-85377c8a8c0b.png)

A workaround is in place for this with the latest tokays.map. But yes, this should be addressed, and the old .map is a good test case.

jdolan:
Demoting to minor, as this affects just one map, and there is an easy workaround.

jdolan:
I actually think the problem here is that node faces are not cleared in between ProcessModels calls, or that MakeTreePortals doesn’t limit to the brushes of the current entity or something. I don’t think that using unique planes per entity will fix this. Planes are not part of bsp_model_t. Only nodes and faces are.

PR #632, opened 2021-04-11
#617 unique planes per inline model

jdolan:
@Paril This is in a branch because I'm a little worried about regression.

The fix to ensure we have unique planes for each BSP inline model is the one-liner in map.c to `memset` the plane hash. But I immediately noticed that maps failed to load with this change, because brush sides were referencing invalid planes. How could that be?

Turns out, there are two instances of brush beveling code, for some reason. One instance is in map.c, AddBrushBevels, the function we all know and love (I hate that code and have yet to make it pretty). But through debugging this, I realized that there's another spot where the code adds bevels: when it is emitting the brushes into the BSP file!

This happens very, very late in the process. In fact, this is even **after** planes have already been written to disk. Any new planes created by this bevel routine would not be included in the BSP. For all of these years, we've just apparently gotten lucky that no late-bevels needed to create new planes. I'm not sure there's anything mathematically guaranteeing that would be the case..

So I decided to tack that late-stage bevel code onto the map.c bevel code. This appears to work as expected, but.. anytime we're dealing with BSP format and collision model, I've learned to be cautious. So here it is, for your review.

### DIFF
--- a/src/tools/quemap/map.c
+++ b/src/tools/quemap/map.c
@@ -391,6 +391,45 @@ static void AddBrushBevels(brush_t *b) {
 			}
 		}
 	}
+
+	// jdolan: This code used to be in EmitBrushes. It added additional bevels (and planes)
+	// late, after portals, after planes were written to disk, etc.. It was bad news.
+
+	// add any axis planes not contained in the brush to bevel off corners
+	for (int32_t axis = 0; axis < 3; axis++) {
+		for (int32_t side = -1; side <= 1; side += 2) {
+			// add the plane
+			vec3_t normal = Vec3_Zero();
+			normal.xyz[axis] = side;
+
+			float dist;
+			if (side == -1) {
+				dist = -b->bounds.mins.xyz[axis];
+			} else {
+				dist = b->bounds.maxs.xyz[axis];
+			}
+
+			const int32_t plane_num = FindPlane(normal, dist);
+
+			int32_t j;
+			for (j = 0; j < b->num_sides; j++) {
+				if (b->sides[j].plane_num == plane_num) {
+					break;
+				}
+			}
+
+			if (j == b->num_sides) {
+				if (num_brush_sides == MAX_BSP_BRUSH_SIDES) {
+					Com_Error(ERROR_FATAL, "MAX_BSP_BRUSH_SIDES\n");
+				}
+
+				s = &b->sides[b->num_sides++];
+				s->plane_num = plane_num;
+				s->texinfo = b->sides[j - 1].texinfo;
+				num_brush_sides++;
+			}
+		}
+	}
 }
 
 /**
@@ -768,6 +807,8 @@ static entity_t *ParseEntity(parser_t *parser) {
 		if (num_entities == MAX_BSP_ENTITIES) {
 			Com_Error(ERROR_FATAL, "MAX_BSP_ENTITIES\n");
 		}
+
+		memset(plane_hash, 0, sizeof(plane_hash));
 
 		entity = &entities[num_entities];
 		num_entities++;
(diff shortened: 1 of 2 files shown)


src/tools/quemap/map.c
```diff
@@ -769,6 +808,8 @@ static entity_t *ParseEntity(parser_t *parser) {
 			Com_Error(ERROR_FATAL, "MAX_BSP_ENTITIES\n");
 		}
 
+		memset(plane_hash, 0, sizeof(plane_hash));
+
```
jdolan, line 812: Were it not for the bevel nonsense, this line alone would have been enough to guarantee unique planes for each inline model, heh.

src/tools/quemap/writebsp.c
```diff
@@ -29,19 +29,21 @@
...
-		out->dist = (float) p->dist;
+		out->normal = in->normal;
+		out->dist = (float) in->dist;
 
 		bsp_file.num_planes++;
 
 		Progress("Emitting planes", 100.f * i / num_planes);
 	}
+
+	return bsp_file.num_planes;
```
jdolan, line 46: Purely cosmetic.

src/tools/quemap/writebsp.c
```diff
@@ -376,42 +378,6 @@ static void EmitBrushes(void) {
...
-				normal.xyz[axis] = side;
-
-				float dist;
-				if (side == -1) {
-					dist = -b->bounds.mins.xyz[axis];
-				} else {
-					dist = b->bounds.maxs.xyz[axis];
-				}
-
-				const int32_t plane_num = FindPlane(normal, dist);
```
jdolan, line 393: So here is the late-stage beveling code. This way too late to be doing this, in my opinion. The BSP is finalized at this point, and we're just flushing it to disk. FindPlane can and does create new planes if necessary, and in this instance, those planes would never make it to the BSP. So I thought it best to relocate this code to AddBrushBevels.

Paril:
I dunno if the plan is to touch AddBrushBevels here, but Q3A did a little adjustment to brush bevels: they ensured that the first six brush sides/planes were always axial, which let them skip some checks in the tracing code. While it likely won't matter for most cases, it might help the new rotation code perhaps since it's a bit heavier?

I recompiled all of the maps locally, going to test them all out and see if I notice anything off. I notice no warnings while compiling (except for 2 benign "entity outside of map" warnings on Tokays, and this one in Warehouse:
```
WARNING: BrushContents: Entity 0, Brush 167: Mixed face contents: 1 expected 1000000
worldspawn @ (0.00 0.00 0.00)
```
)

Paril:
image
Edge has some black spots here

Paril:
image
Some weird lighting here, not sure if this was like this before or not. Just -bsp -light at command line.

Paril:
image
image
image

Fractures. On last image, left has texture misalignments, right has dark faces

Paril:
image
https://streamable.com/bznctv lighting on bmodels here seems all wrong, but I can't put my finger on it.
image
image

Paril:
image

Paril:
image
I think this should be enough info to look into the bugs. Code-wise it looks solid but there's still somethin that isn't working properly.

Paril:
image
Just noticed this one by chance while testing a different bugfix

jdolan:
The black faces and broken planes are a consequence of fog entities getting unique planes, but then being merged into worldspawn. This creates duplicate planes which conflict with each other during BSP tree creation.

Fixing this will be more involved. Since this is a low priority, affecting just one map, and with an easy workaround fix, I'm going to abandon this attempt, to possibly revisit it later. I thought it would be an easy win, but.. as with most things in the BSP compiler, nope.

PR #635, opened 2021-04-11
Use simple bevels, and ensure that brush sides are sorted axial first.

jdolan:
@Paril Here you go, here is a vastly simplified bevels implementation that:

1. Just ensures that all brushes have 6 axially aligned sides, some of which may be bevels.
2. Ensures that those 6 axially aligned sides are first in every brush.

While I see a few minor artifacts from this change (one small missing stair on Edge), I'm not overly concerned. Because we no longer subdivide the world into 1024x cubes, Edge has been a constant battle with missing faces due to the portal complexity of the large round rooms. I'm confident I can fix this missing stair by just rejiggering hints, and this is not necessarily a new regression.

Let me know if you think this is a good idea, and if this will allow you to port those axial optimizations from Q3's trace.

### DIFF
--- a/src/tools/quemap/map.c
+++ b/src/tools/quemap/map.c
@@ -240,157 +240,62 @@ static int32_t BrushContents(const brush_t *b) {
 }
 
 /**
+ * @brief qsort comparator to sort a brushes sides by plane type. This ensures
+ * the first 6 sides of the brush are axial.
+ */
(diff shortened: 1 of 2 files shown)


jdolan:
Actually, I just fixed a couple brush issues on Edge, and it compiles and runs perfectly with this change.

PR #641, opened 2021-05-30
Feature/faces reference brush sides

jdolan:
Something that always bothered me about the Quake BSP format is that faces had no relationship to the brush sides that created them. This patch fixes that. Faces no longer duplicate plane, texinfo and contents. Instead, they point to the original brush side in the map file that created them. This is not only a more efficient format, but it allows us to do more deterministic and WYSIWYG face merging for flares, etc.

There are issues with blended faces in this patch so far. That's why it's WIP. I think it has to do with portal / plane sidedness.

This patch will require no effort from map authors, but it will require a rebake of all maps (BSP format change).


jdolan:
Made some more progress on this yesterday. I'm also cleaning up a lot of the collision and game API to deal with brush sides instead of passing planes and texture references around. It's coming together nicely. Did not spend any time on the blended faces issue tho..

jdolan:
This is now ready for review.

src/tools/quemap/portal.c
```diff
@@ -691,16 +691,16 @@ static face_t *FaceFromPortal(portal_t *p, int32_t pside) {
 
 	face_t *f = AllocFace();
 
-	f->texinfo = side->texinfo;
-	f->plane_num = (side->plane_num & ~1) | pside;
+	f->brush_side = side;
+//	f->plane_num = (side->plane & ~1) | pside;
```
jdolan, line 695: Hmmm.

src/tools/quemap/portal.c
```diff
@@ -691,16 +691,16 @@ static face_t *FaceFromPortal(portal_t *p, int32_t pside) {
...
 	f->portal = p;
 
 	if (pside) {
 		f->w = Cm_ReverseWinding(p->winding);
-		f->contents = p->nodes[1]->contents;
+//		f->contents = p->nodes[1]->contents;
```
jdolan, line 700: Hmmmmmm.

src/collision/cm_types.h
```diff
@@ -373,3 +312,50 @@ typedef struct {
...
+	/**
+	 * @brief The impacted brush side, or `NULL`.
+	 */
+	const cm_bsp_brush_side_t *side;
+
+	/**
+	 * @brief The impacted entity, or `NULL`.
+	 */
+	struct g_entity_s *ent; // not set by Cm_*() functions
+} cm_trace_t;
```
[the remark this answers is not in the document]
Paril: Ideally everything that doesn't need to be a pointer/handle shouldn't be. This was a big gripe of mine with getting cvars and trace surfaces handled properly in WASM modules. It makes things unnecessarily annoying. It's fine for a native API I guess.
The duplicate plane thing is definitely something worth fixing, though, whether it be through pseudo-inheritance (like how Q2 extended the surface name length internally without breaking the game API) or just by changing the design around.

PR #646, opened 2021-12-11
Feature/622 bloom via materials

jdolan:


src/client/renderer/r_bloom.c
```diff
@@ -0,0 +1,179 @@
...
+	GLint in_texcoord;
+
+	GLint texture_color_attachment;
+	GLint texture_bloom_attachment;
+} r_bloom_program;
+
+/**
+ * @brief
+ */
+void R_DrawBloom(const r_view_t *view) {
```
jdolan, line 58: So this is pretty straightforward:

1. Keep rendering to the framebuffer's primary color attachment
2. Bind both the color attachment and bloom attachment to textures
3. Draw a full-screen quad, that re-writes the color attachment and blurs and blends the bloom attachment over it.

src/client/renderer/r_bsp_draw.c
```diff
@@ -452,9 +450,7 @@ static void R_DrawBspInlineModelAlphaTestDrawElements(const r_view_t *view,
 			continue;
 		}
 		
-		R_TIMER_WRAP(va("AlphaTestDrawElem: %s", draw->material->media.name),
```
jdolan, line 455: I took the timers stuff out @Paril 😬 It was driving me nuts with X-code's indentation and syntax highlighting being completely hosed, plus breakpoints not going where I put them. I removed the timers in a separate commit so if we want to re-add it should be relatively easy.

src/client/renderer/r_framebuffer.c
```diff
@@ -24,56 +24,60 @@
...
 	r_framebuffer_t framebuffer = {
 		.width = width,
 		.height = height,
-		.multisample = multisample
 	};
 
 	glGenFramebuffers(1, &framebuffer.name);
 	glBindFramebuffer(GL_FRAMEBUFFER, framebuffer.name);
 
-	if (multisample) {
```
jdolan, line 38: I removed the renderbuffer support here, because to do post-processing, we need to render to textures, not renderbuffers. You can't bind and sample renderbuffers -- that's actually their point. OpenGL can make some optimizations to renderbuffers, *because* you'll never ask to read them. That's incompatible with post-effects.

It's not clear to me if, in removing this, I also broke multisample entirely? It still runs without complaining, and the game looks normal. I couldn't immediately tell if it was antialiasing or not.

Issue #546, opened 2018-03-14
In-game menus for Join, Vote, etc


jdolan:
As a player, I want to use a menu system for joining teams, voting, etc. As a modder, I want to be able to easily extend the default in-game menus.

kaadmy:
The buttons are in the menu already while ingame, they just don't open the corresponding options/dialogs yet.

jdolan:
Yea. I think there's still a little bit of work to figure out how to communicate available votes and whatnot from the server. Config strings are currently used, and that probably makes. But we might need to expand on them (all available teams and their skins and their names, for example?)

Agreed this won't be brain surgery, but we just need to do the work. It'll make a huge difference to the experience.

jdolan:
Also I'd like to get your HUD stuff merged down in before we muck with this too much, as I could see some potential conflicts otherwise.

jdolan:
I decided to kill off the spaghetti that was our voting code, rather than try to slap a UI on it (nightmare).

We could introduce intermission time next_map voting, which is the only thing players would ever actually use, with an entirely different approach.

PR #651, opened 2022-01-03
#546 Teams menu (previously "Join")

jdolan:
This PR wires up the secondary menu button (bottom row) to allow the player to join a team once connected. The vast majority of the work here was actually to clean up and enhance the client's representation of other connected clients:

1. `cl_client_info_t` is now `cg_client_info_t` and is implemented entirely in cgame.
2. Player model sounds had to be migrated from `sound` to `cgame`, because they relied on `cl_client_info_t`. This is actually much cleaner and more flexible, anyway.
3. This nudged me to also move all sound sending and receiving to `game` and `cgame` rather than `server` and `client`. Now mods can extend this packet to add more sound effects, whatever they like.
4. The client info struct and client info strings have been expanded to include a team ID.
5. Found and fixed a nasty issue with the bloom program, and improved framebuffer management in the process.
6. Implemented the teams menu 😂


jdolan:
The menu itself is not quite done. Some layout quirks, and I haven't wired up the team selection action. But the plumbing is all there, and there's so much else in this PR that I figured I'd open it up in case you'd like to review.

PR #660, opened 2022-01-17
This PR simplifies Cm_TransformedBoxTrace and improves readability.

jdolan:
In studying the new Cm_TransformedBoxTrace, which works great btw, I saw at least a couple spots where we were transforming things redundantly, or not at the most opportune time. 

I got to thinking.. couldn't we just un-transform the inputs, and re-transform the outputs, and then not do any transforming inside the actual trace?

It turns out, we can. And then we don't even need to track `is_transformed` or either of the matrices beyond the entrypoint. In addition to being more readable, this is also more performant. We never touch node or brush side planes inside of the trace routine.

I tested this extensively: pmove on Edge stairs, func_rotating on Pits and Lava Tomb, and multiple map bakes with quemap. Everything appears to be working just as well as `develop`.

Please let me know what you think of these changes.

### DIFF
--- a/src/collision/cm_trace.c
+++ b/src/collision/cm_trace.c
@@ -24,7 +24,7 @@
 /**
  * @brief Box trace data encapsulation and context management.
  */
-static __thread struct {
+typedef struct {
 	/**
 	 * @brief The trace start and end points, as provided by the user.
 	 */
@@ -39,6 +39,11 @@ static __thread struct {
 	 * @brief The head node, as provided by the user.
 	 */
 	int32_t head_node;
+
+	/**
+	 * @brief The contents mask to collide with, as provided by the user.
+	 */
+	int32_t contents;
 
 	/**
 	 * @brief The absolute bounds of the trace, spanning the start and end points.
@@ -56,35 +61,30 @@ static __thread struct {
 	vec3_t offsets[8];
 
 	/**
-	 * @brief The contents mask to collide with, as provided by the user.
-	 */
-	int32_t contents;
-
-	/**
 	 * @brief The brush cache, to avoid multiple tests against the same brush.
 	 */
 	int32_t brush_cache[128];
 
 	/**
-	 * @brief The trace result.
-	 */
-	cm_trace_t trace;
-
-	/**
 	 * @brief The trace fraction not taking any epsilon nudging into account.
 	 */
 	float unnudged_fraction;
-} data;
+
+	/**
+	 * @brief The trace result.
+	 */
+	cm_trace_t trace;
+} cm_trace_data_t;
 
 /**
  * @brief
  */
-static inline _Bool Cm_BrushAlreadyTested(int32_t brush_num) {
-	const int32_t hash = brush_num & (lengthof(data.brush_cache) - 1);
+static inline _Bool Cm_BrushAlreadyTested(cm_trace_data_t *data, int32_t brush_num) {
+	const int32_t hash = brush_num & (lengthof(data->brush_cache) - 1);
 
-	const _Bool skip = (data.brush_cache[hash] == brush_num);
+	const _Bool skip = (data->brush_cache[hash] == brush_num);
 
-	data.brush_cache[hash] = brush_num;
+	data->brush_cache[hash] = brush_num;
 
 	return skip;
 }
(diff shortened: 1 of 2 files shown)


src/collision/cm_trace.c
```diff
@@ -24,7 +24,7 @@
 /**
  * @brief Box trace data encapsulation and context management.
  */
-static __thread struct {
+typedef struct {
```
jdolan, line 27: I've read in many places that TLS is slow -- probably slower than passing an additional parameter down the call chain. I also feel that, stylistically, param passing is cleaner. Was there a compelling reason to use TLS? I changed this as a separate commit, so it's easy to undo if necessary.

src/collision/cm_trace.c
```diff
@@ -55,81 +60,45 @@ static __thread struct {
...
-	data.brush_cache[hash] = brush_num;
+	data->brush_cache[hash] = brush_num;
 
 	return skip;
 }
 
-/**
- * @brief 
- */
-static inline _Bool Cm_TraceIntersect(const cm_bsp_brush_t *brush) {
```
jdolan, line 110: Because transforms are done before / after the trace, we never have to adjust brush bounds. So this is just `Box3_Intersects(data->abs_bounds, brush->bounds)` now.

src/collision/cm_trace.c
```diff
@@ -197,53 +160,47 @@ static void Cm_TraceToBrush(const cm_bsp_brush_t *brush) {
...
-			data.trace.surface = side->surface;
-			data.trace.material = side->material;
+		if (enter_fraction > -1.f && enter_fraction < data->unnudged_fraction && nudged_enter_fraction < data->trace.fraction) {
+			data->unnudged_fraction = enter_fraction;
+			data->trace.fraction = nudged_enter_fraction;
+			data->trace.brush_side = side;
+			data->trace.plane = plane;
+			data->trace.contents = side->contents;
+			data->trace.surface = side->surface;
+			data->trace.material = side->material;
```
jdolan, line 178: No material changes in here. This is all just TLS -> param.

src/collision/cm_trace.c
```diff
@@ -414,130 +371,112 @@ next:;
...
-	memset(data.brush_cache, 0xff, sizeof(data.brush_cache));
-
 	// check for position test special case
-	if (Vec3_Equal(data.start, data.end)) {
+	if (Vec3_Equal(data->start, data->end)) {
 		static __thread int32_t leafs[MAX_BSP_LEAFS];
-		box3_t abs_bounds = data.abs_bounds;
 
-		if (data.is_transformed) {
-			abs_bounds = Mat4_TransformBounds(data.inverse_matrix, abs_bounds);
```
jdolan, line 464: This is always done up-front now.

src/collision/cm_trace.c
```diff
@@ -414,130 +371,112 @@ next:;
...
 	}
 
-	if (data.is_transformed) {
-		data.size = Box3_Symetrical(Mat4_TransformBounds(data.inverse_matrix, Box3_Expand(data.bounds, BOX_EPSILON)));
-		Cm_TraceToNode(data.head_node, 0.f, 1.f, Mat4_Transform(data.inverse_matrix, data.start), Mat4_Transform(data.inverse_matrix, data.end));
-	} else {
-		data.size = Box3_Symetrical(Box3_Expand(data.bounds, BOX_EPSILON));
-		Cm_TraceToNode(data.head_node, 0.f, 1.f, data.start, data.end);
-	}
+	Cm_TraceToNode(data, data->head_node, 0.f, 1.f, data->start, data->end);
```
jdolan, line 420: Similarly, all of these transforms, if required, are done in Cm_TransformedBoxTrace.

src/collision/cm_trace.c
```diff
@@ -414,130 +371,112 @@ next:;
...
+	Cm_BoxTrace_(&data);
+
+	data.trace.end = Mat4_Transform(matrix, data.trace.end);
 
-	return Cm_BoxTrace_();
+	const vec4_t p = Mat4_TransformPlane(matrix, data.trace.plane.normal, data.trace.plane.dist);
+	data.trace.plane.normal = Vec4_XYZ(p);
+	data.trace.plane.dist = p.w;
+	
+	return data.trace;
```
jdolan, line 479: Un-transform start and end, and use those to calculate abs_bounds so that it too is un-transformed. Then run the trace. Finally, transform the trace end point, and the impacted plane.

Paril:
Un-transforming start/end and then back again is what Q2 originally did, which causes traces to not "look" correct because we have axis-aligned bounding boxes and the un-transformation doesn't take that into account, so stuff sinks into or hovers above rotating objects. That's why I did the plane stuff, which is what Xotonic does as well as it actually handles bboxes accurately.




This was the issue I was destined to fix; these shells are both "on-ground" in Q2 (and in your latest commits here).

jdolan:
Closing this PR, but will keep the branch around for posterity.

PR #665, opened 2022-02-02
Discord Rich Presence

Paril:
Had to re-make the repo because it was so old.


jdolan (commented):
This is beautiful.  I’ll see if I can get it building on Mac and Linux tonight.

jdolan (approved):

PR #679, opened 2023-01-15
Programmatically generate occlusion queries

jdolan:
The goal of this PR is to completely remove the burden of placing occlusion queries from the mapper, while also improving performance. This approach is inspired by this article: https://developer.nvidia.com/gpugems/gpugems2/part-i-geometric-complexity/chapter-6-hardware-occlusion-queries-made-useful

The general idea is to select nodes from the BSP that satisfy a heuristic (some general size threshold, configurable by cvar), and create queries spanning the visible bounds of those nodes. Since the selection produces a completely contiguous graph, where every valid point in the map is covered by exactly one query, the actual occlusion check changes from:

    find a query that fully contains the box in question, and if that query is occluded, we can skip the object

To:

    find all queries that intersect the box in question, and if all of those queries are occluded, we can skip the object

This is actually a significant improvement for things like dynamic lights and even large sprites which often exceed the solid world bounds (where the mapper had likely not placed an occlusion query).

Lastly, because the occlusion queries are derived from the BSP tree, we can make some optimizations when we receive a query result: a positive result (visible) should propagate up the tree, and a negative result (occluded) should propagate down. This reduces the latency of query results by a frame or two, and allows us to skip certain queries entirely for certain frames. There's still more work to be done to improve this optimization.

Assuming we stick with this approach, I'll update the code and editor configurations to remove all occurrences of `CONTENTS_OCCLUSION_QUERY` since it will no longer be required / supported. This will require stripping these brushes from our maps, too.

This change itself is a BSP format update, because I extended the BSP node struct to include visible bounds (which is often a subset of collision bounds).

### DIFF
--- a/src/client/renderer/r_occlude.c
+++ b/src/client/renderer/r_occlude.c
@@ -22,15 +22,9 @@
 #include "r_local.h"
 
 /**
- * @brief OpenGL occlusion queries persist across multiple frames due to their implicit latency.
+ * @brief Hardware occlusion queries.
  */
 static struct {
-	/**
-	 * @brief The array of known queries.
-	 */
-	r_occlusion_query_t queries[MAX_OCCLUSION_QUERIES];
-	int32_t num_queries;
-
 	/**
 	 * @brief The backing vertex array.
 	 */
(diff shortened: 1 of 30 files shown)


src/client/renderer/r_occlude.c
```diff
@@ -150,7 +242,7 @@ void R_UpdateOcclusionQueries(r_view_t *view) {
 	}
 
 	if (view->flags & VIEW_FLAG_NO_DELTA) {
-		view->num_occlusion_queries = r_occlusion.num_queries = 0;
+		//view->num_occlusion_queries = 0;
```
jdolan, line 245: Right, still need to handle this somehow..

src/client/renderer/r_occlude.c
```diff
@@ -101,43 +124,112 @@ _Bool R_CulludeSphere(const r_view_t *view, const vec3_t point, const float radi
...
+		.result = 1,
+	};
 
-	if (R_CullBox(view, bounds)) {
-		return;
+	glGenQueries(1, &query.name);
+
+	// FIXME: This indexing is ghetto, but should be safe on most GL drivers
+	assert(query.name < MAX_OCCLUSION_QUERIES);
+	Box3_ToPoints(bounds, r_occlusion.vertexes[query.name]);
```
jdolan, line 139: @Paril Do you think this is an issue? Using the query object name as its index into an array? Every driver I've seen generates sequential object names from 1 onward. I'm doing this because I'd like to have linear storage for the query vertexes to simplify uploading and drawing them.

src/client/renderer/r_stain.c
```diff
@@ -105,28 +105,25 @@ static void R_StainNode(const r_stain_t *stain, const r_bsp_node_t *node) {
 		return;
 	}
 
-	if (node->contents & CONTENTS_MASK_SOLID) {
```
jdolan, line 108: This contents check is useless and misleading. Nodes are always `CONTENTS_NODE` (-1), so this always passes. We are not staining leafs; we are staining nodes.

PR #680, opened 2023-02-09
HDR and miscellaneous lighting fixes

jdolan:
- Quemap now generates high dynamic range RGB32F lightmap and lightgrid diffuse textures
- Quetoo uses RGBA32F framebuffers for all rendering so that HDR is fully supported
- Bloom is rewritten to use ping-pong FBO technique with configurable iterations (`r_bloom_iterations`)
- Bloom is moved to a proper post-processing pass (`r_post 1`, enabled by default)
- [ACES tonemapping](https://knarkowicz.wordpress.com/2016/01/06/aces-filmic-tone-mapping-curve/) is performed as the very last step in the post processing pass
- Fix dynamic lights and shadows on decals
- Allow mapper to define patch light attenuation, cone and angular falloff
- Allow mapper to define spotlight falloff
- Don't pre-multiply light intensity
- Calculate luxel distance / dir to winding for patch lights (fixes hot spots around large patch lights)
- Calculate luxel "plane" equation for directional light sidedness check
- Remove alpha blended light direction hack (fixes hot spots on blended faces)
- Remove patch.c, since `Cm_DistanceToWinding` provides much better looking results without subdividing face lights


jdolan:
Just a quick WIP screen. This is just a rebake of dm2, without any changes to the lighting parameters. You can see how HDR really helps bloom "shine" (no pun intended).

quetoo061

src/shared/box.h
```diff
@@ -145,11 +145,11 @@ static inline void Box3_ToPoints(const box3_t bounds, vec3_t *points) {
 */
 static inline _Bool __attribute__ ((warn_unused_result)) Box3_Intersects(const box3_t a, const box3_t b) {
 
-	if (a.mins.x >= b.maxs.x || a.mins.y >= b.maxs.y || a.mins.z >= b.maxs.z) {
+	if (a.mins.x > b.maxs.x || a.mins.y > b.maxs.y || a.mins.z > b.maxs.z) {
 		return false;
 	}
 
-	if (a.maxs.x <= b.mins.x || a.maxs.y <= b.mins.y || a.maxs.z <= b.mins.z) {
+	if (a.maxs.x < b.mins.x || a.maxs.y < b.mins.y || a.maxs.z < b.mins.z) {
```
jdolan, line 152: I'm not sure about this change. I might revert this. My goal was to make Box3_Intersect "consistent" with Box3_Contains. Box3_Contains says that edge equality means that two boxes contain each other, while Box3_Intersects was returning false in that scenario. @Paril do you think this matters / is worth fixing / will break things?

src/client/renderer/r_context.c
```diff
@@ -264,10 +264,12 @@ void R_InitContext(void) {
 
 	Com_Print("  Trying %dx%d..\n", w, h);
 
-	SDL_GL_SetAttribute(SDL_GL_RED_SIZE, 8);
-	SDL_GL_SetAttribute(SDL_GL_GREEN_SIZE, 8);
-	SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE, 8);
-	SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, 8);
+	SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, 1);
+
+	SDL_GL_SetAttribute(SDL_GL_RED_SIZE, 32);
```
jdolan, line 269: Is there any merit to using full 32 bit floats here? Would we see a difference with 16 bit "half" floats?

src/client/renderer/r_framebuffer.c
```diff
@@ -59,14 +59,14 @@ r_framebuffer_t R_CreateFramebuffer(GLint width, GLint height, int32_t attachmen
 	glBindFramebuffer(GL_FRAMEBUFFER, framebuffer.name);
 
 	if (attachments & ATTACHMENT_COLOR) {
-		framebuffer.color_attachment = R_CreateFramebufferTexture(&framebuffer, GL_RGBA8, GL_RGBA, GL_UNSIGNED_BYTE);
-		framebuffer.color_attachment_copy = R_CreateFramebufferTexture(&framebuffer, GL_RGBA8, GL_RGBA, GL_UNSIGNED_BYTE);
+		framebuffer.color_attachment = R_CreateFramebufferTexture(&framebuffer, GL_RGBA32F, GL_RGBA, GL_FLOAT);
```
jdolan, line 62: Similarly, should these use half floats?

src/client/renderer/shaders/bloom_fs.glsl
```diff
@@ -37,8 +37,6 @@ void main(void) {
 
 	vec3 bloom_color = vec3(0.0);
 	for (int i = 1; i <= bloom_lod; i++) {
-		bloom_color += textureLod(texture_bloom_attachment, vertex.texcoord, i).rgb * 1.f / float(bloom_lod);
+		out_color.rgb += textureLod(texture_bloom_attachment, vertex.texcoord, i).rgb * 1.f / float(bloom_lod);
 	}
-
-	out_color.rgb = invert(invert(out_color.rgb) * invert(bloom_color));
```
jdolan, line 43: Had to revert this to the old method to make floating point happy.

SpineyPete:
For bloom, if you have an hdr framebuffer you can just add (+=) a blurred bloom buffer on top of it *before* tonemapping and that’s it. That way you don’t have to do the invert() stuff to prevent color component saturation in the final result. So tonemapping would be its own post processing shader that gets applied to the frame at the end.

src/cgame/cgame.h
```diff
@@ -642,16 +642,6 @@ typedef struct cg_import_s {
...
 
-	/**
-	 * @brief Blits the framebuffer to the specified rectangle on the screen.
-	 * @param framebuffer The framebuffer to blit.
-	 * @param x The horizontal origin of the screen rectangle in drawable pixels.
-	 * @param y The vertical origin of the screen rectangle in drawable pixels.
-	 * @param w The width of the screen rectangle in drawable pixels.
-	 * @param h The height of the screen rectangle in drawable pixels.
-	 */
-	void (*BlitFramebuffer)(const r_framebuffer_t *framebuffer, GLint x, GLint y, GLint w, GLint h);
```
jdolan, line 653: Hm. I can't remember why I removed this.

src/shared/box.h
jdolan: @Paril bumpity

src/client/renderer/r_image.c
```diff
@@ -161,13 +161,18 @@ void R_Screenshot(r_view_t *view) {
 	switch (r_image_state.screenshot) {
 		case SCREENSHOT_NONE:
 			return;
-		case SCREENSHOT_VIEW:
-			surface = R_ReadFramebuffer(view->framebuffer);
+		case SCREENSHOT_VIEW: {
```
jdolan, line 164: `r_screenshot view` will take a screenshot of the 3D view (minus HUD, crosshair, etc).

src/client/renderer/r_image.c
```diff
@@ -194,24 +199,6 @@ void R_Screenshot_f(void) {
 	}
 }
 
-/**
- * @brief
- */
-static GLenum R_GetInternalImageFormat(const r_image_t *image) {
-
-	switch (image->format) {
```
jdolan, line 202: Let's just stop guessing these all the time, and include it in `r_image_t`.

src/tools/quemap/lightgrid.c
```diff
@@ -154,7 +154,7 @@ static void LightgridLuxel_Ambient(const light_t *light, luxel_t *luxel, float s
 
 	Luxel_Illuminate(luxel, &(const lumen_t) {
 		.light = light,
-		.color = Vec3_Scale(light->color, scale)
+		.color = Vec3_Scale(light->color, light->intensity * scale)
```
jdolan, line 157: Is this why ambient lightgrid sprites look overbright?

PR #690, opened 2025-02-02
Fallback SDL_GL to 24 bit depth if 32 bit unavailable

goodspeed34:
32 bit profiles aren't always available for certain machines, fall back to 24 bit if those are unavailable.

Tested on: Debian Sid, Intel Arc A730M GPU


jdolan (approved):
Very cool. Thanks again!

PR #700, opened 2026-01-09
Clustered forward rendering

jdolan:
Introduce "clustered forward rendering." Which is just a fancy term for a better lightgrid. 

 * Create a 3D texture of the voxel grid that stores offset and count into a linear texture buffer object that contains light indices (like a BSP lump).
 * When rendering, resolve voxel for fragment, then resolve light offset and count, then iterate light indices as texelFetches from the TBO.

For static light sources, this dramatically reduces fragment-to-light checks, since the voxelized data is much tighter than iterating all lights that intersect the bounding box of the current BSP block, etc.

Also, instead of baking caustics as an RGB texture, I instead bake a contents mask texture, which works for caustics and may also come in handy for other effects (think: sprite bounces on GPU etc).

### DIFF
--- a/src/client/renderer/r_occlude.c
+++ b/src/client/renderer/r_occlude.c
@@ -113,7 +113,7 @@ bool R_CulludeSphere(const r_view_t *view, const vec3_t point, const float radiu
 }
 
 /**
- * @brief Allocates an occlusion query with the specified initial bounds.
+ * @brief Allocates an occlusion query with the specified bounds.
  */
 r_occlusion_query_t *R_AllocOcclusionQuery(const box3_t bounds) {
 
--- a/src/client/renderer/r_types.h
+++ b/src/client/renderer/r_types.h
@@ -832,7 +832,7 @@ typedef struct {
 
   r_bsp_voxels_t *voxels;
 
-    /**
+  /**
    * @brief The vertex array (VAO) name.
    */
   GLuint vertex_array;


src/client/renderer/r_bsp_model.c
```diff
@@ -349,39 +349,101 @@ static void R_LoadBspVoxels(r_model_t *mod) {
...
-  out->diffuse->internal_format = GL_RGBA8;
-  out->diffuse->format = GL_RGBA;
-  out->diffuse->pixel_type = GL_UNSIGNED_BYTE;
-
-  glActiveTexture(GL_TEXTURE0 + TEXTURE_VOXEL_DIFFUSE);
-
-  R_UploadImage(out->diffuse, data);
-  data += voxels * sizeof(color32_t);
 
+  // Skip contents data for now (will upload to GPU texture)
```
jdolan: This needs some cleanup and simplification.

src/client/renderer/r_image.c
```diff
@@ -138,51 +138,58 @@ void R_SetupImage(r_image_t *image) {
...
+          image->levels = 1;
+          break;
+      }
+    }
+
+    if (image->depth) {
+      glTexStorage3D(image->target, image->levels, image->internal_format, image->width, image->height, image->depth);
+    } else {
+      glTexStorage2D(image->target, image->levels, image->internal_format, image->width, image->height);
+    }
```
jdolan: Refactored R_SetupImage to support GL_TEXTURE_BUFFER (TBO).

Issue #712, opened 2026-02-28
Add r_draw_scale cvar to control console fonts, HUD, crosshair, etc..


jdolan:
Note that this should NOT affect ObjectivelyMVC, which has its own mechanisms for dealing with pixel density.

PR #713, opened 2026-02-28
#712 r_draw_scale for scaling console, HUD, crosshair, etc.

jdolan:


src/cgame/default/ui/settings/SystemViewController.c
```diff
@@ -34,7 +34,7 @@ static void didSelecVideoMode(Select *select, Option *option) {
 
   const SDL_DisplayMode *mode = option->value;
   if (mode) {
-    if (mode->w != cgi.context->w || mode->h != cgi.context->h) {
+    if (mode->w != cgi.context->window_bounds.w || mode->h != cgi.context->window_bounds.h) {
```
jdolan: I'm actually going to kill the video mode select, anyway. We never change display mode anymore. We run at desktop resolution and control game resolution with r_supersample (I'll probably rename it to r_render_scale to complement r_draw_scale).

src/cgame/default/ui/settings/VideoModeSelect.c
```diff
@@ -81,7 +81,7 @@ static void updateBindings(View *self) {
         char *title = va("%dx%d @ %.0fHz", mode->w, mode->h, mode->refresh_rate);
         $(select, addOption, title, mode);
 
-        if (mode->w == cgi.context->w && mode->h == cgi.context->h) {
+        if (mode->w == cgi.context->window_bounds.w && mode->h == cgi.context->window_bounds.h) {
```
jdolan: Going away.

src/client/renderer/shaders/uniforms.glsl
```diff
@@ -71,11 +71,6 @@ layout (std140) uniform uniforms_block {
    */
   ivec4 viewport;
 
-  /**
-   * @brief The 2D projection matrix.
-   */
-  mat4 projection2D;
```
jdolan: Yanked this out of the block, as it is only used by the draw_2d shader, and we now have two projection matrices (scaled vs unscaled). So just set it directly from that program.

src/client/renderer/r_draw_2d.c
```diff
@@ -83,13 +94,14 @@ static struct {
...
+  r_draw_2d_arrays_list_t ui;
 
-  // accumulated vertexes backing the draw arrays
-  r_draw_2d_vertex_t vertexes[MAX_DRAW_2D_VERTEXES];
-  int32_t num_vertexes;
+  // draw list for in-game elements like console, hud, using r_draw_scale window coordinates
+  r_draw_2d_arrays_list_t game;
+
+  // active draw arrays list
+  r_draw_2d_arrays_list_t *active;
```
jdolan: We now accumulate two draw lists: one for the UI (window space) and one for the game (scaled window space).

Issue #709, opened 2026-02-24
Support limited material stages on sky box


jdolan:
It'd be fun to add support for Quake and Quake3 style "layered" parallax sky boxes. We can probably borrow the material scroll stage from material.glsl. I'm not sure what other effects we would want or need? Also not sure what scrolling texture coordinates will do on a cubemap texture, but I guess we'll find out :)

PR #726, opened 2026-04-09
#709 Limited sky materials support

jdolan:
This is now working, tho there are some areas that could be improved:

1. The sky stages use a projection that is relative to the view origin. This means that sky layers like clouds and things sort of "follow" the player as you walk around the map. It's not super noticeable, but it's also not hard to spot. It would be ideal if this projection used a fixed origin of the map's center or something.

2. The scale and scroll parameters are difficult to get a feel for. They do not behave like "normal" stage scale and scroll. For example, here is the best looking version of the Dark Zone classic Quake purple layered sky I could come up with:

```
{
	diffusemap darkzone/sky
	surface sky
	{
		texture common/sky/sky1_clouds
		blend GL_SRC_ALPHA GL_ONE
		color 1 1 1 .666
		scroll.s -.1
		scroll.t -.1
		scale.s .5
		scale.t .5
	}
	{
		texture common/sky/sky1_clouds
		blend GL_SRC_ALPHA GL_ONE
		color 1 1 1 .666
		scroll.s .05
		scroll.t .05
		scale.s .15
		scale.t .15
	}
}
```

Not sure if there's anything to do about this. It's just a bit unobvious coming from authoring regular material stages.


src/client/renderer/r_bsp_draw.c
```diff
@@ -349,11 +350,6 @@ static void R_DrawOpaqueBspEntity(const r_view_t *view, const r_entity_t *entity
         continue;
       }
 
-      if (R_CulludeBox(view, block->visible_bounds)) {
```
jdolan, line 352: There's no sense re-testing this here. The block has its own occlusion query, which we also frustum-cull in R_UpdateOcclusionQueries. So this test here is completely wasted.

src/client/renderer/r_bsp_draw.c
```diff
@@ -450,10 +446,6 @@ static void R_DrawBlendBspEntity(const r_view_t *view, const r_entity_t *entity)
...
-        continue;
```
jdolan, line 454: Same as previous comment. Should provide a small FPS bump.

src/client/renderer/r_sky.c
```diff
@@ -74,18 +168,14 @@ void R_DrawSky(const r_view_t *view, const r_bsp_model_t *bsp) {
       continue;
     }
 
-    if (R_CulludeBox(view, block->visible_bounds)) {
```
jdolan, line 77: Same comment here. The block's occlusion query accounts for frustum culling as well. So this is entirely wasteful.

src/client/renderer/r_sky.c
```diff
@@ -47,10 +47,102 @@ static struct {
...
+
+  if (stage->cm->flags & STAGE_PULSE) {
+    glUniform1f(r_sky_program.stage.pulse, stage->cm->pulse.hz);
+  }
+
+  if (stage->cm->flags & (STAGE_SCROLL_S | STAGE_SCROLL_T)) {
+    glUniform2f(r_sky_program.stage.scroll, stage->cm->scroll.s, stage->cm->scroll.t);
+  }
+
+  if (stage->cm->flags & (STAGE_SCALE_S | STAGE_SCALE_T)) {
```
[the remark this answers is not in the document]
jdolan: Good observation, but not all stage transforms make sense for sky stages.

src/client/renderer/r_sky.c
```diff
@@ -47,10 +47,102 @@ static struct {
...
+  }
+
+  glBlendFunc(stage->cm->blend.src, stage->cm->blend.dest);
+
+  const r_image_t *image = (r_image_t *) stage->media;
+  assert(image);
+
+  glBindTexture(GL_TEXTURE_2D, image->texnum);
+
+  glDrawElements(GL_TRIANGLES, draw->num_elements, GL_UNSIGNED_INT, draw->elements);
```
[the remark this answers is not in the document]
jdolan: Fixed in 592f30941392828682a9c8e64a39d0585e8895a1

PR #811, opened 2026-05-29
Feature/custom fullscreen resolution

Eclipse1982:
## Summary

Adds a **Resolution** picker to **Settings → Display** and makes **Exclusive Fullscreen** (`r_fullscreen 2`) switch the monitor to the chosen resolution. Previously both fullscreen modes always used the desktop resolution, with no way to pick a resolution from the UI. Windowed and Borderless behavior is unchanged.

## Changes

- Add archived cvars `r_display_width` / `r_display_height` (default `0/0` = use desktop resolution).
- In `R_InitContext`, when in exclusive fullscreen, match the requested resolution to the nearest supported mode (highest refresh rate) via `SDL_GetClosestFullscreenDisplayMode` and apply it with `SDL_SetWindowFullscreenMode`. `0/0`, or no matching mode, falls back to the desktop mode (current behavior).
- Add a **Resolution** dropdown to `Settings → Display`, populated from `SDL_GetFullscreenDisplayModes` (deduped to unique WxH). The selection is written to the cvars and applied by the existing **Apply** button via `r_restart`.

## Testing

- [x] Builds without warnings — built the two affected projects via the `Quetoo.vs15` solution (clang-cl, x64 Release): `quetoo.vcxproj` and `cgame.vcxproj`, 0 warnings / 0 errors. (Did not run the autotools `make -j$(nproc)` build.)
- [ ] Tests pass (`make check`) — not run (built via the Windows VS solution).
- [x] Tested in-game on **Windows 11 (x64, clang-cl build)** — selected several non-native modes from the dropdown (`2560x1440`, `3840x1080`, `4096x2160`) and back to `1920x1080`/Desktop; each switched the display correctly (confirmed by the new `Exclusive fullscreen WxH@Hz` console line), persisted across restart (archived cvars), and Window/Borderless behavior was unaffected.

## Notes

- Scope is intentionally limited to exclusive fullscreen. The dropdown's selection persists in Window/Borderless modes but only takes effect in exclusive fullscreen (Borderless is necessarily desktop-resolution).
- Refresh rate is chosen automatically (highest available for the selected WxH) by passing `0` to `SDL_GetClosestFullscreenDisplayMode`, keeping the dropdown to a simple unique-resolution list.


jdolan:
We had this for 18 years or so, the cvars were `r_width` and `r_height` But, imo, the current setup is far better. 

- The game never has to change your display mode, which is less jarring when the game starts / terminates.
- `r_framebuffer_scale` allows you to run at lower _or higher_ (supersampling) resolutions
- `r_draw_scale` allows you to independently scale the UI elements up or down
- Neither of those cvars require a GL context restart; you can play with the sliders and see the results that frame.

Can I ask what the motivation is for this?

Eclipse1982:
If I dont want to play in 5120x1440 but dont wanna goto my display settings change my dektop resolution just to play a game that i dont want super wide i hat to runu fov @ 160 i would like the ability to choose my resolution i play at.

jdolan:
Ah, but.. it doesn't do "letterbox" right? It's not like you get a 16:9 viewport in the center with black bars on the sides with this change, correct? Or is that what your video driver actually does? That would be cool... Because I'm picturing a  2560x1440 window stretched out to double its natural width, and that would look like absolute garbage, no? 😆

Eclipse1982:
yeah so it's not a stretched window, it's an actual fullscreen mode switch. it calls
SDL_GetClosestFullscreenDisplayMode and SDL_SetWindowFullscreenMode so the display really flips to 2560x1440, and
the framebuffer/viewport get sized off the actual mode so it renders native at 1440p, nothing stretched in software.

 the letterbox vs stretch thing is just whatever your gpu/monitor scaler is set to, same as any old game that changed
  resolutions. on my 5120x1440 with nvidia aspect ratio scaling i pick 2560x1440 and get a clean centered 16:9 with
 black bars on the sides. looks great, not garbage lol.
  
 and yeah i know about framebuffer_scale and draw_scale but those scale evenly so on an ultrawide i'm still stuck at
 32:9 either way, still fov 160 or a stretched view. the whole point here is changing the aspect ratio by picking a
 narrower native mode without having to drop my windows desktop to 1440p first every time. it only does anything at
 r_fullscreen 2, default is 0/0 which just uses the desktop mode so nobody else is affected.

jdolan:
Closing out as I merged this in https://github.com/jdolan/quetoo/pull/814

PR #815, opened 2026-05-30
Add emissive stage property to material system

skies912:
Adds an `emissive <float>` keyword to material stage blocks. The value scales the stage's own texture color and adds it to the fragment output as a constant, unlit term — independent of scene lighting. Values above 1.0 feed naturally into the existing HDR bloom pass. (summary by Claude)

This replaces the old `bloom` property to be used only in stages. Use case: Lights with glow textures or computers with glowing bits, etc.


jdolan:
This is _basically_ a way to do `_glow` textures without a separate stage pass, yea?

skies912:
it doesn't replace the glow stage - just adds HDR uniformly no matter the external light values. Glow in the dark bits will actually glow instead of a simple full-bright flat look. Can probably mimic the old bloom by using `material` surf flag so it applies it to the diffuse.

PR #826, opened 2026-06-06
Server cvars for player movement parameters (pm_params_t)

### LINKED ISSUE #589: Player movement parameters should be sent over the wire
As a server admin, I'd like to run a Quake themed server that features Quake level remakes and QuakeWorld style player movement. To do this, I'd like to be able to specify all player movement parameters (friction, acceleration, air acceleration, etc) via console variables. The server should send these to the client as a custom packet type if and when they change, so that client side prediction will match the server.

Eclipse1982:
## Summary

Makes Quetoo's player-movement tuning server-controllable. Acceleration, friction and speed for ground / air / water / spectator / ladder, plus gravity and the water-gravity modifier, move from compile-time constants in `bg_pmove.h` into a networked `pm_params_t` struct.

Because `Pm_Move` runs in client-side prediction (`cg_predict.c`) as well as on the authoritative server, a server-only change to these shared constants would make stock clients mispredict (rubber-banding). So the params are carried per-player in `pm_state_t` — the same channel the client already predicts from — exactly the way `gravity` already travels. Every value defaults to the constant it replaces, so out-of-the-box behavior is unchanged until an admin sets one.

Fixes #589

## Changes

- **`src/shared/shared.h`** — new `pm_params_t` struct (gravity + `gravity_water` + accel/friction/speed for ground, ground-slick, air, water, spectator, ladder), embedded in `pm_state_t` as `params`.
- **`src/game/default/g_main.{c,h}`** — 23 movement cvars in their own alphabetized block (`g_air_acceleration`, `g_water_friction`, `g_jump_speed`, …), deliberately *not* `CVAR_SERVER_INFO` to keep the server-info string under its 1024-byte limit. A `G_MovementParams()` factory hydrates a fresh `pm_params_t` from the cvars — no singleton cache, no dirty-checking.
- **`src/game/default/g_client.c`** — assigns `pm.s.params = G_MovementParams()` at the `Pm_Move` call site, so a mod can override fields per player there. Gravity keeps its existing precedence (maps.lst > worldspawn key > `g_gravity` > default) via `g_level.gravity`, which the factory reads.
- **`src/game/default/bg_pmove.c`** — `Pm_Move` reads `pm->s.params.*` and sanitizes every parameter at its point of use: divisors kept positive, rates and impulses floored at zero. The `PM_*` constants remain as the default seed values.
- **`src/net/net_message.{h,c}`** — the `pm_state` delta bitmask widens `uint16_t` → `uint32_t`; the float params share one new delta flag `PS_PM_PARAMS`, written only when they change (≈zero cost under delta compression). Gravity keeps its existing `PS_PM_GRAVITY` int16 path.
- **`src/game/default/g_types.h`** — `PROTOCOL_MINOR` bumped to 1043 (wire format changed).
- **`src/tests/check_net_message.c`** (new) — round-trips a fully-populated `pm_params_t` through `Net_WriteDeltaPlayerState` / `Net_ReadDeltaPlayerState`, and asserts the params block is skipped when unchanged.

## Testing

- [x] Builds without warnings — client, dedicated, `game.so` and `cgame.so` build on Debian 12 (deps per `linux/docker/Dockerfile.build`); no warnings on the changed files.
- [ ] Tests pass (`make check`) — the new `check_net_message` passes (`100%: Checks: 2, Failures: 0`). Full `make check` remains blocked by the pre-existing `check_master` link failure on `main` (`undefined reference to Net_HttpPostAsync`), unrelated to this change.
- [ ] Tested in-game — not yet; needs a client built from this branch, since the `PROTOCOL_MINOR` bump (correctly) rejects stock clients. Plan is to exercise non-default values on a couple of rotation maps alongside the physics-enum follow-on.

## Notes

- **Follow-on (per review):** a `physics`/`movement` worldspawn + maps.lst enum (`quake` / `quake2` / `default`) alongside the existing `items` key, with a single `CVAR_SERVER_INFO` cvar exposing the active model; the `gravity` worldspawn key retires then.
- **Scope.** This exposes the "feel" surface (accel / friction / speed per movement state + gravity). Engine internals (stop-epsilon, clip-bounce, snap distance, step-normal, fall/land thresholds) stay compile-time.
- Field / cvar names follow the surrounding `snake_case` convention (`pm_state.view_offset`, `g_gravity`) rather than the style-guide table's `camelCase`, for local consistency.

### DIFF
--- a/src/game/default/bg_pmove.c
+++ b/src/game/default/bg_pmove.c
@@ -369,25 +369,27 @@ static void Pm_Friction(const bool flying) {
     return;
   }
 
-  const float control = Maxf(PM_SPEED_STOP, speed);
+  const float control = Maxf(pm->s.params.speed_stop, speed);
 
   float friction = 0.f;
 
   if (pm->s.type == PM_SPECTATOR) { // spectator friction
-    friction = PM_FRICT_SPECTATOR;
+    friction = pm->s.params.friction_spectator;
   } else if (pm->s.flags & PMF_ON_LADDER) { // ladder friction
-    friction = PM_FRICT_LADDER;
+    friction = pm->s.params.friction_ladder;
   } else if (pm->water_level > WATER_FEET) { // water friction
-    friction = PM_FRICT_WATER;
+    friction = pm->s.params.friction_water;
   } else if (pm->s.flags & PMF_ON_GROUND) { // ground friction
     if (pm_locals.ground.ent && (pm_locals.ground.surface & SURF_SLICK)) {
-      friction = PM_FRICT_GROUND_SLICK;
+      friction = pm->s.params.friction_ground_slick;
     } else {
-      friction = PM_FRICT_GROUND;
+      friction = pm->s.params.friction_ground;
     }
   } else { // everything else friction
-    friction = PM_FRICT_AIR;
+    friction = pm->s.params.friction_air;
   }
+
+  friction = Maxf(0.f, friction); // never reverse direction
 
   // scale the velocity, taking care to not reverse direction
   const float scale = Maxf(0.f, speed - (friction * control * pm_locals.time)) / speed;
@@ -424,10 +426,10 @@ static void Pm_Gravity(void) {
     return;
   }
 
-  float gravity = pm->s.gravity;
+  float gravity = pm->s.params.gravity;
 
   if (pm->water_level > WATER_WAIST) {
-    gravity *= PM_GRAVITY_WATER;
+    gravity *= pm->s.params.gravity_water;
   }
 
   pm->s.velocity.z -= gravity * pm_locals.time;
@@ -781,12 +783,13 @@ static void Pm_CheckDuck(void) {
     }
 
     const float height = Box3_Size(pm->bounds).z;
+    const float duck_stand_speed = Maxf(0.f, pm->s.params.speed_duck_stand); // never reverse the transition
 
     if (pm->s.flags & PMF_DUCKED) { // ducked, reduce height
       const float target = pm->bounds.mins.z + height * 0.5f;
 
       if (pm->s.view_offset.z > target) { // go down
-        pm->s.view_offset.z -= pm_locals.time * PM_SPEED_DUCK_STAND;
+        pm->s.view_offset.z -= pm_locals.time * duck_stand_speed;
       }
 
       if (pm->s.view_offset.z < target) {
@@ -799,7 +802,7 @@ static void Pm_CheckDuck(void) {
       const float target = pm->bounds.mins.z + height * 0.9f;
 
       if (pm->s.view_offset.z < target) { // go up
-        pm->s.view_offset.z += pm_locals.time * PM_SPEED_DUCK_STAND;
+        pm->s.view_offset.z += pm_locals.time * duck_stand_speed;
       }
 
       if (pm->s.view_offset.z > target) {
@@ -838,7 +841,7 @@ static bool Pm_CheckJump(void) {
   }
 
   // finally, do the jump
-  float jump = PM_SPEED_JUMP;
+  float jump = Maxf(0.f, pm->s.params.speed_jump);
 
   // factoring in water level
   if (pm->water_level > WATER_FEET) {
@@ -951,7 +954,7 @@ static bool Pm_CheckWaterJump(void) {
     }
 
     // jump out of water
-    pm->s.velocity.z = PM_SPEED_WATER_JUMP;
+    pm->s.velocity.z = Maxf(0.f, pm->s.params.speed_water_jump);
 
     pm->s.flags |= PMF_TIME_WATER_JUMP | PMF_JUMP_HELD;
     pm->s.time = 2000;
@@ -973,12 +976,15 @@ static void Pm_LadderMove(void) {
 
   Pm_Currents();
 
+  const float ladder_speed = Maxf(0.f, pm->s.params.speed_ladder);
+  const float ladder_accel = Maxf(0.f, pm->s.params.accel_ladder);
+
   // user intentions in X/Y
   vec3_t vel = Vec3_Zero();
   vel = Vec3_Fmaf(vel, pm->cmd.forward, pm_locals.forward_xy);
   vel = Vec3_Fmaf(vel, pm->cmd.right, pm_locals.right_xy);
 
-  const float s = PM_SPEED_LADDER * 0.125f;
+  const float s = ladder_speed * 0.125f;
 
   // limit horizontal speed when on a ladder
   vel.x = Clampf(vel.x, -s, s);
@@ -986,16 +992,16 @@ static void Pm_LadderMove(void) {
   vel.z = 0.f;
 
   // handle Z intentions differently
-  if (fabsf(pm->s.velocity.z) < PM_SPEED_LADDER) {
+  if (fabsf(pm->s.velocity.z) < ladder_speed) {
 
     if ((pm->angles.x <= -15.f) && (pm->cmd.forward > 0)) {
-      vel.z = PM_SPEED_LADDER;
+      vel.z = ladder_speed;
     } else if ((pm->angles.x >= 15.f) && (pm->cmd.forward > 0)) {
-      vel.z = -PM_SPEED_LADDER;
+      vel.z = -ladder_speed;
     } else if (pm->cmd.up > 0) {
-      vel.z = PM_SPEED_LADDER;
+      vel.z = ladder_speed;
     } else if (pm->cmd.up < 0) {
-      vel.z = -PM_SPEED_LADDER;
+      vel.z = -ladder_speed;
     } else {
       vel.z = 0.f;
     }
@@ -1007,13 +1013,13 @@ static void Pm_LadderMove(void) {
 
   float speed;
   const vec3_t dir = Vec3_NormalizeLength(vel, &speed);
-  speed = Clampf(speed, 0.f, PM_SPEED_LADDER);
+  speed = Clampf(speed, 0.f, ladder_speed);
 
   if (speed < PM_STOP_EPSILON) {
     speed = 0.f;
   }
 
-  Pm_Accelerate(dir, speed, PM_ACCEL_LADDER);
+  Pm_Accelerate(dir, speed, ladder_accel);
 
   Pm_StepSlideMove();
 }
@@ -1034,7 +1040,7 @@ static void Pm_WaterJumpMove(void) {
 
   // if we've reached a usable spot, clamp the jump to avoid launching
   if (Pm_Trace(pm->s.origin, pos, pm->bounds).fraction == 1.f) {
-    pm->s.velocity.z = Clampf(pm->s.velocity.z, 0.f, PM_SPEED_JUMP);
+    pm->s.velocity.z = Clampf(pm->s.velocity.z, 0.f, Maxf(0.f, pm->s.params.speed_jump));
   }
 
   // if we're falling back down, clear the timer to regain control
@@ -1058,10 +1064,12 @@ static void Pm_WaterMove(void) {
 
   Pm_Debug("%s\n", vtos(pm->s.origin));
 
+  const float water_speed = Maxf(1.f, pm->s.params.speed_water); // also a loop divisor below
+
   // apply friction, slowing rapidly when first entering the water
   float speed = Vec3_Length(pm->s.velocity);
 
-  for (int32_t i = speed / PM_SPEED_WATER; i >= 0; i--) {
+  for (int32_t i = speed / water_speed; i >= 0; i--) {
     Pm_Friction(true);
   }
 
@@ -1096,13 +1104,13 @@ static void Pm_WaterMove(void) {
   }
 
   const vec3_t dir = Vec3_NormalizeLength(vel, &speed);
-  speed = Clampf(speed, 0, PM_SPEED_WATER);
+  speed = Clampf(speed, 0, water_speed);
 
   if (speed < PM_STOP_EPSILON) {
     speed = 0.f;
   }
 
-  Pm_Accelerate(dir, speed, PM_ACCEL_WATER);
+  Pm_Accelerate(dir, speed, Maxf(0.f, pm->s.params.accel_water));
 
   if (pm->cmd.up > 0) {
     Pm_SlideMove();
@@ -1127,7 +1135,7 @@ static void Pm_AirMove(void) {
   vel = Vec3_Fmaf(vel, pm->cmd.right, pm_locals.right_xy);
   vel.z = 0.f;
 
-  float max_speed = PM_SPEED_AIR;
+  float max_speed = Maxf(1.f, pm->s.params.speed_air); // air_speed must stay positive to bound the wish-speed
 
   // accounting for walk modulus
   if (pm->cmd.buttons & BUTTON_WALK) {
@@ -1142,7 +1150,7 @@ static void Pm_AirMove(void) {
     speed = 0.f;
   }
 
-  float accel = PM_ACCEL_AIR;
+  float accel = Maxf(0.f, pm->s.params.accel_air);
 
   if (pm->s.flags & PMF_DUCKED) {
     accel *= PM_ACCEL_AIR_MOD_DUCKED;
@@ -1194,12 +1202,14 @@ static void Pm_WalkMove(void) {
 
   // clamp to max speed
   if (pm->water_level > WATER_FEET) {
-    max_speed = PM_SPEED_WATER;
+    max_speed = pm->s.params.speed_water;
   } else if (pm->s.flags & PMF_DUCKED) {
-    max_speed = PM_SPEED_DUCKED;
+    max_speed = pm->s.params.speed_ducked;
   } else {
-    max_speed = PM_SPEED_RUN;
+    max_speed = pm->s.params.speed_ground;
   }
+
+  max_speed = Maxf(0.f, max_speed); // keep the Clampf range valid
 
   // accounting for walk modulus
   if (pm->cmd.buttons & BUTTON_WALK) {
@@ -1216,7 +1226,8 @@ static void Pm_WalkMove(void) {
   }
 
   // accelerate based on slickness of ground surface
-  const float accel = (pm_locals.ground.surface & SURF_SLICK) ? PM_ACCEL_GROUND_SLICK : PM_ACCEL_GROUND;
+  const float accel = Maxf(0.f, (pm_locals.ground.surface & SURF_SLICK)
+      ? pm->s.params.accel_ground_slick : pm->s.params.accel_ground);
 
   Pm_Accelerate(dir, speed, accel);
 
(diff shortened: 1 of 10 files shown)


jdolan (commented):
Thanks! This is a solid start towards one of our oldest tickets:
https://github.com/jdolan/quetoo/issues/589

Some feedback:

- Why stop at air movement parameters? Shouldn't we include ground accel, friction and speed? Water too?

- Sending these over in pm_state_t is interesting. I hadn't considered that, but that would allow for each player to have different physics -- something pretty cool for a class-based mod. My original idea was to send these parameters over as ConfigStrings, so they'd be global / shared.

- g_level_t is probably not the right home for these. I'd introduce a new struct entirely for pm_params_t or something like that. Move gravity and all of these other variables into it. Pass it in as a parameter to Bg_Pmove().

Love this, tho! Keep it coming!

Eclipse1982:
Thanks, this was really helpful feedback — ended up reworking the whole thing and it's in a much better spot now. It actually lines up with #589 now instead of just doing the air case.

So to your points: I pulled everything into a `pm_params_t` struct like you suggested, and yeah, gravity moved in there too — `g_level_t` doesn't carry any of it anymore. There's one `g_pm_params` on the game side that gets seeded at worldspawn and refreshed when a cvar changes, and `Pm_Move` just reads `pm->s.params.*`. I also took your "why stop at air?" to heart and did the full set: accel/friction/speed for ground, air, water, spectator and ladder, plus gravity and the water-gravity mod. That's 23 cvars, all defaulting to the existing `PM_*` values so nothing changes until someone actually sets one. I left the truly internal stuff alone (stop epsilon, clip bounce, snap distance, etc.) since those aren't really "feel" knobs — but if you want #589's "all parameters" taken literally I can keep going.

The one spot where I went off-script is transport, and I want your read on it. #589 (and Paril's Q2pro `pmp` note) describe a global custom packet. I ended up putting the params in the player-state delta instead, behind a new `PS_PM_PARAMS` flag that only writes when something actually changes. Gravity kept its old `PS_PM_GRAVITY` int16 path, and I widened the `pm_state` delta mask from 16 to 32 bits to make room (we were basically out of flags anyway). The reason I leaned this way is it gets you the per-player physics thing you mentioned almost for free — a global server just sets everyone the same value. But if you'd rather have the literal #589 version (global configstring / dedicated packet), or global-by-default with per-player as an override, say the word — honestly that'd make the wire code simpler, so I'm not attached.

Bumped `PROTOCOL_MINOR` to 1043 since the format changed.

Testing-wise: builds clean (client, dedicated, game.so, no new warnings), and I added a `check_net_message` test that round-trips a full `pm_params_t` through the delta and checks it gets skipped when nothing changed — both pass. Haven't tried it in-game with non-default values yet, since that needs a client off this branch thanks to the protocol bump. Happy to post results once I do, or if you want to eyeball the prediction path yourself that'd be welcome too.

Main thing I need from you: keep the per-player transport or switch to global? And is the cvar set about right, or do you want the rest of the knobs? Once that's nailed down I'll tag it `Fixes #589`.

src/game/default/bg_pmove.c
```diff
@@ -1142,7 +1153,11 @@ static void Pm_AirMove(void) {
     speed = 0.f;
   }
 
-  float accel = PM_ACCEL_AIR;
+  float accel = pm->s.params.accel_air;
+
+  if (accel < 0.f) {
+    accel = 0.f;
+  }
```
jdolan:
```suggestion
  float accel = Maxf(0.f, pm->s.params.accel_air);
```

src/game/default/bg_pmove.c
```diff
@@ -1127,7 +1134,11 @@ static void Pm_AirMove(void) {
   vel = Vec3_Fmaf(vel, pm->cmd.right, pm_locals.right_xy);
   vel.z = 0.f;
 
-  float max_speed = PM_SPEED_AIR;
+  float max_speed = pm->s.params.speed_air;
+
+  if (max_speed < 1.f) {
+    max_speed = 1.f; // air_speed must stay positive to bound the wish-speed
+  }
```
jdolan:
```suggestion
  float max_speed = Maxf(1.f, pm->s.params.speed_air);
```

src/game/default/bg_pmove.c
```diff
@@ -1058,10 +1063,12 @@ static void Pm_WaterMove(void) {
 
   Pm_Debug("%s\n", vtos(pm->s.origin));
 
+  const float water_speed = Maxf(1.f, pm->s.params.speed_water); // also a loop divisor below
```
jdolan, line 1067: Yup, good catch. My general feedback on this entire PR would be:

Now that these are all user-defined, we need to take extra care to avoid SIGFPE because some goofball wants to set these to zero 😄

src/game/default/g_ai_node.c
```diff
@@ -1336,7 +1336,7 @@ GArray *G_Ai_Node_FindPath(const g_client_t *cl, const ai_node_id_t start, const
 
       if (cl && cl->entity) {
         const int32_t water_level = (link_contents & CONTENTS_WATER) ? 1 : 0;
-        const float estimated_damage = G_Ai_EstimatedFallDamage(drop, g_level.gravity, water_level);
+        const float estimated_damage = G_Ai_EstimatedFallDamage(drop, g_pm_params.gravity, water_level);
```
jdolan: Got it. So there's a singleton `pm_params_t` that all players will receive a "copy" of, but we could override them at the call site with player class-specific params. That's perfect.

src/game/default/g_main.c
```diff
@@ -147,6 +149,29 @@ cvar_t *g_frag_limit;
 cvar_t *g_friendly_fire;
 cvar_t *g_gameplay;
 cvar_t *g_gravity;
+cvar_t *g_water_gravity;
```
jdolan: Nit: let's break all of the pmove param cvars out into their own block, alphabetized.

src/game/default/g_main.c
```diff
@@ -556,6 +581,39 @@ static char *G_FormatTime(uint32_t time) {
...
 }
 
+/**
+ * @brief Refreshes the movement-parameter cache from the g_* cvars. Gravity is
+ * managed separately (it has map/worldspawn precedence), so it is not set here.
+ */
+void G_UpdatePmParams(void) {
+  g_pm_params.water_gravity = g_water_gravity->value;
+
+  g_pm_params.accel_ground = g_ground_acceleration->value;
```
jdolan: My initial instinct would be to apply clamping here, but since we might override these at the call site for each player, it probably makes sense to keep it like you have it, where bg_pmove.c itself performs all sanitization of these values.

src/game/default/g_main.c
```diff
@@ -680,7 +738,34 @@ static void G_CheckRules(void) {
   if (g_gravity->modified) { // set gravity, G_ClientMove will read it
     g_gravity->modified = false;
 
-    g_level.gravity = g_gravity->integer;
+    g_pm_params.gravity = g_gravity->integer;
+  }
+
+  if (g_water_gravity->modified ||
```
jdolan: You know, since this struct is absolutely trivial to hydrate from these cvars, I actually don't think any of this dirty checking and updating is necessary at all. What I would probably do instead is get rid of the singleton, and instead, at the `Pm_Move` call site, just grab a fresh stack-allocated instance of the params from a factory function:

```
/**
 * @brief Factory function for `pm_params_t`.
 */
pm_params_t G_MovementParams(void) {
  return (pm_params_t) {
    .ground_acceleration = g_ground_acceleration->value,
    .air_acceleration = g_air_acceleration->value,
    // ...
  };
}
```

```
  // G_ClientThink

  pm.s.params = G_MovementParams();
  
  Pm_Move(&pm);

  // ...
}
```

src/game/default/g_main.h
```diff
@@ -150,6 +154,29 @@ extern cvar_t *g_frag_limit;
 extern cvar_t *g_friendly_fire;
 extern cvar_t *g_gameplay;
 extern cvar_t *g_gravity;
+extern cvar_t *g_water_gravity;
```
jdolan: Same comment about just breaking these vars apart into their own block.

src/game/default/g_physics.c
```diff
@@ -306,7 +306,7 @@ static void G_Accelerate(g_entity_t *ent, const vec3_t dir, float speed, float a
 static void G_Gravity(g_entity_t *ent) {
 
   if (ent->ground.ent == NULL) {
-    float gravity = g_level.gravity;
+    float gravity = g_pm_params.gravity;
```
jdolan: I wonder if this is the right time to rethink gravity.

We have _never_ had a single map that uses this worldspawn key to mess with gravity. It's dumb.

Doubling down on the pattern with a worldspawn / maps.lst key for _every_ pmove parameter would also be gross.

What if, instead, we offered an enum of physics models via worldspawn key / maps.lst, just like we [currently do for items](https://github.com/jdolan/quetoo/blob/main/src/game/default/g_entity.c#L837)?

So a worldspawn enum-style key called `movement` or `physics` with values `quake` `quake2` `default` for example, allowing mappers to turn on Quake-inspired physics for our Quake remakes, that goes along with the existing `items` key to enable Quake weapons, ammo, armor, etc. Could even turn on `quake2` physics on the Quake2 map remakes for ultra purists (talking about you @Panjoo).

src/game/default/g_types.h
```diff
@@ -834,11 +834,6 @@ typedef struct {
    */
   char name[MAX_QPATH];
 
-  /**
-   * @brief Gravity override.
-   */
-  int16_t gravity;
```
jdolan, line 759: 👍

src/net/net_message.c
```diff
@@ -361,6 +366,32 @@ void Net_WriteDeltaPlayerState(mem_buf_t *msg, const player_state_t *from, const
...
+    Net_WriteFloat(msg, to->pm_state.params.speed_air);
+    Net_WriteFloat(msg, to->pm_state.params.speed_water);
+    Net_WriteFloat(msg, to->pm_state.params.speed_ladder);
+    Net_WriteFloat(msg, to->pm_state.params.speed_spectator);
+    Net_WriteFloat(msg, to->pm_state.params.speed_stop);
+    Net_WriteFloat(msg, to->pm_state.params.speed_jump);
+    Net_WriteFloat(msg, to->pm_state.params.speed_ducked);
+    Net_WriteFloat(msg, to->pm_state.params.speed_duck_stand);
+    Net_WriteFloat(msg, to->pm_state.params.speed_water_jump);
+  }
```
jdolan, line 393: These change so infrequently, and likely together, that I think sending them all at once, rather than delta-per-field, is actually the right call 👍

src/shared/shared.h
```diff
@@ -277,6 +277,25 @@ typedef enum {
...
 #define PMF_GAME (1 << 0)
 
+/**
+ * @brief Server-tunable player-movement parameters, networked per-player
+ * inside `pm_state_t` so that client-side prediction matches the server.
+ * Each field defaults to the `PM_*` constant it replaces (see bg_pmove.h).
+ */
+typedef struct {
+  int16_t gravity;     // world gravity; default from g_gravity / map (int16)
+  float water_gravity; // PM_GRAVITY_WATER
```
jdolan: `gravity_water` would follow the naming convention for all of the other variables.

jdolan (commented):
Love it! I think this is quite close now. I left some more feedback.

The worldspawn / maps.lst "physics" enum can be a follow-on PR as well, so that we can merge this and start playing around with it.

src/game/default/g_main.c
Eclipse1982: Done — dropped `g_pm_params` and the per-cvar dirty-checking entirely; `G_MovementParams()` is grabbed fresh at the `Pm_Move` site in `G_ClientMove`, so a class-based mod can override fields there per-player.

One wrinkle worth flagging: resolved gravity carries maps.lst/worldspawn precedence, so it can't be hydrated from a cvar statelessly. I put it back on `g_level.gravity` (seeded at worldspawn, refreshed when `g_gravity` changes) and the factory reads it. That walks back your round-1 "not `g_level`" note for gravity specifically — happy to move it elsewhere, but it lines up with the gravity rethink below, so this felt like the least-bad home for now.

src/game/default/bg_pmove.c
Eclipse1982: Audited the whole pmove path with that in mind. The two air-move spots were the only gaps — fixed via your suggestions. The one parameter used as a divisor (`speed / water_speed`) is already `Maxf(1.f, …)`, and `Pm_Friction`'s `/ speed` is covered by the early `speed < 1.f` return. Everything else (ground/water/ladder/spectator accel, friction, max speeds, jump) routes through `Maxf`/`Clampf` before use.

src/game/default/g_physics.c
Eclipse1982: Agreed — a `physics`/`movement` worldspawn + maps.lst enum (quake / quake2 / default) alongside the existing `items` key is a much cleaner model than a per-parameter key, and the gravity worldspawn key has earned its retirement. I'll pick this up as the follow-on you mentioned; left gravity on the existing path here so this PR stays mergeable.

src/game/default/g_ai_node.c
Eclipse1982: Right — heads up that with the singleton gone it's now a fresh per-move `G_MovementParams()` value, so the per-class override would happen at the `Pm_Move` call site rather than on a shared struct.

src/game/default/g_main.c
Eclipse1982: Done — pulled into their own alphabetized block (same in `g_main.h`).

src/game/default/g_main.h
Eclipse1982: Done — own alphabetized block here too.

src/shared/shared.h
Eclipse1982: Renamed to `gravity_water`.

src/game/default/bg_pmove.c
Eclipse1982: Applied — it's in the pushed commit.

src/game/default/bg_pmove.c
Eclipse1982: Applied — it's in the pushed commit.

src/game/default/g_physics.c
jdolan: Perfect!

src/game/default/g_main.c
jdolan: Yup, that's the right call. We'll clean it up when we introduce the "physics" worldspawn key and yank gravity out of there.

jdolan:
This is awesome. Thank you for this. Do you think it's ready to merge in its current state, or were there any tweaks left you wanted to make first?

src/game/default/g_main.c
```diff
@@ -1060,6 +1123,29 @@ void G_Init(void) {
   g_friendly_fire = gi.AddCvar("g_friendly_fire", "1", CVAR_SERVER_INFO, "Factor of how much damage can be dealt to teammates.");
   g_gameplay = gi.AddCvar("g_gameplay", "default", CVAR_SERVER_INFO, "Selects deathmatch, instagib or arena combat.");
   g_gravity = gi.AddCvar("g_gravity", "800", CVAR_SERVER_INFO, NULL);
+  g_water_gravity = gi.AddCvar("g_water_gravity", "0.33", CVAR_SERVER_INFO, "Fraction of gravity applied underwater. Default 0.33.");
+  g_ground_acceleration = gi.AddCvar("g_ground_acceleration", "10.0", CVAR_SERVER_INFO, "Ground acceleration. Default 10.0.");
+  g_ground_acceleration_slick = gi.AddCvar("g_ground_acceleration_slick", "4.375", CVAR_SERVER_INFO, "Ground acceleration on slick surfaces. Default 4.375.");
+  g_air_acceleration = gi.AddCvar("g_air_acceleration", "2.0", CVAR_SERVER_INFO, "Acceleration applied while airborne. Default 2.0; set 0 for classic-Quake2 movement.");
```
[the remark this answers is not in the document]
jdolan: Oof! Yes, these should not be CVAR_SERVER_INFO -- great catch, Mr Copilot. Maybe we'll introduce a cvar for the worldspawn / enum based approach, and expose just that one as CVAR_SERVER_INFO.

Eclipse1982:
Thanks for the thorough pass! Before I push anything else, I'd like to line up with you on the two open threads so I take this where you want it:

**1. `CVAR_SERVER_INFO` on the movement cvars** — agreed these shouldn't be server-info. I'd drop the flag to `0` on all 23, matching the existing `g_balance_*` tuning cvars:
```c
// now
g_water_jump_speed = gi.AddCvar("g_water_jump_speed", "420.0", CVAR_SERVER_INFO, "...");
// proposed
g_water_jump_speed = gi.AddCvar("g_water_jump_speed", "420.0", 0, "...");
```
…and reserve server-info for the future `physics`/worldspawn enum cvar you mentioned. Do that flag change here, or fold it into the worldspawn-enum follow-on?

**2. Negative param values** (Copilot's flags on `g_duck_stand_speed` / `g_water_jump_speed`). Concretely, `g_water_jump_speed -420` would fire the water-jump *downward* while still setting the jump timer; `g_duck_stand_speed -200` moves the view away from its target so the duck/stand transition never converges. The PR already clamps *some* params at their `bg_pmove.c` use-sites but not these, so it's inconsistent. Options:

**a) Centralize** — clamp every rate once in `G_MovementParams()` and drop the scattered guards in pmove:
```c
// g_main.c — single authoritative chokepoint, before params hit the wire
.speed_water_jump = Maxf(0.f, g_water_jump_speed->value),
.speed_duck_stand = Maxf(0.f, g_duck_stand_speed->value),
// …and so on for every rate; bg_pmove.c then reads pm->s.params.* verbatim
```

**b) Per-site** — leave the existing use-site clamps and just add the two missing:
```c
// bg_pmove.c
pm->s.velocity.z = Maxf(0.f, pm->s.params.speed_water_jump);
```

**c) Leave it** — these are admin-set, garbage-in tolerated.

Leaning (a) since it's one place and the client gets already-sane values. What's your preference?

Either way I'll roll it up cleanly. Thanks!

jdolan:
I think my preference is to clamp them in bg_pmove.c, which runs on both client and server. That keeps the clamping closer to the logic that relies on it, which can help readability. Knowing whether a variable is a divisor or a multiplier just by glancing at the next line 👍  So if we just add the two missing clamps, then we can squash this puppy and move into setting up the enum and actually trying this all out on a couple maps in the rotation.

Eclipse1982:
Done — both clamps live in `bg_pmove.c` now. `speed_water_jump` is floored at zero right where it becomes the vertical impulse, and `speed_duck_stand` clamps once into a local that both the duck and stand transitions share, so the view offset can't walk away from its target.

While I was in there I also dropped `CVAR_SERVER_INFO` → `0` on the 23 new cvars per the Copilot thread (`g_gravity` keeps it, since it predates this PR), and alphabetized the `g_main.c` block to match `g_main.h`.

Squashed to a single commit and force-pushed. Rebuilt clean — client, dedicated, `game.so`, `cgame.so`, no warnings on the changed files — and `check_net_message` still passes 2/2 (the only `make check` casualty remains the pre-existing `check_master` link issue on `main`). Ready for the physics-enum follow-on whenever you are.

jdolan:
🤘🏻😎🤘🏻

PR #831, opened 2026-06-11
Add Valve-220 map format support to quemap.

skies912:
Auto-detects Valve-220 by peeking for '[' after the texture name; falls back to standard Q3 format if not present. Both formats are supported transparently with no flags required.

***
Ran into issues aligning faces in `cavern` - this helps. TB support for patch editing is coming too, which will also help alot.

### DIFF
--- a/Quetoo.vs15/quemap.vcxproj
+++ b/Quetoo.vs15/quemap.vcxproj
@@ -119,6 +119,10 @@
     <Link>
       <AdditionalDependencies>SDL3.lib;SDL3_image.lib;glib-2.0.lib;intl.lib;dbghelp.lib;Wldap32.lib;physfs.lib;ws2_32.lib;dlfcn.lib;Objectively.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <SubSystem>Console</SubSystem>
+      <!-- BSP tree/portal generation recurses to tree depth; the default 1 MB Windows
+           main-thread stack overflows on large maps (e.g. cavern). Reserve 8 MB to match
+           the Linux/macOS main-thread default. -->
+      <StackReserveSize>8388608</StackReserveSize>
     </Link>
     <PostBuildEvent>
       <Command>COPY_DEPENDENCIES "$(QUETOO_HOME)" "$(Platform)" "$(Configuration)" &amp;&amp; COPY_QUEMAP "$(QUETOO_HOME)" "$(Platform)$(Configuration)"</Command>
--- a/Quetoo.vs15/quemap.vcxproj.user
+++ b/Quetoo.vs15/quemap.vcxproj.user
@@ -7,7 +7,10 @@
     <LocalDebuggerCommandArguments>-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp test</LocalDebuggerCommandArguments>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <LocalDebuggerCommandArgumentsHistory>-p "D:\quetoo_dev\github\quetoo\Quetoo.vs15\../../quetoo-data/src/default" -bsp -light tokays|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp test|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp longyard|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp maps/longyard|</LocalDebuggerCommandArgumentsHistory>
-    <LocalDebuggerCommandArguments>-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp maps/longyard</LocalDebuggerCommandArguments>
+    <LocalDebuggerCommandArgumentsHistory>-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp test|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp longyard|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp maps/longyard|-p "C:\Users\chrisg\AppData\Roaming\WickedOldGames\Quetoo\default" -bsp maps/longyard|-p "C:\Users\chrisg\AppData\Roaming\WickedOldGames\Quetoo\default" -bsp maps/cavern|</LocalDebuggerCommandArgumentsHistory>
+    <LocalDebuggerCommandArguments>-p "C:\Users\chrisg\AppData\Roaming\WickedOldGames\Quetoo\default" -bsp maps/cavern</LocalDebuggerCommandArguments>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <LocalDebuggerCommandArgumentsHistory>-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp test|</LocalDebuggerCommandArgumentsHistory>
   </PropertyGroup>
 </Project>
--- a/src/quemap/face.c
+++ b/src/quemap/face.c
@@ -385,7 +385,7 @@ bsp_face_t *EmitFace(const face_t *face) {
   return out;
 }
 
-#define MAX_VERTEX_FACES 32
+#define MAX_VERTEX_FACES 64
 
 static const bsp_model_t *phong_model;
 static float phong_cosine;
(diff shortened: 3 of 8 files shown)


jdolan:
This is great! But before we can merge it, we need to think this through.

What are the implications to the editor? Does it need to be able to write v220 .map? Does it actually Just Work because it uses raw string copies of the brushes to begin with?

Should we create a script that just converts all of our .maps to v220, if it's a superior format?

Do we need to adjust our TrenchBroom config to tell it that we support / prefer v220?

skies912:
> This is great! But before we can merge it, we need to think this through.
> 
> What are the implications to the editor? Does it need to be able to write v220 .map? Does it actually Just Work because it uses raw string copies of the brushes to begin with?
> 
> Should we create a script that just converts all of our .maps to v220, if it's a superior format?
> 
> Do we need to adjust our TrenchBroom config to tell it that we support / prefer v220?

1) I just tested it. Seems to "just work". I even deleted the .map and let the editor replace it.  One note: TB won't immediately recognize the newly written .map BUT it's cosmetic. `// Format: Quake3 (Valve)` is needed up top for 220 loading. After I changed it, map loads and compiles just fine after that.
2) It's def a better format when it's needed. We -could- use a script to convert all the maps - it didn't take much on Claude's end to make a python script for it. 
3) Yes. Though it's a simple one liner in `GameConfig.cfg`:
```
"fileformats": [
        { "format": "Quake3" },
		{ "format": "Quake3 (Valve)" }
    ],

**Comment by jdolan** [comment]
I think you've actually got this one. You should go for it. 

I would generate the script, version it in Quetoo's map-fu folder maybe? Or add it to map-fu itself? Then convert and rebake all .maps. And update the TrenchBroom and Radiant configs. I'll open a PR against TB proper for you once this is all working and merged. 

You got this! 💪🏻😎

File: Quetoo.vs15/quemap.vcxproj [hunk 1]
```diff
@@ -119,6 +119,10 @@
     <Link>
       <AdditionalDependencies>SDL3.lib;SDL3_image.lib;glib-2.0.lib;intl.lib;dbghelp.lib;Wldap32.lib;physfs.lib;ws2_32.lib;dlfcn.lib;Objectively.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <SubSystem>Console</SubSystem>
+      <!-- BSP tree/portal generation recurses to tree depth; the default 1 MB Windows
+           main-thread stack overflows on large maps (e.g. cavern). Reserve 8 MB to match
+           the Linux/macOS main-thread default. -->
+      <StackReserveSize>8388608</StackReserveSize>
```
jdolan, line 125: 🤣 Oh, fucking Windows...

Quetoo.xcodeproj/project.pbxproj
```diff
@@ -97,14 +96,12 @@
 		CE04F4B025CADFF200C31433 /* swap.c in Sources */ = {isa = PBXBuildFile; fileRef = CE12D6BA1C5C58C300CD0B13 /* swap.c */; };
 		CE04F56525CAE12B00C31433 /* libcommon.a in Frameworks */ = {isa = PBXBuildFile; fileRef = CE80FDE31C5E3E1100A21A51 /* libcommon.a */; };
 		CE04F5AA25CAE14500C31433 /* libshared.a in Frameworks */ = {isa = PBXBuildFile; fileRef = CE80FDD51C5E3D4E00A21A51 /* libshared.a */; };
-		CE04FB1925CEDCD400C31433 /* net_http.h in Headers */ = {isa = PBXBuildFile; fileRef = CE04FB1725CEDCD400C31433 /* net_http.h */; };
-		CE04FB1A25CEDCD400C31433 /* net_http.c in Sources */ = {isa = PBXBuildFile; fileRef = CE04FB1825CEDCD400C31433 /* net_http.c */; };
+		CE04FB1925CEDCD400C31433 /* net_http_server.h in Headers */ = {isa = PBXBuildFile; fileRef = CE04FB1725CEDCD400C31433 /* net_http_server.h */; };
+		CE04FB1A25CEDCD400C31433 /* net_http_server.c in Sources */ = {isa = PBXBuildFile; fileRef = CE04FB1825CEDCD400C31433 /* net_http_server.c */; };
```
jdolan, line 100: Hm. I think you may need to up-merge main or something. These changes should not be showing here 🤔 Lmk if it's not easily resolved on your side.

src/quemap/map.h
```diff
@@ -78,6 +78,11 @@ typedef struct brush_side_s {
    */
   vec2_t scale;
 
+  /**
+   * @brief True if this brush side uses Valve-220 explicit texture axes.
+   */
+  bool valve;
```
jdolan: I think this boolean probably belongs as a global, right? We'd never encounter a .map file that uses some Valve brushes and some legacy / standard brushes, right?

src/quemap/qbsp.c
```diff
@@ -168,6 +168,15 @@ int32_t BSP_Main(void) {
 
   LoadMapFile(map_name);
 
+  bool valve = false;
```
jdolan: Yea, what probably makes sense here is to introduce

```c
typedef enum {
  MAP_QUAKE2,
  MAP_VALVE_220
} map_format_t;
```

And have `LoadMapFile` return the resolved format, which you can set a global to.

```c
map_format = LoadMapFile(map_name);
```

And then everywhere downstream can just check `map_format`

src/quemap/qbsp.c
```diff
@@ -168,6 +168,15 @@ int32_t BSP_Main(void) {
...
+  for (int32_t i = 0; i < num_brush_sides; i++) {
+    if (brush_sides[i].valve) {
+      valve = true;
+      break;
+    }
+  }
+  Com_Print("Map format: %s\n", valve ? "Quake3 (Valve)" : "Quake3");
```
jdolan: Com_Verbose.

src/quemap/texture.c
```diff
@@ -68,6 +68,32 @@ static void TextureAxisForPlane(const plane_t *plane, vec3_t *xv, vec3_t *yv) {
  */
 void TextureVectorsForBrushSide(brush_side_t *side, const vec3_t origin) {
 
+  if (side->valve) {
```
jdolan:
```c
if (map_format == MAP_VALVE_220) {
```

jdolan (commented):
Almost there! We'll need a second PR in quetoo-data that updates scripts/map-fu to convert legacy to v220 as part of its preprocessing step. This will ensure that the Makefile in quetoo-data continues to function while _also_ seamlessly updating the .map files.

Quetoo.vs15/quemap.vcxproj.user
```diff
@@ -7,7 +7,10 @@
     <LocalDebuggerCommandArguments>-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp test</LocalDebuggerCommandArguments>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <LocalDebuggerCommandArgumentsHistory>-p "D:\quetoo_dev\github\quetoo\Quetoo.vs15\../../quetoo-data/src/default" -bsp -light tokays|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp test|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp longyard|-p "C:\Users\chrisg\OneDrive\Documents\My Games\Quetoo\default" -bsp maps/longyard|</LocalDebuggerCommandArgumentsHistory>
```
jdolan, line 10: I'm _guessing_ this whole file should be .gitignored? But I'm not going to lose sleep over it.

src/quemap/face.c
```diff
@@ -385,7 +385,7 @@ bsp_face_t *EmitFace(const face_t *face) {
   return out;
 }
 
-#define MAX_VERTEX_FACES 32
+#define MAX_VERTEX_FACES 64
```
jdolan, line 388: 64 faces sharing one vertex? You crazy bastard.. what're you doing? 
Alt Text

src/quemap/map.c
```diff
@@ -871,11 +882,14 @@ static entity_t *ParseEntity(parser_t *parser) {
 
 /**
  * @brief Loads and parses the .map file, populating the global entities, brushes, planes, and patches arrays.
+ * @return The resolved map format, also stored in the global `map_format`.
  */
-void LoadMapFile(const char *filename) {
+int LoadMapFile(const char *filename) {
```
jdolan: uber-nit: should return the enum type here.

jdolan (approved):
Woohoo! Some super minor feedback, feel free to address, or not. This is approved.

src/quemap/face.c
skies912: Clusters of like this produced alot of warning spam in quemap - apparently didn't like verts so close together... SHUT UP. lol
image

Quetoo.vs15/quemap.vcxproj.user
skies912: if we can just gitnore that ONE section (well ALL sections with "LocalDebugger*", that would be awesome - that seems to change each commit 👎 . The rest is relevant.

PR #832, opened 2026-06-11
Add g_fall_damage and g_hook_sky cvars

Eclipse1982:
## Summary

Adds two server-side cvars that let admins opt into classic Quake2 behaviour without maintaining an engine fork:

- **`g_fall_damage`** (default `1`) scales fall damage; `0` disables it entirely, mirroring Quake2's `DF_NO_FALLING` dmflag.
- **`g_hook_sky`** (default `0`) lets the grapple hook attach to sky surfaces instead of detaching on contact.

Both default to current behaviour, so existing servers and demos are unaffected until an admin opts in. Motivation: running an instagib / "RailWarz"-style FFA server tuned to match a Quake2 (q2pro) configuration, where fall damage is off (`dmflags 8`) and the grapple attaches to sky brushes (`hook_sky 1`). These were the only two gameplay behaviours not already reachable via the existing `g_*` cvar set (the `g_air_*` / `g_ground_*` movement params already cover the rest).

No related issue.

## Changes

- `g_main.{c,h}`: declare and register `g_fall_damage` (`CVAR_SERVER_INFO`, default `1`) and `g_hook_sky` (`CVAR_SERVER_INFO`, default `0`), following the existing `g_self_damage` registration pattern.
- `g_client.c` (`G_ClientMove`): scale the computed fall damage by `g_fall_damage->value`, and skip the `G_Damage` call when the result is `< 1`. The fall event (`EV_CLIENT_FALL` / `EV_CLIENT_FALL_FAR`) still fires, so the landing animation and view-kick are unchanged — only the health loss is gated.
- `g_ballistics.c` (`G_HookProjectile_Touch`): when `g_hook_sky` is enabled, treat a sky-surface impact as a structural attach instead of detaching.

Both cvars are read directly at the point of use like the other balance cvars, so there is no `pm_state` / protocol change.

## Testing

- [x] Builds without warnings (`make -j$(nproc)`) — verified via CI (`build-linux` + `build-windows` both green on this PR)
- [ ] Tests pass (`make check`) — not run; `check_master` has a pre-existing link failure on `main` unrelated to this change
- [ ] Tested in-game on 

The change is small and mirrors the existing cvar patterns; defaults preserve current behaviour. Happy to adjust the approach (e.g. make `g_fall_damage` a strict on/off toggle rather than a scale) if you'd prefer.

## Notes

The corresponding cvar values were validated for gameplay parity on a live instagib server.


jdolan (commented):
LGTM

PR #824, opened 2026-06-05
feat: kd-tree plus heap for a* optimizations

jdolan:
This pull request introduces significant improvements to the AI navigation node system, focusing on performance and memory efficiency. The main changes include adding a spatial acceleration structure (kd-tree) for faster nearest-node queries, replacing linear scans with efficient heap-based data structures for A* pathfinding, and optimizing memory usage. Additionally, the changes improve resource management and maintainability by centralizing platform entity tracking and cleaning up memory when nodes are deleted or the game shuts down.

**AI Navigation System Performance and Memory Improvements:**

[...]

PR #833, opened 2026-06-11
Fast ai

jdolan:
Merges in https://github.com/jdolan/quetoo/pull/824


src/game/default/g_ai_grid.c
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+ */
+
+#include "g_local.h"
+
```
jdolan, line 23: @zardoru It feels like we're missing some other copyright here. Where did this grid code come from?

zardoru: I wrote the data structure work here! it's called "grid" because that's what the pathfinding code it was originally written for was called (for the quake 2 vortex mod, because it dropped nodes in a 16 by 16 basis from the sky and downwards), "grid_ds" for "grid data structures".

jdolan: Oh! Well, that's super cool 😎

PR #845, opened 2026-06-12
Use async HTTP requests for UI and stats (#835)

### LINKED ISSUE #835: Use async HTTP requests for UI / stats where possible.
There's a noticeable delay awaiting the HTTP responses for stats / leaderboard API calls when the menu loads, which manifests as a black screen on launch, or a short freeze on re-displaying those menus.

It would be great if we could use the async callback flavor of HTTP calls, and push a custom MVC_NOTIFICATION SDL_Event from the completion callbacks to inform the UI to refresh.

Eclipse1982:
## Summary

Replaces the blocking HTTP requests in the UI / stats code paths with asynchronous requests, so the client no longer shows a black screen at launch (GUID hash fetch) or freezes when opening the Leaderboard and Stats menus while waiting on the stats API.

Completion callbacks run on the HTTP session thread; results are staged and marshaled to the main thread via `MVC_NOTIFICATION_EVENT`, and view controllers refresh in `respondToEvent` — following the pattern suggested in the issue and already used by `NOTIFICATION_SERVER_PARSED`.

Fixes #835

## Changes

- Added `Net_HttpGetInstanceAsync` / `Net_HttpGetInstancesAsync` — async counterparts to the JSON instance helpers — and exposed them through the cgame import (`cgi.HttpGetInstanceAsync`, `cgi.HttpGetInstancesAsync`).
- `Cl_InitGuidHash` now fires an async request at startup; the hash is applied to `guid_hashed` on the main thread via a new `NOTIFICATION_GUID_HASHED` event, so startup never blocks on the network. The call was moved after `Ui_Init`, because `MVC_NOTIFICATION_EVENT` is only registered with SDL once ObjectivelyMVC initializes — a completion firing earlier would push event type `0` and the hash would be silently dropped.
- `LeaderboardViewController` fetches asynchronously (initial load and sort-column clicks); rows are staged under a mutex and applied in `respondToEvent` on `NOTIFICATION_LEADERBOARD_FETCHED`. It also reloads when the GUID hash arrives so the local player's row highlights.
- `StatsViewController` fetches asynchronously, showing placeholders until `NOTIFICATION_STATS_FETCHED` delivers the response; it refreshes when `NOTIFICATION_GUID_HASHED` arrives (previously a cold start could show "Sign in" until the next visit).
- Generation counters discard stale responses when a newer fetch supersedes an in-flight one; callbacks never touch UI state directly, so there are no dangling view controller pointers.
- Added `check_http` unit tests covering both new async helpers against the local test HTTP server.

## Testing

- [x] Tests pass (`make check`)
- [x] Tested in-game on Linux (Debian 12)

## Notes

- Built and tested on Debian 12 with the from-source dependency stack from `linux/docker/Dockerfile.build`. The changed files compile without warnings (pre-existing warnings elsewhere in the tree are untouched).
- `make check`: 18/19 suites pass, including `check_http` with the two new async tests. The one failure is `check_filesystem` ("Failed to load quetoo.cfg") — an environment issue in the build container (no game data installed), unrelated to this change.
- In-game smoke test against the live stats API (`giblets.quetoo.org`), headless under Xvfb/llvmpipe: client reaches the main menu with stats placeholders immediately, the GUID hash arrives asynchronously and the Stats tab populates, the Leaderboard tab loads live rows, sort-column clicks refetch and reorder without any UI stall, and `selectOwnRow` highlights the local player after the async reload. No warnings or errors in the console log.


jdolan:
Fantastic! Thank you. I'm working on some improvements to the JSON deserialization code here first. I'll commit that and up-merge into your branch and get this over the finish line. This is a big win for user experience.

jdolan:
Closing in favor of https://github.com/jdolan/quetoo/pull/851 which borrowed from this.

PR #850, opened 2026-06-14
Feature/trigger void

skies912:
## Summary

Introduces `trigger_void` ent via Claude. It's to replace using `trigger_hurt` in space maps or maps with death-pits. When the player falls into the top face, their POV switches to third person an cam is locked at entry (cam of shame) while they fall to their death when they hit the bottom face. Has single spawnflag that enables teleportation instead of death to a random DM respawn OR to a defined tele_dest pad via the `target` key. Also defines new void-death sound for each player model.

## Notes

- Bottom face of the brush must not touch other solid faces, otherwise it prevents death.
- We still need to make the actual void-death sounds


jdolan:
So is the idea that you actually watch your own corpse fall away into oblivion? 

Do you create a really deep, dark, unlit pit of death beneath the void brush to achieve the effect?

Would it be adequate to just have trigger_void cause a special 3rd person view on death, without anchoring it to where you touched the trigger, etc?

skies912:
> So is the idea that you actually watch your own corpse fall away into oblivion?
> 
> Do you create a really deep, dark, unlit pit of death beneath the void brush to achieve the effect?
> 
> Would it be adequate to just have trigger_void cause a special 3rd person view on death, without anchoring it to where you touched the trigger, etc?

1) yes. "Cam of shame" lol. Any spectator spectating the falling player will share the view (side effect)(not a bad thing?)
2) Did some testing here. A pit of around 512u seems fine for a quick "watch myself fall and then die" sequence. For space maps it's fine as just a trigger. I'll test out some actual pits but should be fine.
3) was hoping a sufficient enough fall would shrink (perspective  via anchoring) player model enough to sell the illusion of not having an infinite void. Maybe a shader to help? Idk.
Lemme load up q3 and rewatch how they did it..

jdolan:
I remade a "lite" version of this here: https://github.com/jdolan/quetoo/pull/941

@skies912 I made the death cam general-purpose, and wired it up to all non-player-projectile deaths by default (so anything humiliating like falling, drowning, getting crushed, trigger_hurt, etc). But you can also force it on for all deaths with `g_death_cam 2`. Was wondering if you'd give it a spin and see what you think. I'm almost tempted to turn it on for all deaths by default. It's kinda fun.

PR #851, opened 2026-06-15
Use Objectively's new JSONContext API for REST client functionality.

jdolan:


Eclipse1982:
CI is red on `build-linux` only; `build-windows` is green. This is a compiler/toolchain difference, not Windows vs Linux: the `build-linux` job pins `ubuntu-22.04`, whose system `gcc` (11.4) is what the autotools `./configure && make` release build uses, and GCC 11 rejects the two file-scope static descriptors in `src/server/sv_game.c`:

```
sv_game.c:220:50: error: initializer element is not constant
sv_game.c:220:50: note: (near initialization for ‘(anonymous)’)
sv_game.c:233:53: error: initializer element is not constant
sv_game.c:233:53: note: (near initialization for ‘(anonymous)’)
```

Lines 220/233 are the `static const JSONProperties sv_frag_properties` / `sv_capture_properties`, initialized via `MakeJSONProperties(...)`. The macro expands to a compound literal whose `.properties` is the address of a *nested* `(const JSONProperty[]){ ... }` whose own elements are per-field `(JSONProperty){ ... }` compound literals. GCC 11 doesn't treat that nested array's address as an address constant in a static initializer (hence the `(anonymous)` in the note); MSVC accepts it, which is why `build-windows` is green. Worth noting this is a GCC-version thing, not a categorical GCC limitation — Objectively's own `Tests/Objectively/JSON.c` keeps the same file-scope-static `MakeJSONProperties` form (lines 281/292) and its CI passes, because that job runs on `ubuntu-latest` (newer GCC). The portable fix that also works on the pinned 22.04 toolchain is to drop static storage.

Both descriptors are used only inside `Sv_PostStats` — `&sv_frag_properties` at line 258, `&sv_capture_properties` at line 274, each in its own `if` block. Moving each to a local `const JSONProperties` next to its use fixes it: an automatic object permits a non-constant initializer, so the nested literal is materialized at runtime and GCC/MSVC both accept it. The property lists move verbatim (just `static const` → `const`, relocated into the function), and the `&` call sites don't change:

```c
static void Sv_PostStats(...) {
    ...
    if (frags_len) {
        const JSONProperties sv_frag_properties = MakeJSONProperties(g_frag_t, /* ...unchanged... */);
        // ... &sv_frag_properties as before
    }
    if (captures_len) {
        const JSONProperties sv_capture_properties = MakeJSONProperties(g_capture_t, /* ...unchanged... */);
        // ... &sv_capture_properties as before
    }
}
```

It's cold POST-path code, so per-call construction is a non-issue, and it scopes each descriptor to the branch that uses it. Want me to push this to `claude/determined-wright-d63fe4`?

Issue #856, opened 2026-06-16
Android Platform Port


jdolan:
## Overview

Port Quetoo to Android using SDL3 and OpenGL ES 3.0. The Android port shares most of its work with the [iOS port (#855)](https://github.com/jdolan/quetoo/issues/855) — particularly the GL ES 3.0 renderer adaptation — but has meaningful differences in the build system, module loading, and graphics API options.

---

## Graphics: OpenGL ES 3.0

Same `#define QUETOO_GLES` dual code path as the iOS port. See [#855](https://github.com/jdolan/quetoo/issues/855) for the full breakdown. The sacrifices are identical: Peter Panning in shadow maps (accepted), wireframe debug mode dropped, GPU timer queries dropped.

**Android-specific option:** Vulkan is available on Android 7.0+ (API 24, ~95% of active devices). However, a Vulkan renderer would be a near-complete rewrite of `src/client/renderer/`. OpenGL ES 3.0 is the recommended starting point — Vulkan can be revisited as a future optimization.

---

## Architecture: Dynamic Module Loading

Unlike iOS, **Android fully supports `dlopen()`**. The existing `game.so` / `cgame.so` module architecture is preserved without modification. This is a significant advantage over the iOS port.

---

## Build System

This is the most Android-specific work item. There is currently no Android build infrastructure.

- **Android NDK + CMake**: add `CMakeLists.txt` as an alternative build system (benefits both Android and iOS). SDL3 provides an Android CMake integration out of the box.
- **Gradle wrapper**: standard Android Studio project structure in `android/`
- **ABI targets**: `arm64-v8a` (primary), `armeabi-v7a` (legacy, optional), `x86_64` (emulator)
- Game data (`default.pk3`) needs to be bundled as an Android asset or downloaded on first launch via the existing updater

---

## Input

Same virtual joystick / tap / swipe approach as iOS. SDL3's gamepad API covers Bluetooth controllers (Xbox, PS, generic HID). Android also supports physical keyboards on some devices/form factors — the existing keyboard input path would work there with no changes.

---

## Dependencies

| Library | Android Status |
|---|---|
| SDL3 | ✓ Official Android support, ships Java glue layer |
| OpenAL-soft | ✓ Android target (uses OpenSL ES / AAudio backend) |
| PhysicsFS | ✓ Portable, NDK compatible |
| libcurl | ✓ Available for Android NDK |
| glib2 | ⚠️ Same risk as iOS — no official NDK support, likely needs replacement |

[...]

Issue #855, opened 2026-06-16
iOS Platform Port


jdolan:
## Overview

Port Quetoo to iOS using SDL3 (which ships first-class iOS support) and OpenGL ES 3.0 for rendering. The engine's game logic, physics, networking, and audio are all portable C with no platform-specific dependencies. The renderer requires a targeted adaptation rather than a rewrite.

This issue assumes [ObjectivelyMVC mobile support](https://github.com/jdolan/ObjectivelyMVC/issues/38) lands first, since the in-game UI is built on ObjectivelyMVC.

---

## Graphics: OpenGL ES 3.0 Alongside GL 4.1

The pragmatic path is an OpenGL ES 3.0 code path that coexists with the existing GL 4.1 Core Profile path. Since iOS and desktop produce separate builds, there is no runtime overhead — the right path is selected at compile time via `#define QUETOO_GLES`.

### What ES 3.0 Supports (no changes needed)
- `GL_TEXTURE_2D_ARRAY` — shadow atlas, lightmaps ✓
- UBOs with `std140` layout — `uniforms_block`, `lights_block` ✓
- `layout(location = X)` for vertex attributes ✓
- `sampler2DArrayShadow` for PCF shadow lookups ✓
- Float textures, multiple render targets, VAOs ✓

### The Dual Code Path (estimated ~5–6 `#ifdef` sites)

**`shaders/version.glsl`** — single file included by all shaders:
```glsl
#ifdef QUETOO_GLES
  #version 300 es
  precision highp float;
  precision highp sampler2DArray;
  precision highp sampler2DArrayShadow;
#else
  #version 410
#endif
```

**`r_shadow.c`** — remove `glEnable/Disable(GL_DEPTH_CLAMP)` under `#ifndef QUETOO_GLES`

**`r_main.c`** — remove `glPolygonMode` wireframe under `#ifndef QUETOO_GLES`

**`r_main.c`** — request ES context: `SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES)` and `SDL_GL_CONTEXT_MAJOR_VERSION 3`

### Sacrifices on ES 3.0
| Feature | Impact | Mitigation |
|---|---|---|
| `GL_DEPTH_CLAMP` | Peter Panning may return in shadow maps | Accept it — barely perceptible on a phone screen at normal viewing distance |
[...]

PR #860, opened 2026-06-18
Android platform port: SDL3 + OpenGL ES 3.0 client (#856)

Eclipse1982:
Tracks #856 (Android); shares the GL ES renderer, glib2 replacement, and virtual-controls work with the iOS port #855.

This brings up a **client-only** Android port (SDL3 + OpenGL ES 3.0). It currently builds for `arm64-v8a` (device) and `x86_64` (emulator), packages into a signed APK, boots through full init to a rendered GL ES 3.0 **main menu**, and **renders maps in-game with full textures and lighting** on real hardware (tested on a Samsung Galaxy S24 Ultra, Adreno 750).

The desktop GL / Windows / Linux / macOS builds are untouched — every change is gated behind `QUETOO_GLES` (renderer) or `__ANDROID__` (platform).

### 📄 Read first: [`android/PORT_CHANGES.md`](https://github.com/Eclipse1982/quetoo/blob/feature/android-port-v78/android/PORT_CHANGES.md)
A retrospective of everything that had to change, with root causes. The other `android/*.md` files (`PORTING.md`, `RENDERER_GLES.md`, `DEPENDENCIES.md`, …) go deeper per subsystem.

### Highlights
- **Build:** root `android/CMakeLists.txt` for the NDK; the full dependency stack (SDL3/image/ttf, OpenAL-soft, PhysicsFS, libsndfile, Objectively/MVC, mbedTLS+libcurl) cross-compiled via scripts; Gradle-free APK assembly. All shipped `.so` are 16 KB page-aligned.
- **glib2 → `android/qglib`:** an in-tree, SDL3-backed reimplementation of exactly the glib subset Quetoo uses (7 container types + primitives), differential-tested against system glib-2.0. Drops a heavy dep on every platform.
- **GL ES 3.0 renderer:** the bulk of the shader work is plain **GLSL-ES-3.0 strictness vs desktop GL 4.x** and should apply to iOS 3.0 (#855) too — interface blocks → struct varyings, mandatory default-precision preamble, implicit int→float fixes, `textureQueryLod` → derivative-based LOD, the voxel light-index TBO → integer 2D texture, `MAX_LIGHTS` 96→32 (uniform-vector limit), `glDrawElementsBaseVertex` resolved at runtime, and the tonemap presented straight to the default framebuffer (Adreno rejects the scene-FBO `glBlitFramebuffer`).
- **One gnarly runtime bug worth calling out:** parallax occlusion mapping wrote its result through an `inout` parameter and `return`ed early when disabled — the **Adreno GLSL compiler drops the `inout` write-back across that early return**, leaving the texcoord divergent so every BSP surface sampled a flat top mip (world rendered as flat average colors, meshes fine). Fixed by setting the texcoord in the caller scope. Details in `PORT_CHANGES.md §5c`.
- **Online stack:** HTTPS via mbedTLS + libcurl behind Objectively's `URLSession`, with a shipped `cacert.pem` (`CURLOPT_CAINFO`).

### Notes for review
This is a large change set (clean re-port off current `main`, ~12 commits / ~128 files, mostly additive). It's posted **for review and discussion**, not as a claim of merge-readiness — known-incomplete items (virtual on-screen controls wiring, ES shadow atlas, on-demand asset download) are listed in `PORT_CHANGES.md §7`. Happy to split the cleanly-upstreamable pieces (e.g. the GLSL ES 3.0 conformance fixes, the `qglib` shim) into smaller standalone PRs if that's easier to land.


jdolan:
Wow! I think we should probably migrate all glib usage to SDL _first_, in a separate PR, to reduce the size / risk of this one. Is that difficult to cherrypick into its own branch?

jdolan:
Actually.. why is glib not "available" on Android? Is it that difficult to compile? Is it so different from OpenAL, physicsFS, etc?

Eclipse1982:
**Re: why glib isn't "available" on Android** — it's the build system + transitive deps, not glib itself. SDL3, OpenAL-soft, PhysicsFS, and libsndfile all ship NDK-friendly cross-compile paths (CMake toolchain or a simple configure). glib only builds under Meson and drags in a chain — libffi, PCRE2, proxy-libintl/gettext, iconv — none of which cross-compile cleanly to the NDK without patching each one. So it's not conceptually different from the others; its dependency tree is just hostile to the NDK, which is a lot of yak-shaving for the ~5% of glib that Quetoo actually touches.

**Re: migrate to SDL first / cherrypick** — two parts:

1. A pure glib→SDL migration isn't quite possible: SDL3 covers the *primitives* we use from glib, but it has no equivalent for the container types, and those are most of the usage — `GArray`, `GList`, `GSList`, `GPtrArray`, `GQueue`, `GHashTable`, `GString`. That's why I went with `android/qglib`: a small in-tree shim that delegates the primitives to SDL3 and reimplements those 7 containers (+ `GChecksum`/`GDateTime`/`GRand`/`GRegex`) with glib-faithful semantics. It's differential-tested against system glib-2.0 (~350 checks, identical results), so it's a drop-in on every platform, not just Android.
2. It's easy to cherrypick. The shim is fully self-contained in `android/qglib/`, and on the Android side it's wired in purely at the build level — a `compat/glib.h` that forwards `<glib.h>` to `qglib.h`, first on the include path — so `src/` is completely untouched. To make it a standalone cross-platform PR, the only addition is flipping that include in `src/quetoo.h` and adding qglib to the autotools build so desktop uses it too. Happy to spin that out first; it drops the heavy glib dep on every platform, which is probably worth landing on its own merits regardless of Android.

If that works for you I'll split this into: (a) the qglib shim replacing glib everywhere, (b) the GLSL-ES-3.0 conformance fixes, then (c) the Android platform/build layer on top.

jdolan:
I'm gonna close this, because there have been significant changes to our multiplayform strategy since this PR was openend:

- Removed glib entirely
- Pursuing SDL3_gpu over GL ES3.

There may be some useful bits in here for general Android portability that we can cherry pick once the SDL_gpu stuff lands.

PR #861, opened 2026-06-18
renderer: add an OpenGL ES 3.0 code path

Eclipse1982:
## Summary

Adds an OpenGL ES 3.0 code path to the renderer, behind a new `--enable-gles`
configure option (off by default), to support the Android (#856) and iOS (#855)
ports. Desktop GL builds are completely unaffected.

This is the renderer-only slice of the Android port work, separated out so it can
be reviewed and merged on its own.

## Changes

- Guard the ES-specific renderer code behind `QUETOO_GLES` (`<GLES3/gl3.h>` on ES,
  the desktop glad loader otherwise). All changes are confined to
  `src/client/renderer/`.
- `r_gl_compat.h`: a small runtime `R_DrawElementsBaseVertex` entry where ES lacks
  the desktop GL one, plus no-op shims for ES-absent desktop calls.
- GLSL ES-3.0 conformance in the shared shader bodies (struct varyings, explicit
  `float()` casts, sampling BSP materials at the vertex texcoord). These are
  semantically identical on desktop GL; a `version_es.glsl` preamble is selected by
  `r_program.c` on ES.
- `configure.ac`: `--enable-gles` defines `QUETOO_GLES` and links the ES driver
  (`libGLESv3`) instead of desktop GL.

## Testing

- [x] Builds without warnings (`make -j$(nproc)`) — full desktop build
  (`quetoo`, `quetoo-dedicated`, `quemap`, `game.so`, `cgame.so`) with `--enable-gles`
  **off**: 0 warnings, 0 errors. No desktop regression.
- [x] `--enable-gles` configure verified (defines `QUETOO_GLES`, links `libGLESv3`);
  the ES code path compiles via the Android NDK cross-build.
- [x] Tested in-game on **Android (Adreno 750 device + x86_64 emulator)** — boots to
  the GL ES 3.0 main menu and renders.
- [ ] `make check` — not run (the pre-existing `check_master` link failure on `main`
  is unrelated to this change).

## Notes

All engine changes are inside `src/client/renderer/` and guarded by `QUETOO_GLES`;
with the option off (the default) the desktop renderer is unchanged. The vendored
`glad-gles` loader from the port branch is intentionally **not** included here — the
ES path uses the system `<GLES3/gl3.h>` entry points directly, so no loader is
needed.


jdolan:
We're gonna pursue an SDL3_gpu renderer backend instead of GL3 ES, which is already EOL on iOS.

Issue #867, opened 2026-06-19
Current gameplay bugs June '26


jdolan:
- [x] Hook beam starts far in front of player in 1st person view and just doesn't look right. Maybe `cg_fov 130` related? But it should start from _yo deeick_ or thereabouts. Should be able to use either `Cg_Self()->s.origin` newer `Cg_MuzzleOrigin()` to help with this. I think beams have an overloaded state field to identify "my beams".
- [x] Hook travels through teleporters, lol. Kinda hilarious, but should probably disable.
- [x] `sv_min_clients` counts human spectators as players when it should not.
- [x] When two players are touching and one fires either a rocket or a grenade, the projectile suspends in space, vertically oriented, and does not blow up. LOL. When the players part, the projectile travels in some undefined direction, with incorrect orientation. Obviously the right behavior is that the projectile should immediately explode and not get into this weird suspension state. But this could also be a hilarious easter egg with a little tweaking, where you could turn grenades into "sticky grenades" if you manage to plant one right on your opponent at point blank range.
- [x] Is it just me or is Quake lightning ammo pickup and initial quantity just too low? Or the burn rate is too high? Idk. I know the weapon is OP, but god damn does it piss through ammo.

PR #869, opened 2026-06-19
Detonate point-blank explosives instead of suspending them (#867)

Eclipse1982:
## Summary

When a player fired a rocket or grenade while flush against another
player, the projectile suspended in space (vertically oriented, inert)
instead of exploding, then flew off in an undefined direction once the
players separated. Part of #867.

`G_ImmediateWall()` traces from the shooter to the projectile's spawn
point against `CONTENTS_MASK_SOLID` to accommodate firing flush against a
wall (it pulls the projectile back to the shooter's origin). A point-blank
enemy is also solid, so the projectile was relocated *into* both bodies,
where it sat in an all-solid state and never ran its touch handler.

## Changes

- `g_ballistics.c`: add `G_ImmediateImpact()`, which keeps the existing
  wall accommodation but detonates the projectile immediately when the
  obstruction is a damageable entity (point-blank enemy). Apply it to the
  rocket, quake rocket, grenade and hand-grenade spawners.

## Testing

- [ ] Tested in-game on 

## Notes

Scoped to the explosives named in the issue that link normally on spawn.
`G_QuakeGrenadeProjectile` is intentionally left out: on current `main`
it is never linked by itself or its caller, so its lifecycle is unusual
and unrelated to this fix.

### DIFF
--- a/src/game/default/g_ballistics.c
+++ b/src/game/default/g_ballistics.c
@@ -54,6 +54,34 @@ static bool G_TakesDamage(g_entity_t *ent) {
 }
 
 /**
+ * @brief Resolves an explosive projectile that spawns flush against a wall or
+ * an enemy. Against world geometry the projectile is pulled back to the
+ * shooter's origin so it detonates against the surface on its first move.
+ * Against a point-blank enemy it is detonated immediately, rather than
+ * suspending inside the entity with no valid trajectory (#867).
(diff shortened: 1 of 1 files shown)


jdolan (approved):
Lmk if you wanna test this on your local server or something, because the bots might not stand still enough for you ;)

Eclipse1982:
I don't have rox on my server though.

PR #876, opened 2026-06-19
renderer: optional Vulkan/RTX rendering backend (BSP hardware ray tracing) [WIP]

Eclipse1982:
## Summary

Adds an **optional** Vulkan rendering backend with hardware ray tracing (RTX) of
BSP world geometry, selectable at runtime and fully gated behind a build flag so
the default OpenGL path is unaffected. This is a large work-in-progress opened as
a **draft** for early visibility and direction — not for merge yet.

No tracking issue exists for this work.

## Changes

- New Vulkan/RTX renderer backend (phases 0–1: device/swapchain bring-up, clear +
  present, lifecycle wiring into the engine).
- Hardware ray tracing of loaded BSP geometry from the player view, including
  BSP-light shading, soft shadows via jittered shadow rays, and temporal
  accumulation.
- Per-material diffuse: real diffuse textures sampled in the RTX closest-hit via a
  bindless descriptor array; mesh models placed into the ray-traced scene.
- Vulkan 2D pass renders the ObjectivelyMVC menu/HUD/console and an RTX overlay,
  including uploading ObjectivelyMVC textures to the bindless array.
- Runtime OpenGL/Vulkan backend toggle in the settings menu; backend torn down on
  `r_restart`.
- Visual Studio build wiring for the Vulkan backend; `.gitattributes` to normalize
  line endings (LF).
- Validation harnesses (raster + ray tracing) and integration/run-step docs.

## Testing

- [ ] Tested in-game on 

**Build verified** on an RTX 3090 box (Ubuntu 24.04, gcc 13, Vulkan 1.3.275):
`autoreconf -i && ./configure --enable-vulkan && make -j` completes with **0
warnings / 0 errors**, producing `quetoo`, `quetoo-dedicated`, `cgame.so`,
`game.so`; the `quetoo` binary links `libvulkan.so.1` and exports the RTX entry
points (`BUILD_VULKAN=1`). The default OpenGL build is unaffected.

`make check` not run (configured without `--with-tests`; this change adds no unit
tests). In-game RTX rendering was validated on the RTX 3090 in earlier iterations;
this exact rebased tree has been built but not yet re-run in-game — I'll tick that
box after a runtime smoke test.

## Notes

- Opened as a **draft** — feedback on scope and approach is welcome before this is
  polished for review. Happy to split it into smaller, sequenced PRs (e.g. backend
  scaffolding first, then RTX shading, then UI) if that's easier to review.
- The 28 commits are rebased cleanly onto current `main`.
- **Heads-up on deps (not caused by this PR):** current `main` builds against the
  *pre-2.x* Objectively / ObjectivelyMVC API (e.g. it still uses `MutableArray` and
  the `RESTClient`/`JSONSerializers` API). The `HEAD` of both dependency repos has
  since merged "Objectively 2.x", which removes `MutableArray` — so building `main`
  against dependency `HEAD` (as `.github/workflows/build.yml` resolves) currently
  fails to compile. I built against the last pre-2.x commits
  (Objectively `476286b`, ObjectivelyMVC `4947d7c`).


Eclipse1982:
Closing in favor of a clean re-architecture. The Vulkan/RTX work is being restarted with a proper renderer-backend boundary first (de-GL'd cgame ABI + backend vtable), then a Q2RTX-derived path tracer on top, landed as small reviewable increments. The branch is preserved for reference.

PR #882, opened 2026-06-20
Measure client ping at true packet arrival time (#877)

### LINKED ISSUE #877: [feedback] In-game ping seems inaccurate
the server browser ping times fluctuate wildly between refreshes and in-game ping displayed on the scoreboard always appears at least 20ms higher for me compared to directly pinging the server outside the game

Eclipse1982:
## Summary

In-game ping was consistently reported ~20 ms higher than an external `ping` to the same server, and both the scoreboard and server-browser values jittered. The root cause is that a dedicated server serviced its UDP socket only on 40 Hz simulation-frame boundaries: a client's acknowledgement could sit in the kernel receive buffer for most of a 25 ms tick before being timestamped, so every round-trip sample was quantized to (and inflated by) the frame interval.

Fixes #877

## Changes

- **Server:** while idling out the remainder of the frame interval, a dedicated server now blocks on the socket (`Net_Sleep`) instead of `SDL_Delay`, stamping and reading each datagram at its true arrival time. This yields millisecond-resolution latency samples and also lets the server answer connectionless queries (e.g. server-browser pings) without frame-boundary latency. Movement is command-driven and already processed on receipt, so only wall-clock timing shifts — not the deterministic simulation.
- **Scoreboard:** report the most recently measured round-trip, so the displayed ping tracks live latency without smoothing lag.
- **Server browser:** exponentially smooth the per-request ping across refreshes so the listed value converges instead of bouncing on each one-shot status reply.

## Testing

- [x] Tested in-game on Linux dedicated (x86_64) + Windows client, over a real internet path

Live A/B against a real server (client ICMP baseline ~21.5 ms):

| Build | Scoreboard ping |
|-------|-----------------|
| Current `main` (frame-quantized) | **42 ms** (≈ ICMP + 20) |
| This PR | **~21 ms**, tracking the true RTT |

## Notes

- `make check` is blocked by the pre-existing `check_master` link failure (`undefined reference to Net_HttpPostAsync`) on `main`, unrelated to this change; the only compiler warnings in a full build are pre-existing unused-variable warnings in `src/tests/check_mem.c`. The engine and all changed files compile warning-free.
- The poll-on-socket change is gated to dedicated servers; listen servers already spin the main loop at render rate and were never frame-quantized.
- The server-browser change is client-side and was verified by build + logic review (the live in-game verification covered the server-side scoreboard path).

### DIFF
--- a/src/client/cl_server.c
+++ b/src/client/cl_server.c
@@ -129,7 +129,17 @@ void Cl_ParseServerInfo(void) {
       line = end ? end + 1 : NULL;
     }
 
-    server->ping = Clampf(quetoo.ticks - server->ping_time, 1u, 999u);
+    const int32_t sample = Clampf(quetoo.ticks - server->ping_time, 1u, 999u);
+
+    // smooth across refreshes so the displayed ping converges instead of
+    // bouncing on each request's frame-quantized round-trip
+    if (server->ping_smoothed <= 0) {
+      server->ping_smoothed = sample;
+    } else {
+      server->ping_smoothed = (server->ping_smoothed * 3 + sample) / 4;
+    }
+
+    server->ping = server->ping_smoothed;
     server->error[0] = '\0';
 
   } else {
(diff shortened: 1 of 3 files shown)


Eclipse1982:
Superseded by https://github.com/jdolan/quetoo/pull/883 (same commit, pushed as an in-repo branch instead of from my fork).

PR #891, opened 2026-06-22
net: fragment oversized netchan messages for Internet transmission (#888)

### LINKED ISSUE #888: High player counts + no PVS/PHS or server side culling = datagram overflows
Got this bug report from RailWolf who runs a "botopia" server with 24+ bots 🤖

```
[BOT] Widow blew up :explosive:   
│[BOT] Gunner 1 chewed on [BOT] Enforcer 's boomstick :shotgun: 
│[BOT] Medic 1 tried to invade newbie's personal space :telefrag: 
│newbie wants some
│[BOT] Berserker chewed on [BOT] Stroggo 1's boomstick :shotgun: 
│Sv_SendClientDatagram: Frame exceeds MAX_MSG_SIZE (4441) 
│Server shutdown...  
│Game shutdown...
│Game module shutdown... 
│Game down
│Sending shutdown to 52.6.189.159:1996
│Server down 
│main: Error detected, recovering.. 
```

You can see clearly here that a client connected and immediately overflowed due to receiving _lots_ of delta entities for the 24 bots, their projectiles, temp events, etc..

This never happens locally (listen server), because that uses the loopback connection which need not enforce `MAX_MSG_SIZE`. But it makes perfect sense that this would happen on a public Internet server using real UDP networking.

So what should we do here? Allow the message buffer to increase / be unbound, and then chop it up into multiple 1400 byte sub-packets that are fit for Internet transmission? Would that work? They may arrive unordered, which may or may not be a problem depending on if we can make each sub-packet's delta entities block self-contained.

Should we instead introduce some form of coarse server-side PVS? Have clients actually send the BSP block node indexes that they can see each time they receive a server frame? Delta compress that list as well? (That's actually kinda interesting).

Eclipse1982:
## Summary

On dedicated Internet servers with high entity counts (e.g. RailWolf's 24+ bot "botopia" server), `Sv_SendClientDatagram()` builds a single frame (player state + all delta entities) that exceeds the UDP MTU. The frame was sent as one datagram relying on IP-layer fragmentation — any dropped IP fragment loses the whole frame — and a frame exceeding `MAX_MSG_SIZE` crashed the dedicated server outright via `Com_Error(ERROR_DROP)`. Listen servers never hit this because the loopback path is not MTU-bound.

This adds Q3-style application-layer message fragmentation to the netchan, as outlined in the issue. Messages larger than the MTU are split into sub-MTU fragments and reassembled by the receiver; this is transparent to all game logic (no entity culling or gameplay changes).

Fixes #888

## Changes

- **`net_chan.c` — `Netchan_Transmit()` / new `Netchan_TransmitNextFragment()`**: a message that exceeds `FRAGMENT_SIZE` (1300 b) is staged in the netchan and sent one fragment per packet; callers drain via `Netchan_TransmitNextFragment()` while `unsent_fragments` is set. Every fragment carries the same sequence number tagged with `FRAGMENT_BIT`, plus a 16-bit offset/length header; the sequence advances only after the final fragment. A message that is an exact multiple of `FRAGMENT_SIZE` is terminated by a final, empty fragment.
- **`net_chan.c` — `Netchan_Process()`**: reassembles in-order fragments per channel, deferring delivery (and reliable acknowledgement) until the message is complete.
- **Bit layout**: `FRAGMENT_BIT` is sequence bit 30; bit 31 remains the reliable flag (Q2 style), leaving a 30-bit sequence (`SEQUENCE_MASK`). This resolves the bit-31 conflict noted in the issue.
- **`net_types.h`**: `FRAGMENT_BIT` / `FRAGMENT_SIZE` / `SEQUENCE_MASK` constants and per-channel outgoing/incoming fragment state on `net_chan_t`.
- **`sv_send.c` / `cl_cmd.c`**: drain pending fragments after each transmit that can produce a large message (frame path, loading-client reliable flush, demo path, client command path).
- **`sv_send.c`**: `Sv_SendClientDatagram()` no longer crashes the whole server on frame overflow — oversized frames are fragmented for transmission, and a frame that exhausts the message buffer is skipped with a warning.
- **`PROTOCOL_MAJOR`** bumped to 2028 (core packet-header change; requires coordinated client + server update).

## Testing

- [x] Builds without warnings (`make -j$(nproc)`) — built clean in a Debian 12 container against current `jdolan/Objectively` + `jdolan/ObjectivelyMVC` HEAD; the changed files compile clean under `-Wall -Wextra`.
- [x] Tests pass (`make check`) — 19/19 PASS, including `check_net_message` (netcode delta round-trip).
- [ ] Tested in-game — not yet exercised on a live dedicated server under load; would appreciate a test on a high-bot server.

## Notes

- One deliberate deviation from the implementation sketch in the issue: rather than fully removing the `ERROR_DROP` in `Sv_SendClientDatagram()`, I demoted it to a non-fatal warning + skip. Fragmentation lifts the *MTU* limit but not the 16 KB message-buffer ceiling, so a guard is still warranted — but it must never take down the whole server (the reported symptom).
- Backwards-incompatible at the netchan layer only when fragmentation actually triggers; the `PROTOCOL_MAJOR` bump gates mismatched builds at connect.

### DIFF
--- a/src/client/cl_cmd.c
+++ b/src/client/cl_cmd.c
@@ -138,6 +138,12 @@ void Cl_SendCommands(void) {
 
       if (cls.net_chan.message.size || delta > 1000) {
         Netchan_Transmit(&cls.net_chan, NULL, 0);
+
+        // a large reliable message may have fragmented
+        while (cls.net_chan.unsent_fragments) {
+          Netchan_TransmitNextFragment(&cls.net_chan);
+        }
+
         cl.packet_counter[cl.sample_index]++;
       }
 
@@ -157,6 +163,12 @@ void Cl_SendCommands(void) {
       Cl_WriteMovementCommand(&buf);
 
       Netchan_Transmit(&cls.net_chan, buf.data, buf.size);
+
+      // a large reliable message (e.g. user info) may have fragmented
+      while (cls.net_chan.unsent_fragments) {
+        Netchan_TransmitNextFragment(&cls.net_chan);
+      }
+
       cl.packet_counter[cl.sample_index]++;
 
       Cl_InitMovementCommand();
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -48,7 +48,7 @@
  * of core net messages or serialized data types change. The game and client
  * game maintain `PROTOCOL_MINOR` as well.
  */
-#define PROTOCOL_MAJOR 2027
+#define PROTOCOL_MAJOR 2028
 
 /**
  * @brief The IP address of the master server, where the authoritative list of
--- a/src/net/net_chan.c
+++ b/src/net/net_chan.c
@@ -25,11 +25,14 @@
  *
  * packet header
  * -------------
- * 31  sequence
+ * 30  sequence
+ * 1  is this message a fragment of a larger message
  * 1  does this message contain a reliable payload
  * 31  acknowledge sequence
  * 1  acknowledge receipt of even/odd message
  * 8  qport
+ * [16  fragment offset, if fragmented]
+ * [16  fragment length, if fragmented]
  *
  * The remote connection never knows if it missed a reliable message, the
  * local side detects that it has been dropped by seeing a sequence acknowledge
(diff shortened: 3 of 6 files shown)


jdolan:
We already have packetization of large network messages. I wrote it like 6 years ago. It's right there in Sv_SendClientDatagram. Tho, there was a bug in its implementation (wrong constant referenced in the overflow check) that caused it to not behave correctly, and explode on RailWolf's server.

The root fix here is one line of code 😂 Closing this out.

Issue #840, opened 2026-06-11
[Feat. Req] Traverse selection ray using mwheel in editor


skies912:
When nested objects are present such as a light inside a `func_group` or even a surface below both, it can be difficult (or impossible) to select which to edit. I forgot to create a ticket for it when we discussed it awhile back. I think we came up with where you'd use the scroll wheel to cycle through the objects found in the ray when trying to select. Probably good idea to cap the distance of the object to 512 or 768u or something so you aren't selecting something across the map.

jdolan:
Yea, this is a great idea. So it's not just all-or-nothing, but you can rotate through the entities in the path of your trace. Very smart.

skies912:
Completed via https://github.com/jdolan/quetoo/commit/2b001b0876d5f7d95d6427af0d8c07ae814e73c9 ([feature/material-stage-editor](https://github.com/jdolan/quetoo/compare/feature/material-stage-editor))

Waiting for SDL_GPU to be finished so it can be re-implemented properly.

PR #895, opened 2026-06-23
In-game material stage editor + editor UI overhaul

skies912:
**Work in Progress**
Generated with Claude Code
## Summary

Extends the in-game editor with live material stage editing, entity/mesh editing, unified ray selection, and a UI overhaul. Currently merged onto `jdolan-sdl-gpu-renderer` to track the Vulkan renderer rewrite, so the diff against `main` includes that branch's history pending its own merge.

### Material stage editor
Add/remove/tune material stages (scroll, pulse, envmap, warp, terrain, animation, blend, …) live in-game, then Save to the `.mat`. Chunked/progressive stage-view construction avoids the multi-second freeze when switching to a heavily-staged material; Add/Remove Stage patch the list in place instead of a full rebuild. Per-effect editor rows, editable/validated texture paths, live-editable Blend src/dest (updated for the renderer's new `cm_blend_t` enum).

### Entity tab
classname is the group-box title, remaining keys wrapped inside it; Team Master box shown only for `light` entities; Create/Delete moved onto the Entity tab; worldspawn `sky`/`message` live-editable with path validation.

### Mesh tab
Enabled/disabled based on the current ray-selection (kept always-present but disabled, rather than added/removed, to avoid a tab-bar width ratchet).

### Unified ray selection (#840)
One shared, nearest-first candidate list feeds all three tabs (entity def / surface material / mesh model), so the mouse wheel cycles through overlapping hits regardless of the active tab -- fixes decorative brushes (e.g. `common/dust`) blocking selection of what's behind them.

### Editor chrome
Top menu strip, docked panel, viewport-filling panel height derived from `window_bounds` so it adapts to resolution.

### Reusable widgets promoted upstream
`Scrollbar`/`ScrollThumb` (bound directly to a StackView's own opt-in `scroll`) and `KeyValueView`/`KeyValueTableView` (two-column key/value rows) were generalized and moved into ObjectivelyMVC itself -- see [jdolan/ObjectivelyMVC#43](https://github.com/jdolan/ObjectivelyMVC/pull/43), which this branch depends on.

## Notes for review
- Depends on jdolan/ObjectivelyMVC#43.
- Merged `jdolan-sdl-gpu-renderer` to track the Vulkan/SDL_gpu rewrite; adjusted the Blend dropdown + `Cm_FinalizeStage` for the new `cm_blend_t` enum, and dropped an obsolete OpenGL texture-bind workaround in `r_sky.c` (structurally impossible under the new texture model).
- Per-tab CSS binding is still unreliable in ObjectivelyMVC (a tab's own stylesheet doesn't reliably style its root/descendants); some styling remains C-hardcoded pending that fix.

## Testing
Builds clean (Debug, full solution incl. `quetoo.exe`/`libui`) against ObjectivelyMVC (rebased onto its current main) and `jdolan-sdl-gpu-renderer`. Exercised in-game: stage add/remove/tune/scroll, entity create/delete/edit, mesh tab enable/disable, ray-selection cycling across all three tabs, panel layout at multiple resolutions.


jdolan:
Oh man. The editor files in src/client/ui are all stale! I should have deleted them months ago but I guess I forgot. I'm sure those caused some confusion. You can nuke them in this PR, that would be helpful. I'd do it but I don't want to create conflicts for your branch.

skies912:
> Oh man. The editor files in src/client/ui are all stale! I should have deleted them months ago but I guess I forgot. I'm sure those caused some confusion. You can nuke them in this PR, that would be helpful. I'd do it but I don't want to create conflicts for your branch.

Done. Yeah that did confuse me. I figured one was newer n you just kept the old for ref/backup or something. Forgot to ask about it.

skies912:
Further commits use this branch on MVC: https://github.com/jdolan/ObjectivelyMVC/tree/feature/layout-updates

jdolan:
@skies912 I rebuilt the features this PR depended on in MVC and those are in main now. But there were significant changes from your first pass at those features:

- ScrollBar is now a member and subview of ScrollView
- Simplified ScrollBar to completely remove the top/bottom buttons, like most modern UX does
- Decided against KeyValueView, it is achievable with TableView
- Layout fixes and perf optimizations

This PR will have to be rebased (redone, really) against MVC@main. Lmk how you want to divide and conquer. Sorry for the churn. Claude often needs a guiding hand when making major changes to a library like MVC in order to keep it inline with the established patterns / philosophy.

skies912:
@jdolan I reset my local copy to account for the changes on MVC main. Gonna port over some of the remaining changes but still wanted to get your input on some. Eg: TableView needs updating to support JSON driven tables, etc.

jdolan:
> @jdolan I reset my local copy to account for the changes on MVC main. Gonna port over some of the remaining changes but still wanted to get your input on some. Eg: TableView needs updating to support JSON driven tables, etc.

Say more about that. Did you know about the `Input` class? Input binds a `Label` and `Control` together in a row. It's what the UI uses everywhere for e.g. the key binding form, the Create Server form, etc.. 

It's what I'm using for the [editor in main](https://github.com/jdolan/quetoo/blob/main/src/cgame/default/ui/editor/MaterialViewController.json#L11), actually. It even supports tabbing through to navigate.

So what's wrong with this approach that led you to create `KeyValueTableView` / `KeyValueView`? You can style it differently, if you were trying to tighten up margins and padding and things. Just give the containing `StackView` a class or whatever and tweak all you like.

Anyway, curious what motivated you to create this.

PR #902, opened 2026-06-29
Light glow & flare sync (target_light material toggles, light-style/pulse sync, sprite lighting refactor)

skies912:
Per-surface switchable and animated emissive glow stages + flares, driven by `target_light` entities and baked BSP lights, with **no BSP format/version bump** (reuses a spare surface bit).

## What this adds

- **`SURF_TOGGLE` (0x2000) + `toggle` stage keyword (`STAGE_TOGGLE`)** — a `target_light` switches a material's glow stages and flares on/off per-surface, bound by a `material` key on the light. quemap `FaceCmp` splits `SURF_TOGGLE` faces per brush side so panels sharing one material can chase independently.
- **Continuous brightness sync** — glow stages and flares track a controlling light's brightness via `toggle_alpha`. A *styled* `light` makes them flicker/pulse; a `target_light` switches them (style × on/off). Bound to baked BSP lights (phase-correct style/drift) or `target_light` entities.
- **Pulse→light sync** — a styleless `light` with a `material` key whose material has a `pulse` stage drives its own cast intensity from that pulse (e.g. jump pads breathing), using the same sine/clock/drift as the shader so the cast light and the glow stay phase-locked.
- **`target_light` extras** — `intensity` key (matches point lights, via `s.termination.y`) and a `back_and_forth` spawnflag (teamed chases bounce instead of wrapping).
- **quemap warning** when a `SURF_TOGGLE` face's material has no drawable/flare stage (catches mapper mistakes at compile).

## Sprite lighting refactor

Replaces `sprite_vs`'s bespoke distance-only lighting loop with a call to `light.glsl::vertex_lighting` via a local `common_vertex_t`, so particles get the full lighting model (ambient, occlusion, exposure, per-light diffuse, caustics) instead of flat distance attenuation. Billboards have no surface normal, so a camera-facing normal is synthesized for the lambert/sky terms. This also fixes dust/sprites staying flatly lit in shadowed corners — they're now properly grounded (net −10 lines in the shader path).

## Compatibility

`SURF_TOGGLE` is a spare surface bit, so existing BSPs and the current release branch load unchanged and simply ignore the new entity keys.

## Testing

Verified in-game (rage): per-surface glow + flare toggling, light-style sync on a flickering fluorescent, pulse→light on a jump pad, and grounded dust under the refactored sprite path. Perf under heavy particle scenes still wants more soak testing (the sprite path now does a few more per-vertex samples).

### DIFF
--- a/src/cgame/default/cg_entity_effect.c
+++ b/src/cgame/default/cg_entity_effect.c
@@ -192,7 +192,7 @@ void Cg_EntityEffects(cl_entity_t *ent, r_entity_t *e) {
       .origin = e->origin,
       .radius = ent->current.termination.x,
       .color = Color32_Color(ent->current.color).vec3,
-      .intensity = 1.f,
+      .intensity = ent->current.termination.y ?: 1.f,
       .source = ent
     });
   }
--- a/src/cgame/default/cg_flare.c
+++ b/src/cgame/default/cg_flare.c
@@ -22,6 +22,37 @@
 #include "cg_local.h"
 
 /**
+ * @brief A binding from a `SURF_TOGGLE` surface (material stage or flare) to the
+ * `target_light` or styled `light` that controls its brightness.
+ */
+typedef struct {
+
+  /**
+   * @brief True if a controlling light was resolved. Unbound surfaces render normally.
+   */
+  _Bool found;
+
+  /**
+   * @brief True if the controller is a `target_light` (switched on/off at runtime via
+   * `EF_LIGHT`); false for a static `light` (always on, brightness from its style).
+   */
+  _Bool switchable;
+
+  /**
+   * @brief The controlling light's origin, used to resolve its live entity for the
+   * `EF_LIGHT` switch state (switchable lights only).
+   */
+  vec3_t origin;
+
+  /**
+   * @brief The controlling light's `a`-`z` style string (may be empty) and phase offset,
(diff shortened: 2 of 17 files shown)


skies912:
[rage.zip](https://github.com/user-attachments/files/29446909/rage.zip)
Didn't want to push this in update just yet. Has updated stuffs used in testing this feature branch (target_lights and light-pulse sync).

jdolan:
Oof. This was pre-ObjectivelyGPU, huh? Rebasing it will probably require some effort.

We should talk about this feature. I'm not against it exactly, but it adds a fair amount of complexity, some overloads, and non-deterministic material <-> light binding that makes me a little squeamish. I wonder if there's another way, a simpler way, to achieve similar effects..

I do think this would look very cool. The switches on Frag Pipe, Pits, etc.. that use the server-side trigger_light along with a material animation could benefit from this. Or perhaps we borrow from that technique, I dunno.

jdolan:
Thank you for this, @skies912. It made the gap between light entities and material glow clear, and it started the design discussion.

Since June, the renderer moved to SDL_gpu, and the game and cgame moved to `common/`, so this branch cannot merge as it is. We want to continue the idea in a different direction: a material stage drives its own light, with `light.*` stage keys, and quemap emits the lights into the BSP. The design, the order of work, and the reasons are in #1084. Please add your comments there, especially about the chase case on your map.

Issue #864, opened 2026-06-18
Port the Quetoo renderer from OpenGL to ObjectivelyGPU (SDL3 GPU: Metal/Vulkan/D3D12)


jdolan:
## Context

OpenGL is deprecated on macOS (since 10.14) and effectively dead on iOS. SDL3's `SDL_gpu` gives us one C
abstraction over **Metal, Vulkan, and D3D12** — the only forward-looking, cross-platform option that
doesn't mean hand-writing three backends. We already ship on SDL3, so the runtime is available today.

**What changed since the original assessment.** That assessment asked
"is `SDL_gpu` up to the task?" and proposed *building* two things first: a "thin internal `Rgpu`-style
abstraction" and an ObjectivelyMVC `SDL_gpu` backend. **Both now exist.**

- **[ObjectivelyGPU](https://github.com/jdolan/ObjectivelyGPU)** is the `Rgpu`-style abstraction, shipped:
  `RenderDevice`, `Framebuffer` (MRT + depth + MSAA auto-resolve), typed `RenderPass`/`ComputePass`/
  `CopyPass`, `GraphicsPipeline`/`ComputePipeline` PSOs, `Buffer`/`Texture`/`Sampler`/`Shader`, and
  `drawIndexedPrimitivesIndirect` / `dispatchComputeIndirect` for the future GPU-driven work.
- **ObjectivelyMVC is already ported onto ObjectivelyGPU.** Its `Renderer` is now a concrete class that
  records through a `CommandBuffer`/`Framebuffer`; `WindowController` uses `initWithDevice(RenderDevice*)`
  and `render(commands, framebuffer)`. This is precisely the API that replaces Quetoo's `QuetooRenderer`.

So #864 becomes an **execution plan for Quetoo itself**: acquire a `RenderDevice` in `R_InitDevice`
(today's `R_InitContext`), render everything through ObjectivelyGPU, delete `QuetooRenderer` + old MVC, and
remove OpenGL.

## Decisions (locked)

| # | Decision | Choice | Consequence |
|---|----------|--------|-------------|
| 1 | Shader authoring | **Keep GLSL** (HLSL only if we hit a wall) | Offline GLSL→SPIR-V→MSL/DXIL asset step; reuse ObjectivelyGPU's toolchain and existing shaders |
| 2 | Occlusion culling | **Defer (in-branch)** | Stub `r_occlude`; accept a temporary fps regression; build the compute Hi-Z replacement **in this same branch** once the game is drawing again |
[...]

PR #904, opened 2026-07-02
#864 Port Quetoo to ObjectivelyGPU (SDL_gpu) to support modern rendering APIs

jdolan:
See #864


src/client/renderer/shaders/Makefile.am
```diff
@@ -29,3 +29,35 @@ shaders_DATA = \
...
   voxel.glsl
+
+# --- SDL_gpu shader transpilation -------------------------------------------
+#
+# Entry-point shaders ported to SDL_gpu. The compiled blobs (.spv for Vulkan,
+# .metal for Metal) are versioned in the repository, so normal builds never
+# invoke glslc or shadercross. Run 'make shaders' after editing a .glsl to
+# regenerate them. Extend these lists as more shaders are ported (issue #864).
+
+GPU_VERT_SHADERS = depth_pass_vs draw_2d_vs draw_3d_vs post_vs bsp_vs mesh_vs shadow_vs
```
jdolan: VERTEX_SHADERS and FRAGMENT_SHADERS would be nicer names for these source sets.

src/client/renderer/shaders/uniforms.glsl
```diff
@@ -19,6 +19,63 @@
...
+ *   Stage     | Sampled textures / storage | Uniform buffers
+ *   ----------+----------------------------+----------------
+ *   vertex    | set 0                      | set 1
+ *   fragment  | set 2                      | set 3
+ *
+ * Within a set, bindings are assigned in declaration order. We use a fixed,
+ * named binding map (below) so the C side binds one consistent table; a program
+ * declares only the resources it uses and shadercross preserves the indices.
+ */
+#if defined(VERTEX_SHADER)
```
jdolan, line 42: Why can't we use the same set indexes globally? Didn't we determine that SDL_gpu will retain sparse bindings during transpilation?

src/client/renderer/shaders/uniforms.glsl
```diff
@@ -155,25 +212,41 @@ layout (std140) uniform uniforms_block {
...
+  /**
+   * @brief The number of light sources in the lights storage buffer.
+   */
+  int num_lights;
+
+  /**
+   * @brief The number of static BSP lights (leading entries, lit via voxels).
+   * Lights at [num_bsp_lights, num_lights) are dynamic and lit directly.
+   */
+  int num_bsp_lights;
```
jdolan: Not sure what the benefit of moving these here was. They lived in the lights uniform block before, which I think was better encapsulation.

jdolan: Oh, because the lights UBO is now a storage buffer, not a uniform buffer, is that why? Can we not access storage buffers as easily or something?

src/client/renderer/r_bsp_draw.c
```diff
@@ -22,665 +22,230 @@
...
-  glUniform1i(r_bsp_program.stage.flags, STAGE_NONE);
-
-  glActiveTexture(GL_TEXTURE0 + TEXTURE_MATERIAL);
-
-  if (draw->surface & SURF_MASK_BLEND) {
-    glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
-  } else {
-    glDisable(GL_BLEND);
-  }
+void R_InitBspProgram(void) {
```
jdolan: We should probably rename all of these "Program" setup/teardown functions to e.g. R_InitBspPipeline / R_ShutdownBspPipeline yea? Also, let's introduce something like:

```c
static struct {
  Sampler *diffusemap_sampler;
  Sampler *voxel_data_sampler;
  GraphicsPipeline *pipeline;
} r_bsp_pipeline;
```

That will provide better encapsulation and reduce the number of globals, and it's akin to what we had before in the OpenGL code path with `r_bsp_program_t`. Do this for r_mesh_draw.c, r_sprite.c, r_sky.c, etc..

src/client/renderer/r_bsp_draw.c
```diff
@@ -22,665 +22,230 @@
...
-        R_Draw3DLines(GL_LINES, line_points, 2, color, true);
-      }
-    }
-  }
-}
-
-/**
- * @brief Draws a single material stage pass for a BSP draw elements batch.
- */
-static void R_DrawBspDrawElementsMaterialStage(const r_view_t *view,
```
jdolan, line 179: Wow, we still have a lot to re-enable.

src/client/renderer/r_context.c
```diff
@@ -0,0 +1,215 @@
...
+
+r_device_t r_device;
+
+/**
+ * @brief Objectively `ResourceProvider` bridging the GPU library's Resource
+ * lookups (e.g. `loadShader`) to Quetoo's virtual filesystem. This lets
+ * `RenderDevice::loadShader` resolve compiled shader blobs (`shaders/*.metal`,
+ * `shaders/*.spv`) via `Fs_Load` and the game's search paths.
+ */
+static Data *R_ResourceProvider(const char *name) {
```
jdolan, line 32: I think this completely negates the need for Ui_Data, the ResourceProvider registered for Quetoo's MVC assets.

Which really begs: we should just kill off ui_data.c entirely. It'll be precisely 3 lines long if we remove Ui_Data.

src/client/renderer/r_device.c
```diff
@@ -0,0 +1,215 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+ */
+
+#include "r_local.h"
+
+#include <Objectively/Resource.h>
+
+r_device_t r_device;
```
jdolan: I should not have renamed this file/type. r_context.c and r_context_t, with r_context as the instance, are all still perfectly good names. r_context.device mirrors s_context.device too.

src/client/renderer/r_device.c
```diff
@@ -0,0 +1,215 @@
...
+
+  SDL_SetWindowIcon(r_device.window, surf);
+
+  SDL_DestroySurface(surf);
+}
+
+/**
+ * @brief Updates the renderer device to reflect the current SDL window state and size.
+ */
+void R_UpdateDevice(void) {
```
jdolan: Back to R_UpdateContext

src/client/renderer/r_device.c
```diff
@@ -0,0 +1,215 @@
...
+  Com_Verbose("   GPU driver: %s\n", SDL_GetGPUDeviceDriver(r_device.device->device));
+
+  // Bridge the GPU library's Resource lookups (loadShader, ...) to Quetoo's filesystem.
+  $$(Resource, addResourceProvider, R_ResourceProvider);
+
+  R_UpdateDevice();
+
+  // The present-target framebuffer. The UI (and, later, the resolved 3D scene)
+  // composite into this; RenderDevice::endFrame blits it to the swapchain.
+  const SDL_GPUTextureFormat format = $(r_device.device, getSwapchainTextureFormat);
```
jdolan: Comment is stale, also we may want to implement separate framebuffers for 2D and 3D to correctly restore r_framebuffer_scale functionality, which should only apply to the 3D / view framebuffer.

src/client/renderer/r_draw_2d.c
```diff
@@ -103,26 +99,18 @@ static struct {
...
+  Sampler *sampler;
 
-  GLint projection2D;
+  // the dynamic vertex buffer and its capacity, in vertexes
+  Buffer *vertex_buffer;
+  uint32_t vertex_buffer_capacity;
 
-  GLint texture_diffusemap;
-} r_draw_2d_program;
+} r_draw_2d;
```
jdolan, line 106: Is there an opportunity for this to use MVC's GraphicsPipeline?

jdolan: Or even.. Hm. Move all 2D drawing to MVC? Implement the console and HUD as MVC Views? That should be a separate ticket.

src/client/renderer/r_draw_2d.c
```diff
@@ -459,64 +447,36 @@ void R_Draw2DImage(GLint x, GLint y, GLint w, GLint h, const r_image_t *image, c
...
     .num_vertexes = 6
   };
 
   r_draw_2d_vertex_t quad[4];
 
-  quad[0].position = Vec2s(x, y);
-  quad[1].position = Vec2s(x + w, y);
-  quad[2].position = Vec2s(x + w, y + h);
-  quad[3].position = Vec2s(x, y + h);
+  quad[0].position = Vec2(x, y);
```
jdolan, line 483: Not sure why we moved to floating point positions for 2D pixels?

src/client/renderer/r_draw_2d.c
```diff
@@ -570,6 +531,51 @@ void R_Draw2DLines(const GLint *points, size_t count, const color_t color) {
...
+  const GraphicsPipeline *bound = NULL;
+
+  const r_draw_2d_arrays_t *d = list->draw_arrays;
+  for (int32_t i = 0; i < list->num_draw_arrays; i++, d++) {
+
+    if (!d->texture) {
+      continue;
+    }
+
+    GraphicsPipeline *pipeline = d->mode == SDL_GPU_PRIMITIVETYPE_LINESTRIP
```
jdolan: Like we did in MVC, we should rasterize everything as triangles. So for lines / line strips, we still emit triangle verts to the accumulation buffers.

jdolan: That way we collapse this down to a single pipeline and never need to bind per-iteration.

src/client/renderer/r_draw_2d.c
```diff
@@ -581,88 +587,75 @@ void R_Draw2D(void) {
...
-  glViewport(viewport.x, viewport.y, viewport.w, viewport.h);
-
-  glEnable(GL_BLEND);
-  glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
-
-  glUseProgram(r_draw_2d_program.name);
-  glBindVertexArray(r_draw_2d.vertex_array);
+  CommandBuffer *commands = r_device.device->commandBuffer;
+  if (!commands || !r_draw_2d.pipeline_triangles) {
+    goto reset;
```
jdolan: Wtf absolutely not 😂

src/client/renderer/r_draw_2d.c
```diff
@@ -581,88 +587,75 @@ void R_Draw2D(void) {
...
-      glBindTexture(GL_TEXTURE_2D, d->texture);
-      glDrawArrays(d->mode, d->first_vertex + r_draw_2d.game.num_vertexes, d->num_vertexes);
+    CopyPass *copyPass = $(commands, beginCopyPass);
+
+    if (game_count) {
+      $(copyPass, uploadData, r_draw_2d.vertex_buffer->buffer, r_draw_2d.game.vertexes,
+        game_count * sizeof(r_draw_2d_vertex_t), 0, true);
+    }
+    if (ui_count) {
+      $(copyPass, uploadData, r_draw_2d.vertex_buffer->buffer, r_draw_2d.ui.vertexes,
```
jdolan: I'm confused. Are we drawing MVC twice here?

In main, QuetooRenderer emits MVC draw calls through r_draw_2d because MVC 1.x had a pluggable backend.

MVC 2.0 uses ObjectivelyGPU for its backend, and has its own pipeline, manages its own command flushing, etc. So I don't think this r_draw_2d.ui struct makes any sense at all anymore. We can kill it and consolidate down to a single vertex array object like we have in main. The only difference is, now it will only contain console and HUD geometry. All MVC geometry is completely external to r_draw_2d now.

src/client/renderer/r_draw_2d.c
```diff
@@ -581,88 +587,75 @@ void R_Draw2D(void) {
...
-  glUseProgram(0);
+  $(pass, bindVertexBuffers, 0, &(SDL_GPUBufferBinding) { .buffer = r_draw_2d.vertex_buffer->buffer }, 1);
 
+  // Game elements (console, HUD) use r_device pixel coordinates.
+  const mat4_t game_projection = Mat4_FromOrtho(0.f, r_device.w, r_device.h, 0.f, -1.f, 1.f);
+  R_Draw2DList(pass, &r_draw_2d.game, game_projection, 0);
+
+  // UI elements (menus) use raw window coordinates.
+  const mat4_t ui_projection = Mat4_FromOrtho(0.f, r_device.window_bounds.w, r_device.window_bounds.h, 0.f, -1.f, 1.f);
+  R_Draw2DList(pass, &r_draw_2d.ui, ui_projection, (int32_t) game_count);
```
jdolan: Like I said above, I think this goes away, and can probably just be inlined here.

src/client/renderer/r_draw_2d.c
```diff
@@ -581,88 +587,75 @@ void R_Draw2D(void) {
...
  * @brief Initializes the specified bitmap font. The layout of the font is square,
- * 2^n (e.g. 256x256, 512x512), and 8 rows by 16 columns. See below:
- *
- *
- *  !"#$%&'()*+,-./
- * 0123456789:;<=>?
- * @ABCDEFGHIJKLMNO
- * `PQRSTUVWXYZ`[\]^_
- * 'abcdefghijklmno
- * pqrstuvwxyz{|}"
```
jdolan, line 665: Do not lose this comment. The character map is helpful when debugging.

src/client/renderer/r_entity.c
```diff
@@ -154,15 +154,11 @@ void R_DrawEntities(const r_view_t *view) {
 
   thread_t *decals = Thread_Create((ThreadRunFunc) R_UpdateDecals, (void *) view, THREAD_NONE);
 
-  R_DrawOpaqueBspEntities(view);
```
jdolan, line 157: Hm. Why is this control flow changed?

src/client/renderer/r_framebuffer.c
```diff
@@ -8,10 +8,6 @@
  * as published by the Free Software Foundation; either version 2
  * of the License, or (at your option) any later version.
  *
- * This program is distributed in the hope that it will be useful,
```
jdolan, line 11: Why are you messing with copyright boilerplate?

src/client/renderer/r_image.c
```diff
@@ -164,66 +146,8 @@ void R_Screenshot_f(void) {
...
-      glTexStorage3D(image->target, image->levels, image->internal_format, image->width, image->height, image->depth);
-    } else {
-      glTexStorage2D(image->target, image->levels, image->internal_format, image->width, image->height);
-    }
-  }
-
-  R_RegisterMedia((r_media_t *) image);
-
-  R_GetError(image->media.name);
+  // TODO(#864): retired; GPU textures are created in R_LoadImage via ObjectivelyGPU.
```
jdolan: Yes but is all of this functionality restored in the new R_LoadImage or are there params / ops here not yet in this branch?

src/client/renderer/r_image.c
```diff
@@ -419,116 +331,6 @@ r_image_t *R_LoadImage(const char *name, r_image_type_t type) {
  * @brief Dump the image to the specified output file.
  */
 static void R_DumpImage(const r_image_t *image, const char *output, bool mipmap, bool raw) {
-
-  if (image->format != GL_RGB && image->format != GL_RGBA) {
```
jdolan, line 423: This is a debugging tool. I'm not 100% attached to it; it's helpful when debugging the atlas packer (sprites, shadowmaps, etc).

src/client/renderer/r_light.c
```diff
@@ -39,37 +39,29 @@ void R_AddLight(r_view_t *view, const r_light_t *l) {
...
+ * the lights storage buffer.
+ * @remarks TODO(#864): occlusion culling (l->query, r_draw_light_bounds) is
+ * deferred; every light is uploaded and assigned a shadow atlas tile.
  */
 void R_UpdateLights(r_view_t *view) {
 
   r_light_uniform_block_t *block = &r_lights.block;
   memset(block, 0, sizeof(*block));
 
-  cm_trace_t tr = { 0 };
```
jdolan, line 49: Not sure why this wasn't left alone. It's useful for inspecting light bounds in realtime to make sure our voxel inclusion is optimal.

src/client/renderer/r_light.c
```diff
@@ -39,37 +39,29 @@ void R_AddLight(r_view_t *view, const r_light_t *l) {
...
     r_light_uniform_t *out = &block->lights[i];
     out->origin = Vec3_ToVec4(l->origin, l->radius);
     out->color = Vec3_ToVec4(l->color, l->intensity);
-    out->shadow = Vec4(0.f, 0.f, 0.f, 0.f);
-
-    if (l->query) {
-      l->occluded = l->query->result == 0;
-    } else {
-      l->occluded = R_CulludeBox(view, l->bounds);
-    }
```
jdolan, line 68: This is probably the single largest perf drop in this branch. We need to get the depth pyramid with BSP blocks working before we can merge this down. it allows us to dismiss entire chunks of the map _and entire light sources_, plus `misc_dust` and other expensive objects.

src/client/renderer/r_light.c
```diff
@@ -78,78 +70,62 @@ void R_UpdateLights(r_view_t *view) {
...
-      }
+    const int32_t j = i - first;
+    if (j >= MAX_DYNAMIC_LIGHTS) {
+      break;
     }
-  }
 
-  if (len < MAX_DYNAMIC_LIGHTS) {
-    active_lights[len++] = -1;
+    if (!l->bsp_light && Box3_Intersects(l->bounds, bounds)) {
```
jdolan: `l->bsp_light` is always going to be false here. There's no point checking it conditionally. We could add an _assert_ before this line perhaps..

src/client/renderer/r_local.h
```diff
@@ -24,5 +24,7 @@
 #define __R_LOCAL_H__
 
 #include "renderer.h"
-#include "r_gl.h"
 #include <Objectively/Vector.h>
+
+// The shared GLSL/C descriptor binding map (pure #defines).
+#include "shaders/bindings.glsl"
```
jdolan: This feels wrong. These languages happen to use a similar enough syntax for this to work. But we should just repeat the bindings here, perhaps as an enum instead of #defines.

src/client/renderer/r_main.c
```diff
@@ -72,44 +72,11 @@ cvar_t *r_window_width;
 cvar_t *r_draw_stats;
 
 /**
- * @brief Queries OpenGL for any errors and prints them as warnings.
+ * @brief No-op retained for source compatibility with the many `R_GetError()`
+ * call sites; ObjectivelyGPU / SDL_gpu report errors directly.
+ * @remarks TODO(#864): remove the R_GetError call sites and this function.
  */
 void R_GetError_(const char *function, const char *msg) {
-
```
jdolan, line 78: Kill it.

src/client/renderer/r_main.c
```diff
@@ -152,6 +119,8 @@ void R_UpdateUniforms(const r_view_t *view) {
     out->editor = editor->integer;
     out->developer = developer->integer;
     out->wireframe = r_draw_wireframe->integer;
+    out->num_lights = view->num_lights;
+    out->num_bsp_lights = r_models.world ? r_models.world->bsp->num_lights : 0;
```
jdolan: Again, I don't see why these live here when we have a separate storage buffer for the lights themselves. These counts should live with their data.

src/client/renderer/r_main.c
```diff
@@ -262,82 +212,39 @@ void R_DrawViewDepth(r_view_t *view) {
...
-    R_ResolveFramebufferDepth(view->framebuffer);
-    glBindFramebuffer(GL_FRAMEBUFFER, render_fbo);
-    glDrawBuffers(1, (const GLenum []) { GL_COLOR_ATTACHMENT0 });
-  }
+  R_DrawMeshEntities(view);
 
   R_DrawSprites(view);
-
-  if (r_draw_wireframe->integer) {
-    glPolygonMode(GL_FRONT_AND_BACK, GL_FILL);
```
jdolan, line 299: I saw a note about the Vulkan alternative for glPolygonMode is not well supported on Android. That's fine. r_draw_wireframe is a debugging tool for mappers to visualize how quemap cut up their brushes. We don't have to worry about Android or iOS support for this particular feature. Best effort is fine.

src/client/renderer/r_main.h
```diff
@@ -245,6 +245,18 @@ typedef struct {
...
+     * @brief The number of light sources in the lights storage buffer.
+     */
+    int num_lights;
+
+    /**
+     * @brief The number of static BSP lights, which occupy the leading entries
+     * of the lights storage buffer and are lit via the voxel clusters. Lights at
+     * [num_bsp_lights, num_lights) are dynamic and lit directly.
+     */
+    int num_bsp_lights;
```
jdolan: Just noting again that I don't think these belong here. They belong in the storage buffer where the actual lights live. Unless there's some requirement that storage buffers be homogenous with a set stride or something.

src/client/renderer/r_main.h
```diff
@@ -93,22 +93,22 @@ typedef struct {
...
    */
-  GLint max_3d_texture_size;
+  int32_t max_3d_texture_size;
 
   /**
    * @brief The maximum uniform block size in bytes.
    */
-  GLint max_uniform_block_size;
+  int32_t max_uniform_block_size;
 } r_config_t;
```
jdolan, line 115: Do we even use r_config_t anymore?

src/client/renderer/r_sky.c
```diff
@@ -40,156 +29,70 @@ static struct {
...
+  // texture_sky sits at the fixed global slot SLOT_SAMPLER_SKY. SDL_gpu requires
+  // every slot in [0, num_samplers) to be bound (it validates presence, not just
+  // the slots the shader samples), so fill the leading unused slots with the same
+  // binding; only slot SLOT_SAMPLER_SKY is actually sampled.
+  SDL_GPUTextureSamplerBinding sky[SLOT_SAMPLER_SKY + 1];
+  for (int32_t i = 0; i <= SLOT_SAMPLER_SKY; i++) {
+    sky[i] = (SDL_GPUTextureSamplerBinding) {
+      .texture = r_sky.image->texture->texture,
+      .sampler = r_sky_sampler->sampler,
+    };
```
jdolan: Why doesn't every pipeline call the same helper function to bind all Samplers to their real enum-based slots, universally? Rather than crutching each of them up with a silly loop like this, just have a helper in e.g.  r_pipeline.c. Or, if we wanted to go crazy, we could even subclass Pipeline as QuetooPipeline and give it this default behavior at the object level. We could encapsulate all of our binding tic-tac-toe in one place.

src/client/renderer/r_material.h
```diff
@@ -26,6 +26,24 @@
...
+ */
+typedef struct {
+  int32_t surface;
+  float alpha_test;
+  float roughness;
+  float hardness;
+  float specularity;
+  float parallax;
+  float shadow;
+  float _pad;
```
jdolan: We should be using alignas(16) on all uniform types rather than relying on _pad.

jdolan:
## Review reconciliation (2026-07-03)

Resolving the completed review threads. Mapping to the commits that addressed them:

- **`Program`→`Pipeline` rename + pipeline structs** (bsp/mesh/sprite/sky) — *Encapsulate … pipeline globals*
- **Kill `Ui_Data`** — redundant with `R_ResourceProvider` — *Remove redundant Ui_Data resource provider*
- **Revert `r_device`→`r_context`** (file/type/instance + `R_UpdateContext`) — *Revert r_device rename back to r_context*
- **Stale render-scale comment** — *Fix stale render-scale comment in R_InitContext*
- **Light counts belong with their data** (`num_lights`/`num_bsp_lights` → lights storage buffer; the storage-vs-UBO question) — *Move light counts into the lights storage buffer*
- **Descriptor bindings as a C enum, not a shared `#include`** — *Give the C renderer its own descriptor slot enum*
- **Kill `r_draw_2d.ui` / "drawing MVC twice"** — *Remove the dead r_draw_2d UI list*
- **Rasterize 2D lines as triangles → single 2D pipeline** — *Rasterize 2D lines as triangles; single 2D pipeline*
- **Preserve the character-map comment** — *Restore the font character-map comment*
- **Copyright boilerplate** (`r_framebuffer.c`) — restored to match `main`
- **`l->bsp_light` always false → assert** — assert is in place
- **`r_draw_wireframe` (best-effort)** — *Restore r_draw_wireframe for BSP geometry*
- **"Lot to re-enable"** — since re-enabled: soft particles, translucent BSP surfaces, mesh shadow casters, inline BSP-model entities, full per-pixel mesh material lighting (normal/specular/parallax/shadows), and decals.

### Left open intentionally (remaining work / your call)
- **`r_device.c:185` dedicated 3D-view framebuffer** so `r_framebuffer_scale` scales only the scene — the stale comment is fixed; the FB split itself is still TODO (kept open).
- **`r_light.c:68` occlusion perf** — Hi-Z was retired; scene-FB early-Z landed, but broad-phase occlusion remains the known perf gap.
- **Questions not yet actioned:** global set indexes (`uniforms.glsl:40`), float vs int 2D positions (`r_draw_2d.c`), `R_LoadImage` parity + atlas-dump / light-bounds debug tools, `r_entity.c:157` control-flow, `r_main.c:78`, `r_main.h:112` (`r_config_t`), `r_sky.c:95` shared sampler-bind helper, `r_material.h:46` `alignas(16)`, and "move all 2D to MVC" (separate ticket).

src/client/renderer/shaders/Makefile.am
```diff
@@ -23,9 +23,51 @@ shaders_DATA = \
...
+# .metal for Metal) are versioned in the repository, so normal builds never
+# invoke glslc or shadercross. Run 'make shaders' after editing a .glsl to
+# regenerate them. Extend these lists as more shaders are ported (issue #864).
+
+VERTEX_SHADERS = depth_pass_vs draw_2d_vs draw_3d_vs post_vs bsp_vs mesh_vs shadow_vs sky_vs sprite_vs decal_vs
+FRAGMENT_SHADERS = depth_pass_fs draw_2d_fs draw_3d_fs post_fs bsp_fs mesh_fs shadow_fs sky_fs sprite_fs decal_fs
+
+# Second, MATERIAL_STAGES variant of the bsp program, used only for material stage
+# draws (bsp_vs.glsl / bsp_fs.glsl compiled with -DMATERIAL_STAGES).
+STAGE_SHADERS = bsp_stage_vs bsp_stage_fs
```
jdolan: Don't we need to do this for mesh as well? Make a pipeline variant for the material stages (pulse, shell, etc..) for mesh models?

src/client/renderer/r_bsp_draw.c
```diff
@@ -22,665 +22,677 @@
...
-    GLint warp;
-    GLint lighting;
-    GLint emissive;
-    GLint lerp;
-  } stage;
-
-  r_image_t *warp_image;
-} r_bsp_program;
-
-static float R_StageDriftHash(const void *a, const void *b) {
```
jdolan, line 84: Looks like we lost stage.drift in the shuffle. This feature adds randomization to stages so that repeated instances of the same material don't all animate in lock-step (think: computer screens, flames, things that should not all look identical frame-to-frame).

src/client/renderer/r_bsp_draw.c
```diff
@@ -22,665 +22,677 @@
...
+    .address_mode_v = SDL_GPU_SAMPLERADDRESSMODE_REPEAT,
+    .address_mode_w = SDL_GPU_SAMPLERADDRESSMODE_REPEAT,
+  });
 }
 
 /**
- * @brief Debug visualization for voxels and lights
+ * @brief Releases the BSP pipeline and samplers.
  */
-static void R_DrawBspVoxels(const r_view_t *view, const r_bsp_model_t *bsp) {
```
jdolan, line 130: I don't see why these debug tools cant' come back. They only rely on r_draw_3d.c.

src/client/renderer/r_framebuffer.c
```diff
@@ -21,343 +21,43 @@
 
 #include "r_local.h"
```
jdolan: r_framebuffer.c needs to be deleted. It's basically a wrapper around a single call, $(device, createFramebuffer, ..), and it doesn't even honor the ATTACHMENTS mask anymore.

src/client/renderer/r_light.c
jdolan: The Hi-Z pyramid did not work well with CPU readback (attempting to reproduce OpenGL occlusion queries 1:1). So r_occlude.c has been deleted entirely. I guess we close this one out for now..

src/client/renderer/shaders/bindings.glsl
```diff
@@ -0,0 +1,160 @@
...
+ * the two voxel volumes (caustics, occlusion) and the sky cubemap for image-based
+ * ambient at [3, 5]. The latter three are always part of the lit family; the lit
+ * fragment shaders define VOXEL_CAUSTICS_OCCLUSION and LIGHT_SKY (see bsp_fs).
+ */
+#define BINDING_SAMPLER_MATERIAL            0
+#define BINDING_SAMPLER_VOXEL_LIGHT_DATA    1
+#define BINDING_SAMPLER_SHADOW_ATLAS        2
+#define BINDING_SAMPLER_VOXEL_CAUSTICS      3
+#define BINDING_SAMPLER_VOXEL_OCCLUSION     4
+#define BINDING_SAMPLER_SKY_AMBIENT         5
```
jdolan: This is _better_ than it was earlier in this branch, but still gross.

Can we avoid #if defined (THIS_AND_THAT) entirely, and just have a single contiguous set of constants for ALL bindings supported in the game? There are not that many. The Pipelines can all share them. If they must bind a texture or two that they do not use, that is okay.

Either that, or we break this into per-pipeline binding constants in e.g. bsp_bindings.glsl, mesh_bindings.glsl, sky_bindings.glsl, ... And create corresponding enums on the C-side in each pipline source file (r_bsp_draw, r_mesh_draw). 

But this current game of Plinko "if this is on, then these constants mean _this_ instead of _that_" needs to go.

src/client/renderer/shaders/bsp_fs.glsl
```diff
@@ -19,11 +19,25 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
  */
 
-uniform mat4 model;
+#version 450
 
-in common_vertex_t vertex;
+#define UNIFORMS_NO_SAMPLERS
```
jdolan: Carving up the shaders with these #ifdef's is an anti-pattern and needs to go away. What are we trying to do here? Setup bindings? Eliminate dead code not used by this shader? There are better ways.

src/client/renderer/r_main.h
jdolan: Yes — checked again, it is used: `r_atlas.c:158` and `r_shadow.c:349` both read `r_config.max_texture_size` to clamp atlas/shadow-atlas dimensions. The struct + hardcoded conservative values (SDL_gpu has no device-limits query API) stay as-is.

src/client/renderer/r_entity.c
jdolan: Two things got tangled here: (1) sky is now drawn after opaque BSP instead of before (intentional — it depth-tests against the already-drawn world so it only fills unoccluded texels, cutting overdraw vs the GL renderer's sky-first approach; added a comment explaining this at the call site), and (2) the decal-update thread got moved to start after the BSP+sky draws instead of before, which lost the CPU/GPU overlap main gets by kicking it off first (R_UpdateDecals is pure CPU BSP-node clipping, no GPU calls, so it races nothing). Fixed in `52b0307a8`: thread creation moved back to the top, matching main, sky placement kept with rationale documented inline.

src/client/renderer/r_material.h
jdolan: Done in `a4b8843bc` — r_material_uniforms_t, r_stage_uniforms_t, and r_light_uniform_block_t all now use `alignas(16)` (on the first scalar member, or directly on the `lights` array for the block that needs an offset guarantee, not just a size one) instead of hand-counted _pad fields. Verified sizeof/offsetof are byte-identical to before.

src/client/renderer/r_device.c
jdolan: Done in `3b48b3c93` — r_framebuffer.c is gone, and the separate-2D/3D-framebuffer split it was asking about already exists: the present framebuffer created here (native res, swapchain format, UI/2D) is distinct from the scaled 3D scene framebuffer (now R_CreateFramebuffer in this file, HDR + depth-copy + depth, scaled by r_framebuffer_scale). Fixed the stale comment to describe that instead of asking for it.

src/client/renderer/r_framebuffer.c
jdolan: Done in `3b48b3c93` — deleted r_framebuffer.c/.h; R_CreateFramebuffer/R_DestroyFramebuffer moved into r_context.c (which already owns the device and the present framebuffer). Left the attachments-mask behavior as-is (still advisory/ignored) since fixing that is a separate, larger change tracked on #864, not part of this cleanup.

src/client/renderer/r_light.c
jdolan: Restored in `de4dbd994`, ported straight from main: trace along view forward, draw R_Draw3DBox for lights within 64 units of the trace endpoint.

src/client/renderer/r_image.c
jdolan: Audited (background agent, cross-checked against main line-by-line): wrap modes, per-type filtering (nearest/linear), anisotropy, and format selection are all correctly restored -- just relocated from per-texture GL state onto per-pipeline Samplers (r_bsp_draw.c, r_mesh_draw.c, r_sprite.c), which is the right home now that SDL_gpu separates textures from filtering state. One real gap found: mip generation for the generic (non-cubemap) R_LoadImage path is missing -- textures are created with num_levels=1 and never call generateMipmaps, unlike main which called glGenerateMipmap for every image with levels>1 including IMG_MATERIAL. Since materials route through R_LoadImage directly in this branch (they didn't in main), this affects material textures. Filed as its own tracked item on #864 rather than fixed here, since it's a feature gap, not a cleanup-pass item.

src/client/renderer/r_sky.c
jdolan: Root-caused instead of abstracted: the only pipeline needing a fill loop was sky, and only because SLOT_SAMPLER_SKY was pinned to a fixed high slot (6) to match the lit family's sampler count even though sky shares no bindings with that family at all. Moved it to slot 0 (`e88eb5152`), its own dense family of one, matching every other pipeline's convention. The fill-loop is gone entirely -- no shared helper needed since there's no remaining boilerplate to share.

src/client/renderer/r_draw_2d.c
jdolan: main used vec2s_t (GLshort) bound via glVertexAttribPointer with an integer type + normalized=false, which GL auto-widens to float during vertex fetch. SDL_gpu has no equivalent: its non-_NORM integer vertex formats (SHORT2, INT2, ...) are fetched as true integer attributes -- the vertex shader input would need to be ivec2, with an explicit int-to-float cast in the shader -- and the _NORM variants divide by 32767, which is wrong for pixel coordinates that can exceed that range at high resolutions anyway. So there's no direct equivalent to GL's unnormalized-short-to-float fetch; float positions (vec2_t) are the pragmatic match, and the bandwidth difference (4 bytes/vertex) is negligible for 2D UI geometry.

src/client/renderer/r_main.h
jdolan: Re-reading this with fresh eyes: the descriptor SET indexes (0/1 vertex, 2/3 fragment, per SAMPLER_SET/UNIFORM_SET above) already ARE global and fixed across every pipeline -- that part was never actually conditional. What DOES vary per-pipeline-variant is the BINDING number within a set (bindings.glsl's #if defined(MATERIAL_STAGES) blocks), because storage buffers must sit immediately after the last sampler (confirmed empirically: Metal packs storage-buffer [[buffer]] indices right after the samplers, and a gap desyncs the C-side slot from the GLSL binding -- this bit us once, see the bindings.glsl file comment). So sparse SAMPLER bindings do transpile fine (confirmed by the sky pipeline sitting alone at binding 0 while lit pipelines use 0-5); it's specifically the sampler-count-dependent storage-buffer offset that forces the per-variant conditionals. Untangling that fully (single contiguous global map, or split into bsp_bindings.glsl/mesh_bindings.glsl) is a bigger shader-architecture change than this cleanup pass -- tracked as a standalone item on #864, not resolved here.

src/client/renderer/shaders/Makefile.am
jdolan: Turned out to be the wrong question in retrospect -- mesh doesnt need a pipeline variant for stages at all anymore, and neither does bsp. Ported main's actual dispatch shape instead (`5a8236c636`): one shader stays bound for the whole draw sequence; a stage draw is just another draw call with a runtime `if (stage.flags == STAGE_NONE)` branch selecting behavior, exactly like main's bsp_fs/mesh_fs. bsp_stage_vs/fs and mesh_stage_vs/fs are deleted; the only remaining per-variant pipeline cache is for SDL_gpu's baked-in blend state (unavoidable, orthogonal to the shader).

src/client/renderer/shaders/bindings.glsl
jdolan: Gone -- bindings.glsl deleted entirely (`a8236c636`). Every pipeline now declares its own local, dense-from-0 binding enum in its own .c file and matching #defines in its own .glsl; no shared file, no #if defined(...), no cross-pipeline name collisions. r_local.h keeps only the two constants truly shared by every pipeline (globals + locals UBO slots).

src/client/renderer/shaders/bsp_fs.glsl
jdolan: Agreed and fixed the same way you suggested independently -- turned out the #ifdef(MATERIAL_STAGES) branch didn't need to be compile-time at all. main uses a runtime `if (stage.flags == STAGE_NONE)` branch (one compiled shader, selected per-draw by a uniform) instead of two separate compiled variants; ported that shape directly (`a8236c636`). No more #ifdef carving in bsp_fs.glsl or mesh_fs.glsl.

src/client/renderer/shaders/bsp_fs.glsl
```diff
@@ -70,26 +115,20 @@ void parallax_occlusion_mapping(in common_vertex_t vertex, inout common_fragment
...
   if (fragment.view_dist >= lighting_distance) {
     fragment.ambient = vertex.ambient;
     fragment.diffuse = vertex.diffuse;
     fragment.specular = vec3(0.0);
     return;
   }
 
-  // For fragments within range, do full per-fragment lighting
-
-  if ((stage.flags & STAGE_LIGHTING_FLAT) == STAGE_LIGHTING_FLAT) {
```
jdolan, line 86: Looks like we lost STAGE_LIGHTING_FLAT? This is needed for non-emissive stages. They use vertex lighting (no normalmap) to look more natural.

src/client/renderer/shaders/bsp_fs.glsl
```diff
@@ -137,24 +177,22 @@ void main(void) {
 
     vec2 st = fragment.parallax;
 
-    if ((stage.flags & STAGE_WARP) == STAGE_WARP) {
-      st += (texture(texture_warp, st + vec2(ticks * stage.warp.x * 0.000125)).xy - 0.5) * stage.warp.y;
+    if ((material.flags & STAGE_WARP) == STAGE_WARP) {
```
jdolan, line 204: Interesting so we smooshed the materials and stage uniforms into the material struct. GLSL structs support nesting -- we could re-separate these a bit, cosmetically, so that it more closely resembles the C side:

```glsl
struct material_t {
  float hardness;
  float roughness;
  float specular;
  // ..
};

struct stage_t {
  int flags;
  vec4 color;
  float pulse;
  float drift;
  // .. 
};

// Make *this* the uniform block shape
struct material_uniform_t {
  material_t material;
  stage_t stage;
};

PR #908, opened 2026-07-23
feat(rtx): add native Vulkan RTX scene path

Eclipse1982:
## Summary
- add an optional native Vulkan RTX device path behind `BUILD_RTX`
- expose RTX/VK settings only when the local client reports compatible Vulkan ray-tracing support
- ray trace the loaded BSP scene into a headless Vulkan storage image and bridge it back through the existing SDL GPU renderer

## Verification
- Built on Windows x64 Debug with `Quetoo.vs15/quetoo_all.sln`
- Launched with `+set dedicated 0 +set net_port 27999 +set r_rtx 1 +set r_rtx_overlay 1 +set r_rtx_smoke 0 +map edge`
- Log showed `Native Vulkan RTX support: available`
- Log showed `RTX: built acceleration structures for 23497 triangles and 131 lights`
- Client remained responsive; no new crash dump was produced

## Notes
This is intentionally incremental: it keeps SDL GPU as the shipping renderer/presentation path and only enables RTX controls on compatible clients. The larger historical QuetooPlus Vulkan backend/Q2RTX work can be layered on top in follow-up PRs.


jdolan:
I have no way to test this, no way to play it, no way to even contribute to it.

I intentionally chose SDL_gpu for a lowest common denominator that still gets us on a bunch of platforms.

This PR takes the project in a different direction from that, where the game will look radically different for a subset of users.

This is probably something best achieved in a fork, honestly. If you want to make QuetooRTX, you should!

PR #911, opened 2026-07-29
Add func_bob: self-contained bobbing platform entity

skies912:
## Summary
- Adds `func_bob`, a self-contained bobbing-platform entity: no `path_corner`s needed (unlike `func_train`), so it's cheap to place in quantity, e.g. an array of small ambient platforms.
- Motion is a single continuous half-sine spanning the whole leg (zero at both ends, peak at the midpoint) rather than stitched accel/cruise/decel zones, so there's no interior seam or discretization drift. Supports `compression` (curve shaping), `wait`, `drift` (cycle-scaled startup phase stagger so identical entities don't move in lockstep), a looping `sound` while moving, and a `start_off`/`Use` toggle.
- Fixes a latent discretization bug in the shared `G_MoveInfo_Linear_Accelerate` ramp (used by `func_door`/`func_plat`/`func_button`/`func_train`): the non-clamped branch planned its cruise phase against the nominal speed instead of the realized peak speed reachable within the truncated `accel_frames`, understating travel distance and leaving a "crawl then snap" residual for `G_MoveInfo_Linear_Final` to absorb. Found while tuning `func_bob`'s original (later-abandoned) linear-ramp approach.
- Gives `func_bob` its own `MOD_BOB` obituary ("was crushed by Bob. Not the builder.") distinct from the generic crush message.

## Test plan
- [x] Builds clean (`game.vcxproj`)
- [x] Manually tested in-map: spawn, bobbing motion, `wait`/`drift` desync across an array, `start_off` + `Use` toggle, sound while moving, crush obituary
- [ ] FGD entry (`Quetoo.fgd`) updated separately to document the new entity for TrenchBroom

### DIFF
--- a/src/cgame/default/cg_entity.c
+++ b/src/cgame/default/cg_entity.c
@@ -45,6 +45,7 @@ static const cm_entity_t *Cg_FindEntity(const cm_entity_t *from, const Cg_Entity
       start++;
     }
     assert(start < bsp->num_entities);
+    start++;
   }
 
   for (int32_t i = start; i < bsp->num_entities; i++) {
--- a/src/cgame/default/cg_entity_misc.c
+++ b/src/cgame/default/cg_entity_misc.c
@@ -783,7 +783,7 @@ static void Cg_misc_sprite_Think(cg_entity_t *self) {
       }
 
       s.origin = Vec3_Mix(this->sprite.origin, that->sprite.origin, Randomf());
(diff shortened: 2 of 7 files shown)


jdolan (approved):
Woo! Nice! Don't forget to update quetoo-data for entities.def / TrenchBroom config.

skies912:
> Woo! Nice! Don't forget to update quetoo-data for entities.def / TrenchBroom config.

def and fgd updated

Issue #906, opened 2026-07-08
Changing game cvar does not switch game and cgame modules correctly


MovoWR:
## Description

Changing `game` in the console should switch Quetoo to a different game folder on the next map. For example, `game banana` should switch to the `banana` folder.

That switch has two parts:

1. The server should load `game.dll` from the new folder.
2. The client should load `cgame.dll` from the new folder.

Right now one or both parts can fail. The server may still load the old `game.dll`, and the client may still keep the old `cgame.dll`.

This can leave the game in a mixed state: server logic comes from one game folder, but the client HUD, UI, and prediction come from another.

## Steps to Reproduce

1. For testing, copy `lib/default` and rename the copy to `lib/banana`.
2. Launch Quetoo with the default game module.
3. Open the console and type: `game banana`.
4. Load a map: `map warehouse`.
5. Check the console output for which `game.dll` was loaded.
6. Check whether the log loads `game.dll` and `cgame.dll` from `lib/banana`.

## Expected Behavior

After changing `game` and loading a map:

1. The server should switch to the new game folder before it loads `game.dll`.
2. The server should load `lib/banana/game.dll`.
3. The client should see that the server is using `banana`.
4. The client should reload cgame and load `lib/banana/cgame.dll`.

## Actual Behavior

The server may load the old game module, `lib/default/game.dll`, because the new game folder is added too late.

The client may also keep the old cgame module, `lib/default/cgame.dll`. Console only shows `cgame.dll` loading once at startup from `lib/default`; after switching to `banana`, there is no later `lib/banana/cgame.dll` load line.

## Environment

- **OS**: Windows 11
- **Quetoo version**: 9966
- **Build type**: Release
## Logs / Output

```text
HTTP server listening on port 1998
Parsed 24 maps from maps.lst
...
Game initialization...
  Loading C:\best_game_ever\quetoo\lib/default/game.dll...
  Version:    x86_64-pc-windows 9966
Game module initialized
...
Client game initialization...
  Loading C:\best_game_ever\quetoo\lib/default/cgame.dll...
  Version:    9966 x86_64-pc-windows
Client game initialized
Quetoo 9966 initialized
...
map warehouse
Server initialization...
Game initialization...
  Loading C:\best_game_ever\quetoo\lib/default/game.dll...
  Version:    x86_64-pc-windows 9966
Game module initialized
Game initialized, starting...
  Loaded map warehouse, 99 entities.
Server initialized
...
game
game "default" (server, latched)
[...]

PR #915, opened 2026-07-31
Mod support: a framework for composing gameplay rules with hooks, and CTF and Lithium mods to prove it out

jdolan:
Implements #906.

The game module loading race is fixed, `ctf` becomes a game and cgame module of its own, and the code both modules inherit moves into `src/game/common` and `src/cgame/common`.

## Shape

A mod is its own gameplay, not a fork of the engine's furniture. The common directories hold what every module inherits; each module owns the rest.

**Common sources are compiled per module, never linked from one prebuilt library.** They resolve `g_types.h` / `cg_local.h` from the compiling module's directory, so a single prebuilt copy would bake one module's struct layout into all of them. The concrete hazard: `cg_state_t` embeds `cg_team_info_t teams[MAX_TEAMS]` plus `g_gameplay_t` and `g_items_t`, all from the per-module `g_types.h`, so any mod changing its team roster or item set would read `cg_state` at the wrong offsets.

**Mechanism is shared; content stays per module.** The tell is a gameplay-mode branch — `g_level.gameplay`, `g_level.teams`, `GAME_INSTAGIB`. Zero branches means mechanism. By that rule `g_ballistics.c` is shared despite being weapon code, while `bg_item.c`, `g_item.c`, `g_main.c`, `g_combat.c`, `g_weapon.c`, `g_client.c`, `g_entity.c`, `g_types.h` and the client's `cg_hud.c`, `cg_score.c`, `cg_team_mode.c` stay per module.

**A named per-module function beats an `#if` guard.** A guard teaches common code about every module that will ever exist. The seams are `G_ResetDroppedItem` (`src/game/common/g_module.h`) and `Cg_TeamModes` (`src/cgame/common/cg_team_mode.h`). Guards are reserved for optional features a module opts into wholesale: `G_HOOK` for the grapple, `G_CTF` for capture play.

**Where only presentation varies, share primitives rather than adding hooks.** `common/cg_hud_draw.c` holds the crosshair, blends, center print, weapon bar and icon/vital/powerup helpers; each module's `cg_hud.c` is composition only. Hooks were rejected here — roughly six would have been needed, and they freeze the layout so a module could append elements but never restructure.

A module can always override a common file by substituting its own in its source list. That escape hatch is what makes sharing safe.

## Build systems

All three are updated. Autotools and Xcode are build verified.

`Quetoo.vs15` needed repair as well as extension: the file moves had left 116 of `cgame.vcxproj`'s 134 sources and 35 of `game.vcxproj`'s 54 pointing at files that no longer existed. The common lists now live in `game_common.props` / `cgame_common.props`, imported by all four module projects, mirroring how the autotools builds share `sources.mk`. A module's own directory goes in its project's `AdditionalIncludeDirectories` rather than the global include path, where two modules' `g_types.h` and `cg_local.h` would be ambiguous.

## Verification

- Autotools: clean `make` from scratch, exit 0, no new warnings; all four modules link.
- Xcode: builds through `Quetoo.xcworkspace` (not `-project`, so the Objectively frameworks resolve from the sibling checkouts).
- **MSVS is not build verified** — it needs a Windows/MSBuild toolchain. What was checked mechanically: XML well-formedness, every source and project-reference path resolving on disk, GUID agreement across projects and both solutions, config coverage for every solution configuration, and each module's compiled source set matching autotools exactly.
- Runtime smoke (`quetoo-dedicated +set game <module> +map edge +quit`) is **not** re-run since the rebase; it needs an install, as modules load from `PKGLIBDIR`.

## Notes for review

- Rebased on `origin/main`, which brought in `func_bob`. The rename carried it into `src/game/common/g_entity_func.c`, but ctf's own `g_types.h`, `g_entity.c` and `g_client.c` predated it, so the last commit gives ctf its `MOD_BOB`, class registration and obituary. Common code referencing `MOD_BOB` made ctf fail to compile until then — the contract working as intended.
- Three things named "ctf" are not capture-play content: `DEFAULT_TEAM_SKIN` and the bot's `gork/ctf` skin are the team-coloured player skin all team play uses, and `cg_beam_hook` is a generic `sprites/rope` beam also used by the AI nav editor. `BUTTON_HOOK` likewise survives in `default` as that editor's delete-node input.


src/game/common/g_ctf.c
```diff
@@ -0,0 +1,481 @@
...
+    return;
+  }
+
+  super.ResetDroppedItem(ent);
+}
+
+/**
+ * @brief Resolves "flag" to whichever flag the client is carrying.
+ */
+static void G_DropInventoryItem_Ctf(g_client_t *cl, const char *name) {
```
jdolan: **Every flag is named `"Enemy Flag"`, so `drop flag` only works for the red flag carrier.**

`G_DropInventoryItem_Ctf` resolves the carried flag to `flag->def.name`, and all four flags in `src/game/ctf/bg_item.c` set `.name = "Enemy Flag"`. `G_DropInventoryItem_Common` then does `G_FindItem(name)`, which scans `WEAPON_FIRST..ITEM_TOTAL` ascending and returns the first available match — always `FLAG_RED`, since `G_ItemAvailable` is true for `ITEM_TYPE_FLAG`.

So a TEAM_RED player carrying the blue flag types `drop flag`, `cl->inventory[FLAG_RED]` is 0, and they get `"Out of item: Enemy Flag"` — the flag is never dropped. Before this PR `G_Drop_f` special-cased `"flag"` and called `G_TossFlag(cl)` on the carried flag directly, so this is a regression rather than a pre-existing quirk.

The name round-trip is the root cause: resolving an item to a string only to look the string back up cannot express "this particular flag". Suggest `DropInventoryItem(g_client_t *, const g_item_t *)`, with the name→item step staying in `G_Drop_f` where the user's string actually arrives. Giving each flag a unique `.name` would also work but changes the pickup messages players see.

src/game/ctf/g_types.h
```diff
@@ -0,0 +1,1720 @@
...
+ * @brief The name this module answers to: its directory under lib/quetoo, and
+ * the game_name it advertises.
+ */
+#define GAME_NAME "ctf"
+
+/**
+ * @brief Game protocol version (protocol minor version). To be incremented
+ * whenever the game protocol changes.
+ */
+#define PROTOCOL_MINOR 1044
```
jdolan, line 44: **All three modules advertise `PROTOCOL_MINOR 1044`, so the one field that could reject a module mismatch cannot.**

`default`, `ctf` and `lithium` each carry a different wire layout — ctf inserts `STAT_CAPTURES` and `STAT_TECH`, shifting every later stat index; lithium inserts a different subset and puts `MOD_HOOK` at a different ordinal again — yet all three set the same minor, which `src/client/cl_parse.c:298` compares against `cls.cgame->protocol`.

The reachable path: a client without the ctf module joins a ctf server. `Fs_SetGame("ctf")` mounts nothing new, `Sys_OpenLibrary` resolves `cgame.so` through the whole search path and falls back to `default/cgame.so`, and `cls.cgame_game` records the *requested* directory rather than the image's real origin. The minor check passes, and the default cgame reads a ctf server's `ps.stats[]` one index out — garbage health and ammo instead of a clean refusal.

Two fixes, and they compose: derive the minor per module (a base plus a feature bitmask would keep it mechanical), and have `Sys_OpenLibrary` take the required game directory and fail rather than silently resolving to `default`. The second is worth doing on its own — the silent fallback is what turns a missing mod into corruption rather than an error message.

src/client/cl_main.c
```diff
@@ -637,6 +637,10 @@ void Cl_Frame(const uint32_t msec) {
     return;
   }
 
+  if (game->modified && q_strcmp(cls.cgame_game, Fs_Game())) {
```
jdolan: **Changing `game` while connected to a remote server hot-swaps the cgame in place, with no media load and no protocol recheck.**

`game` is `CVAR_LATCH`, but `Cvar_Set_` only defers when `Com_WasInit(QUETOO_SERVER)` — i.e. only when a *local* server is running. Connected to a remote server with `cls.state == CL_ACTIVE`, `game lithium` therefore applies immediately and sets `modified`.

This line then calls `Cl_InitCgame()`, which shuts the old module down and inits the new one. `Cg_Init` does not load media — those come only from `cls.cgame->LoadMedia()` inside `Cl_LoadMedia`, which nothing calls here. Execution falls through to `Cl_UpdateScene()` in the same frame with every cgame media pointer NULL, and the client stays connected, decoding the server's stats and item indices with a different module's layout. The `minor != cls.cgame->protocol` guard is only reached on `server_data`, so this path is entirely unvalidated.

A bare `Cl_InitCgame()` mid-session is not sufficient. Either honour the reload only when `cls.state == CL_DISCONNECTED` and defer otherwise, or `Cl_Disconnect()` first so the normal connect path re-runs the protocol check and media load.

src/client/cl_media.c
```diff
@@ -31,6 +31,13 @@ static void Cl_CheckManifestEntry_(const HashTable *table, ident key, ident val,
   const cm_manifest_entry_t *entry = (const cm_manifest_entry_t *) val;
   if (Fs_Exists(entry->path)) {
     if (!Cm_CheckManifestEntry(entry)) {
+
+      // the bsp is the one asset a mismatch on is not cosmetic: the client
+      // predicts against its own collision model, so a different bsp desyncs
+      if (!q_strcmp(entry->path, cl.config_strings[CS_BSP])) {
+        Com_Error(ERROR_DROP, "%s differs from server (expected %s)\n", entry->path, entry->hash);
```
jdolan: **The new fatal BSP check also fires for the host of a listen server, where client and server read the same file.**

`Cl_RequestNextDownload` runs for the local client too, so this `Com_Error(ERROR_DROP)` is reachable when hosting. On `main` the equivalent check in `Cl_UpdatePrediction` was explicitly skipped for a listen server — `if (!Com_WasInit(QUETOO_SERVER) || cl.demo_server || !Cm_NumModels())` — so a locally inconsistent bsp/`.mf` pair was a warning.

Now: a host with a loose newer `~/.quetoo/default/maps/x.bsp` shadowing a pk3 that also supplies `maps/x.mf`, or any hand-repacked bsp, gets `Cm_CheckManifestEntry` failing, `Com_Error` → `Sv_ShutdownServer` + `Cl_Disconnect`. They cannot start their own map, even though there is no client/server divergence by construction — both sides are the same process reading the same file.

Suggest restoring the listen-server exemption (keep the warning) and reserving the drop for the case it was written for.

Separately, the `Com_Error` longjmps out of `$(manifest, enumerate, ...)`, so `Cm_FreeManifest(manifest)` below is never reached — worth recording the mismatch and erroring after the enumeration returns.

src/client/cl_predict.c
```diff
@@ -304,17 +304,10 @@ void Cl_UpdatePrediction(void) {
 
   // ensure the world model is loaded
   if (!Com_WasInit(QUETOO_SERVER) || cl.demo_server || !Cm_NumModels()) {
-    int64_t bs;
```
jdolan, line 307: **Dropping `CS_BSP_SIZE` removed the only server-authoritative BSP check; the manifest is local data, so a self-consistent stale map passes.**

`CS_BSP_SIZE` was computed by the server in `Sv_LoadMedia` and compared on the client, so it proved *server vs client*. `maps/x.mf` is read from the client's own filesystem, and `Cl_CheckOrDownloadFile` skips the download entirely when the file already exists.

So: client has `edge` v1 with a mutually consistent `edge.bsp` + `edge.mf`; server runs `edge` v2. Nothing downloads, the local hashes agree, no warning and no error — and the client predicts against a different collision model. The comment added just above, that a mismatch "is fatal in `Cl_CheckManifestEntry_`", holds only when the client's own bsp and manifest disagree, which is not the case this check existed to catch.

If `CS_BSP_SIZE` was worth removing (and a size is a weak fingerprint), a server-computed hash in a config string — `CS_BSP_HASH` — would be both cheaper to compare and strictly stronger than what was there. As it stands the branch is a net loss of a cross-check, which is worth being deliberate about given `PROTOCOL_MAJOR` was bumped anyway.

src/game/default/Makefile.am
```diff
@@ -1,59 +1,17 @@
...
-	g_sound.h \
-	g_types.h \
-	g_util.h \
-	g_weapon.h
+GAME_COMMON_DIR = $(top_srcdir)/src/game/common
 
-noinst_LTLIBRARIES = \
-	libitem.la \
-	libpmove.la
+vpath %.c $(srcdir):$(GAME_COMMON_DIR)
```
jdolan, line 3: **`make dist` is broken by the vpath source lists.**

Verified by running it:

```
(cd default && make ... distdir)
cp: ./bg_pmove.c: No such file or directory
make[8]: *** [distdir-am] Error 1
make: *** [dist] Error 2
```

Common sources are listed as bare names and found through `vpath`, but automake still emits `DIST_SOURCES = $(game_la_SOURCES)`, and `distdir-am` looks for each entry in the build dir then `$(srcdir)`. `src/game/default/bg_pmove.c` is in neither. Compounding it, the old `noinst_HEADERS` lists were deleted and nothing replaced them, so no module header would ship either.

Nothing in CI or packaging runs `dist` — Linux packaging goes through `make install` into staging — so this does not block the merge. But it is a standard target that worked before, and the failure is one a distro packager hits rather than you.

The fix that preserves the vpath design: mark the common entries `nodist_game_la_SOURCES` (they still compile; they just leave `DIST_SOURCES`) and add an `EXTRA_DIST` naming the `src/game/common` and `src/cgame/common` trees once, at `src/game/Makefile.am` and `src/cgame/Makefile.am`. Listing the common sources with `$(GAME_COMMON_DIR)`-relative paths instead would also work but gives up the per-module manifest readability the branch is built around.

src/game/common/g_hook.c
```diff
@@ -0,0 +1,621 @@
...
+  trail->s.angles = projectile->s.angles;
+
+  gi.LinkEntity(trail);
+  return projectile;
+}
+
+/**
+ * @brief Detach the player's hook if it's still attached.
+ */
+void G_HookDetach(g_client_t *cl) {
```
jdolan, line 447: **Disabling the hook mid-level strands a client who is attached at that moment.**

`G_CheckCvars_Hook` returns `super.CheckCvars()` without requesting a restart, unlike `G_CheckCvars_Tech` which sets `restart = true`. Both `G_HookDetach` here and `G_HookThink` return immediately when `!g_hook_enabled`.

So a client pulling on the hook when an admin sets `g_hook 0`: `g_hook_enabled` goes false, no restart happens, `G_HookThink` never runs again, and `G_HookDetach` refuses to act — `cl->hook.pull` stays true, so `G_PrepareMove_Hook` keeps forcing `PM_HOOK_PULL` and the projectile and trail entities keep thinking, until the client dies or the map changes.

The early-return is what makes it unrecoverable rather than merely odd. Either tear down an already-attached hook before returning, or set `restart = true` as the techs do and let the level restart clear it.

src/client/cl_parse.c
```diff
@@ -275,13 +275,22 @@ static void Cl_ParseServerData(void) {
...
-  if (q_strcmp(Cvar_GetString("game"), str)) {
 
-    Fs_SetGame(str);
+  if (!Fs_ValidGame(str)) {
+    Com_Error(ERROR_DROP, "Server sent an invalid game directory\n");
+  }
+
+  Com_SetGame(str);
+
+  Cvar_ForceSetString("game", str);
```
jdolan: **`Cvar_ForceSetString("game", ...)` permanently rewrites the user's choice and discards a pending latch.**

`Cvar_Set_`'s force branch frees `latched_string`, so a listen-server host with `game ctf` latched who joins any remote server loses the pending change silently.

The sharper edge is the interaction with the check two lines below: if `minor != cls.cgame->protocol` drops, `Error()` longjmps past `game->modified = false` in `Frame`, leaving the client disconnected with the *server's* game mounted, its cgame loaded, and `game` set to it — so a subsequent local `map` hosts a module the user never chose.

Suggest setting the cvar only after the protocol checks pass, and restoring the previous value on `Cl_Disconnect`.

src/client/cl_cgame.c
```diff
@@ -313,8 +313,14 @@ void Cl_InitCgame(void) {
   import.Draw3DLines = R_Draw3DLines;
   import.Draw3DBox = R_Draw3DBox;
 
-  cgame_handle = Sys_OpenLibrary("cgame", true);
+  cgame_handle = Sys_OpenLibrary("cgame");
```
jdolan, line 316: **A failed `Cl_InitCgame` leaks `cgame_handle` and leaves a stale `addClassImage` registration.**

If `Sys_OpenLibrary` succeeds and `addClassImage(cgame_handle)` runs, but `Sys_LoadLibrary` then errors — an unresolved `Cg_LoadCgame`, or the `api_version` mismatch below — `cls.cgame` stays NULL. On the next `Cl_InitCgame`, the `if (cls.cgame)` guard is false, so `Cl_ShutdownCgame` is skipped: the previous `dlopen` handle is overwritten and never closed, and its class image is never `removeClassImage`d, so `classForName` keeps a registered image nobody owns.

`MAX_CLASS_IMAGES` is 8 with an `assert`, so repeated failed loads eventually abort. This matters more now than it would have before, because dropping `RTLD_GLOBAL` makes the `addClassImage`/`removeClassImage` pairing load-bearing rather than incidental.

Suggest tracking and cleaning `cgame_handle` independently of `cls.cgame` — unregister and close at the top of `Cl_InitCgame`, or on each error path.

jdolan:
## Review: 362 files, five reviewers, five areas

Reviewed the engine, the game common seams, the cgame/HUD split, the three build systems, and cross-module divergence. Verification I ran myself, not just read:

- Clean rebuild of all six module targets (`game`/`cgame` x `default`/`ctf`/`lithium`): **290 objects, zero warnings**.
- `src/tools/verify_projects.py`: **6/6 modules agree across autotools, MSVS and Xcode**, with the expected per-module define sets.
- Every subset of `G_CTF`/`G_HOOK`/`G_TECH` now compiles (it did not before — see below).
- `make dist` (fails), and CI green on linux + windows.

The architecture holds up. The mechanism/content split, the chainable hooks with their `_Common` tails, and the install-once-per-image discipline are consistent across all 13 game hooks and the one cgame hook; `doc/game-module-hooks.md` matches the code. `lithium` really does demonstrate the seam — no source shared with `ctf`, and the composition is visible in the built objects. The engine-side race the PR set out to fix **is** fixed: moving `Fs_SetGame` ahead of `Cl_Init` plus `Com_SetGame` inside `Sv_InitEntities` closes it on dedicated startup, latched mid-session change, listen-server map change and shutdown, and PhysFS mount order is right.

### Fixed and pushed (7 commits)

| Area | Defect |
|---|---|
| Windows/macOS install | **lithium was never installed, and installing it clobbered ctf.** Both lithium projects ran `COPY_GAME_CTF`/`COPY_CGAME_CTF`, which hard-code `bin/<build>/ctf` → `lib/ctf`, while lithium's `OutDir` is `lithium\`. Parameterized the scripts by module name. |
| Windows | `quetoo.sln` had the ctf projects but not lithium's, so the primary solution built 4 of 6 modules. Only `quetoo_all.sln`, which CI uses, built them all. |
| macOS | `apple/Makefile` enumerated the `dylibbundler -x` list by hand and stopped at ctf — lithium's modules would keep link-time Homebrew paths in a signed bundle and fail to `dlopen` off the build machine. Now generated from `MODULEDIR`, as `move-resources` and `sign` already do. |
| Hook contract | **`G_CTF` or `G_TECH` without `G_HOOK` did not compile at all.** `g_module.h` declares `PrepareMove` in terms of `pm_move_t` but included only `g_types.h`, and ctf's manifest had dropped `bg_pmove.h`. ctf built only because `g_local.h` pulls `g_hook.h` in first. |
| Create Server menu | **The teams `CvarSelect` wrote a truncated `cg_team_mode_t *` into `g_teams`.** Option values became pointers while the control stayed bound to the cvar, so opening the menu set `g_teams` to a large nonzero value and started team play regardless of what was shown; `updateBindings` compared the same pointer against an integer, so no mode ever appeared selected. |
| Create Server menu | The `"Default"` gameplay option was dropped, but `g_gameplay` still defaults to `"default"` meaning "defer to the map" — so nothing matched out of the box and there was no way back from Instagib. |
| Startup | `game->modified` was left set by `Cvar_Add`, so **`autoexec.cfg` executed twice on every client launch**, including plain `default`. |
| HUD | `STAT_CHASE` admitted `MAX_ENTITIES` itself (one past `entities[]`); `STAT_TECH` indexed `cg_items` unchecked and dereferenced the icon found; `CROSSHAIR_HEALTH_LAST` named a nonexistent enumerator. |
| Seams | `g_effect.h` was the only `game/common` header without the `__G_LOCAL_H__` guard; the `G_CTF` teams announcement could only ever say "enabled"; `g_ctf.c` was out of alphabetical order; `cgame-lithium`'s filters still filed lithium's `bg_item` under `src\game\ctf`. |

### Raised inline, not fixed — these are yours to call

Two I would want resolved before merge:

1. **`drop flag` only works for the red flag carrier** (`g_ctf.c:153`). All four flags are named `"Enemy Flag"`, so resolving the carried flag to a *name* and looking it back up always returns `FLAG_RED`. A regression — `G_Drop_f` used to call `G_TossFlag` on the carried flag. The name round-trip is the root cause; the hook probably wants to take a `const g_item_t *`.
2. **All three modules advertise `PROTOCOL_MINOR 1044`** despite incompatible wire layouts, and `Sys_OpenLibrary` silently falls back to `default/cgame.so` when a module is missing. Together those turn "you don't have this mod" into a garbage HUD rather than a refused connection. The silent fallback is worth fixing on its own merit.

Also raised: the cgame hot-swap while connected to a remote server (`cl_main.c:640`, no media load, no protocol recheck); the new fatal BSP check firing for listen-server hosts, who were explicitly exempt on `main` (`cl_media.c:38`); the loss of the server-authoritative BSP cross-check with `CS_BSP_SIZE` (`cl_predict.c:307`); `make dist` broken by the vpath source lists (verified, but nothing in CI or packaging runs it); `Cvar_ForceSetString("game", ...)` discarding a pending latch (`cl_parse.c:291`); a leaked `cgame_handle` + stale class registration on failed init (`cl_cgame.c:316`); and the hook stranding an attached client when disabled mid-level (`g_hook.c:447`).

Three more that had no diff line to attach to:

- `src/server/sv_client.c:48` advertises `Cvar_GetString("game")` rather than `Fs_Game()`. Since `Fs_SetGame` silently rejects invalid names and the client now treats a bad value as fatal, a one-character admin typo drops every connecting client instead of falling back.
- `src/common/filesystem.c` — `Fs_SetGame` cannot report failure. `PHYSFS_unmount` fails while files are open (a demo under the old game), and the failure path returns with `fs_state.game` stale and the search path half dismantled, which makes `Cl_InitCgame` re-fire on every `server_data`. Also leaks `paths`. Worth making it return `bool` and propagating through `Com_SetGame`.
- `configure.ac:146` — `Objectively >= 2.0.0` no longer expresses the requirement now that `addClassImage`/`removeClassImage` are load-bearing; a conforming 2.0.x without them fails at link time rather than at configure.

### Checked and clean

Per-module item invariants hold in all three modules (roster/enum/tag order agree within each; stat counts under `MAX_STATS`; `g_score_t` blitting resolves each module's own `g_types.h` in all three build systems). Every hook chain has a non-NULL `_Common` initializer, so hooks are safe before `G_Init`; install order gives the documented composition. Guard balance verified across all 80 guard sites. Objectively class registration is balanced across init/shutdown with removal while the handle is still open, and `removeClassImage` only unlinks — so a surviving image re-registering on reload is safe. No residual cross-module naming in `lithium` or `default`. UI assets fully re-homed from the 12 deleted `ui/**/Makefile.am`, both directions.

One thing worth recording: the removals of the per-map `teams`, `num_teams`, `give`, `hook` and `techs` metadata all read as bugs from the diff alone, and two reviewers flagged them as such. They are deliberate, with the rationale in `e7acf31e8`. `doc/game-module-hooks.md`'s "Behaviour that changed on purpose" section doesn't mention them — adding them there would save the next reader the same detour.

Nothing I found argues against merging today once (1) and (2) are settled.

src/game/common/g_ctf.c
jdolan: Fixed in c3f085596. The chain now resolves to a `const g_item_t *` rather than to another name, so ctf and the techs return the item being carried instead of naming it. Dropping is a plain `G_DropInventoryItem(cl, it)` over an already-resolved item, and `G_Drop_f` keeps the one message that needs the string the client typed. The techs had the same latent bug and only escaped it because their five names happen to be unique.

src/client/cl_main.c
jdolan: Fixed in 1b5e29b81, by removing the mechanism rather than guarding it. `game` is no longer a cvar, so there is no `modified` flag for two different frames to poll: changing games is a command that brings any session down first. The block this comment points at is gone.

src/client/cl_parse.c
jdolan: Fixed in 1b5e29b81. `Cvar_ForceSetString("game", ...)` is gone along with the cvar itself — `Cl_ParseServerData` now calls `Com_SetGame(str)`, which records the state and has no latch to clobber.

src/game/default/Makefile.am
jdolan: Won't fix, per the author: `make dist` is a relic, and nothing in CI or packaging runs it — people clone the repo.

Recording the diagnosis only so the next person who tries it is not surprised: automake emits `DIST_SOURCES` from the `_SOURCES` lists, which name the common sources by bare name, and `distdir-am` looks for each in the build dir then `$(srcdir)` — neither of which is where `vpath` found them.

src/game/ctf/g_types.h
jdolan: Addressed, though not by changing the minor. Two commits closed the hole this described from both ends:

- `567d33b62` — `Sys_OpenLibrary` now resolves `game.so`/`cgame.so` only within the directories that *are* the game that is current, so the silent fallback to `default/cgame.so` is gone. A missing module is a missing module: `Couldn't find cgame.so for game ctf`.
- `e539aeefd` — `cg_export_t` carries the `GAME_NAME` each module's own `g_types.h` defines, and `Cl_ParseServerData` drops when it disagrees with the game directory the server advertised. The module answers for itself, rather than the client trusting what it asked for.

With those, a `default` cgame can no longer be handed a `ctf` server's stream, which was the corruption path. `PROTOCOL_MINOR` is back to its original job — version skew within one module — so per-module minors would now be defence in depth rather than a correctness fix. Leaving them shared.

src/client/cl_cgame.c
jdolan: Still open, but narrowed. The `game` command pre-checks `Sys_HasLibrary` before it tears anything down, so the most reachable cause — asking for a game that ships no `cgame` — no longer reaches this path. What remains is a failure *after* `dlopen` succeeds: an unresolved `Cg_LoadCgame`, or the `api_version` mismatch below. Those still leave `cgame_handle` open and its class image registered, with `cls.cgame` NULL so the next `Cl_InitCgame` skips the shutdown.

Worth tracking separately rather than holding the branch.

src/client/cl_media.c
jdolan: Still open, and of everything left on this branch it is the one I would take before merging: it is a regression rather than a pre-existing wart. `main` skipped this check whenever a local server was running (`!Com_WasInit(QUETOO_SERVER) || cl.demo_server || !Cm_NumModels()`), so a listen-server host with a stale local `.mf` got a warning. Now they get `ERROR_DROP` and cannot start their own map — and by construction there is nothing to detect there, because client and server are the same process reading the same file.

jdolan:
## Review closed out

**Resolved on the branch since the review**, with the inline threads resolved:

| Finding | Fixed by |
| --- | --- |
| `drop flag` only worked for the red flag carrier | `c3f085596` — the chain resolves to a `g_item_t *`, not to another name |
| cgame hot-swapped under a live remote connection | `1b5e29b81` — `game` is a command, so there is no `modified` flag for two frames to race over |
| `Cvar_ForceSetString("game")` discarded a pending latch | `1b5e29b81` — the cvar is gone |
| `default/cgame.so` silently substituted for a missing mod | `567d33b62` — the loader resolves only within the game's own directories |
| Shared `PROTOCOL_MINOR` could not detect a module mismatch | `e539aeefd` — each module advertises its `GAME_NAME`, checked against the game the server sent |
| Create Server wrote a truncated pointer into `g_teams` | `38145ad8f` |
| `autoexec.cfg` ran twice at startup | `ead87a529` |
| lithium never installed on Windows, and clobbered ctf | `08d54ce17` |
| lithium omitted from `dylibbundler` | `08d54ce17` |
| `G_CTF`/`G_TECH` without `G_HOOK` did not compile | `bdbbdac1a` |

`make dist` is a **won't fix** — a relic, and nothing in CI or packaging runs it.

**Carried forward as issues**, none of which blocks this branch:

- #918 — listen server host cannot start a map with a stale local `.mf` (`bug:major`; the one true regression against `main` here)
- #919 — dropping `CS_BSP_SIZE` lost the server-authoritative BSP check (`bug:major`)
- #920 — failed `Cl_InitCgame` leaks the cgame handle and its class registration
- #921 — `Fs_SetGame` cannot report failure, leaving a half-dismantled search path
- #923 — grapple strands an attached client when disabled mid-level
- #924 — autocomplete matches come out unsorted (pre-existing, hit while adding `game` completion)

`doc/game-module-hooks.md` now records the deliberate deletions — the per-map gameplay settings, `G_CreateTeamSpawnPoints`, the `g_ctf` cvar — since two reviewers read them as regressions from the diff alone, and what a module must ship for the loader to accept it.

**Verification.** All six module targets build clean with zero warnings; `src/tools/verify_projects.py` reports autotools/MSVS/Xcode agreement for all six; CI green on linux and windows. Runtime, against the installed tree: `default`/`ctf`/`lithium` each load from their own directory and spawn 80/85/85 entities on `edge`; `game` reports the current module; an invalid name and a module-less game are both refused and rolled back.

**Not verified by me:** anything requiring the GUI client. Every runtime check above drove `quetoo-dedicated`, so `game <name>` from the in-game console and the Create Server menu fixes are verified by construction only.

src/client/cl_media.c
jdolan: Retracting this. `quemap` writes the manifest unconditionally after every compile and includes the BSP itself as an entry, hashed from the file it just wrote, so the pair is consistent by construction — a host does not fall into the state I described.

Filed as #918 and closed as invalid; the one real defect in this function, the manifest leaking when the drop longjmps out of the enumeration, is carried to #919.

Nothing here should hold the merge.

jdolan:
### Correction to the review summary above

I called #918 the one thing to take before merging. That was wrong, and @jdolan caught it: `quemap` writes the manifest unconditionally after every compile (`src/quemap/main.c:378`) and includes the BSP itself as an entry hashed from the file just written (`src/quemap/manifest.c:309`), so the `.bsp`/`.mf` pair is consistent by construction. #918 is closed as invalid, and its one real residue — the manifest leaking when the drop longjmps out of the enumeration — moved to #919.

**So there is no outstanding merge blocker from this review.** The carried-forward list is now:

- #919 — dropping `CS_BSP_SIZE` lost the server-authoritative BSP check (`bug:major`)
- #920 — failed `Cl_InitCgame` leaks the cgame handle and its class registration
- #921 — `Fs_SetGame` cannot report failure, leaving a half-dismantled search path
- #923 — grapple strands an attached client when disabled mid-level
- #924 — autocomplete matches come out unsorted (pre-existing)

The only thing I could not check myself remains the GUI client: `game <name>` from the in-game console and the Create Server menu fixes are verified by construction, since every runtime check I ran drove `quetoo-dedicated`.

Issue #925, opened 2026-08-07
quemap should support -game parameter for searchpath convenience


jdolan:
Now that Quetoo properly supports mods with e.g. `+game ctf` it would be nice to add a `quemap -game` switch for convenience. You can already compile maps in mod directories if you mess with `-w /path/to/mod` but it's tedious and should not be required.

Acceptance criteria:

- I have a Capture the Flag map `myctf.map` in my user-local Quetoo CTF folder, e.g. `~/Library/Application Support/WickedOldGames/Quetoo/ctf`
- I run `quemap -game ctf -bsp maps/myctf.map`
- `quemap` registers both the read and write paths correctly
- `quemap` finds the .map file and authors the .bsp file in the same directory

jdolan:
Fixed in #932

PR #932, opened 2026-08-07
Add -game switch to quemap for mod search paths

jdolan:
## Summary
Implements #925. `quemap` had no convenient way to compile a map for a mod (e.g. CTF) — you had to manually pass `-w /path/to/mod`, or the map compiler would use only the base game's search paths.

## Fix
Added a `-game <name>` switch that reuses the engine's existing `Com_SetGame`/`Fs_SetGame` machinery (the same code path `src/main/main.c` uses for `+game`), so `quemap -game ctf -bsp maps/myctf.map` mounts both the read and write search paths for the `ctf` mod and authors the `.bsp` in the same directory as the `.map`. Defaults to `DEFAULT_GAME` when omitted, preserving existing behavior.

## Test plan
- [x] Builds cleanly (`autoreconf -i && ./configure && make -j4`)
- [x] Manually verified against real Quetoo user data: `quemap -game ctf -bsp maps/myctf.map` correctly mounted the ctf read/write paths and compiled the BSP stage into the ctf maps directory (verified with a placeholder brush-less map; lighting stage errors as expected for an empty map, which is unrelated to path resolution)

### DIFF
--- a/src/quemap/main.c
+++ b/src/quemap/main.c
@@ -227,6 +227,7 @@ static void PrintHelpMessage(void) {
   Com_Print("-t --threads <int> - Specify the number of worker threads (default auto)\n");
   Com_Print("-p --path <game directory> - add the path to the search directory\n");
   Com_Print("-w --wpath <game directory> - add the write path to the search directory\n");
+  Com_Print("-game <name> - compile against the given game/mod, e.g. ctf\n");
   Com_Print("\n");
 
   Com_Print("-bsp               BSP stage options:\n");
@@ -265,6 +266,7 @@ static void PrintHelpMessage(void) {
  */
 int32_t main(int32_t argc, char **argv) {
   int32_t num_threads = 0;
+  const char *game = DEFAULT_GAME;
 
   printf("Quemap %s %s\n", VERSION, BUILD);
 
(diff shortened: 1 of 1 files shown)


jdolan:
Renamed `-game` to `-g`/`--game` to match the short/long convention used by every other quemap flag (`-v`/`--verbose`, `-d`/`--debug`, `-t`/`--threads`, `-p`/`--path`, `-w`/`--wpath`).

src/quemap/main.c
```diff
@@ -227,6 +227,7 @@ static void PrintHelpMessage(void) {
   Com_Print("-t --threads <int> - Specify the number of worker threads (default auto)\n");
   Com_Print("-p --path <game directory> - add the path to the search directory\n");
   Com_Print("-w --wpath <game directory> - add the write path to the search directory\n");
+  Com_Print("-g --game <name> - compile against the given game/mod, e.g. ctf\n");
```
jdolan:
```suggestion
Com_Print("-p --path <path> - add the path to the search directory\n");
Com_Print("-w --wpath <path> - add the write path to the search directory\n");
  Com_Print("-g --game <name> - compile against the given game/mod, e.g. ctf\n");
```

PR #933, opened 2026-08-07
Merge g_teams into g_gameplay as a bitmasked enum

### LINKED ISSUE #927: g_gameplay should strictly coerce to enum types
In the servers list, it looks like server operators are using random-ish values for `g_gameplay`:

Image

This cvar should be coerced to the already defined gameplay enum types:

- deathmatch
- instagib
- arena

No other values should be supported / persisted in the cvar. The `g_gameplay->modified` check can enforce this.

I do see, however, value in advertising "team deathmatch" or "team arena" tho..

Perhaps we should merge `g_teams` with `g_gameplay` so that there are 6 enum values?

- deathmatch
- team deathmatch 
- instagib
- team instagib
- arena
- team arena

We would need a simple shorthand for checking "Are teams enabled?"; perhaps these enum values are bitmasked with G_GAMEPLAY_TEAMS occupying some high bit?

This would also simplify the menus: Cg_TeamModes would completely go away, removing one of the warts on the cgame cg_module API.

jdolan:
## Summary
Fixes #927. Server operators were setting `g_gameplay` to arbitrary/garbage string values, which showed up as noise in the server browser. Beyond the literal ask (strictly coerce `g_gameplay` to valid values), this implements the larger redesign floated in the issue: merge the separate `g_teams` cvar into `g_gameplay` as a bitmask, giving 6 total modes instead of 3+a-toggle, and removing `Cg_TeamModes` from the cgame menu code since team-ness becomes intrinsic to gameplay.

## Design
- `g_gameplay_t` gains a `GAME_TEAMS` high bit (`1 << 3`) plus named combined constants: `deathmatch`, `team_deathmatch`, `instagib`, `team_instagib`, `arena`, `team_arena`.
- `g_gameplay->modified` now parses the string via `G_GameplayByName`, coerces it back to canonical via `G_GameplayCvarString` + `gi.SetCvarString` (this is the actual fix for the reported bug — garbage no longer survives into the server-info string the browser reads), derives `g_level.teams` from the bit, and calls `G_InitNumTeams()`.
- `g_teams` is removed entirely (cvar, registration, modified handler, worldspawn read). CTF still forces `GAME_TEAMS` unconditionally, same as its prior forced-teams behavior — now folded into the cvar itself before it's written, so a CTF server's `g_gameplay` correctly reads e.g. `team_instagib`.
- `Cg_TeamModes` and its "Teams" menu selector are removed from all three cgame modules (default/ctf/lithium); the existing `gameplay` `CvarSelect` now lists all six canonical values directly. Swept through autotools, MSVS, and Xcode project files — `verify_projects.py` confirms all six modules still agree.
- Incidental fix: `Cg_UpdateConfigString` had no `CS_GAMEPLAY` case, so `cg_state.gameplay` was permanently 0 and the HUD instagib check / Discord mode string were dead code. Added the missing case since the redesign is unreachable in cgame otherwise.

## Test plan
- [x] Builds cleanly across game/cgame (default, ctf, lithium), client, server, master, quemap
- [x] `verify_projects.py` reports all six modules agreeing across autotools/MSVS/Xcode
- [ ] Manual: set `g_gameplay garbage`, confirm it coerces to a valid value with exactly one restart/broadcast; confirm a CTF server's browser entry shows a `team_` prefixed mode

## Breaking change
Existing server configs doing `set g_teams 1` alongside `g_gameplay deathmatch` will silently lose team play — `g_teams` becomes an inert, unrecognized cvar. Operators need `g_gameplay team_deathmatch` instead. Worth a release-note callout.


jdolan:
Pushed two more commits addressing a follow-up ask: instead of trimming \`GAME_INSTAGIB\`/\`GAME_ARENA\` out of CTF's enum (which would break compilation of six shared files that reference those symbols directly), added two non-chained module hooks matching the existing pattern in \`doc/game-module-hooks.md\`:

- **\`ClampGameplay\`** (game side, \`g_module.h\`): lets a module coerce a parsed gameplay value to what it actually supports. CTF's override always returns \`GAME_TEAM_DEATHMATCH\`. Replaces two \`#if defined(G_CTF)\` guards in \`g_main.c\`/\`g_entity.c\` that only forced the teams bit, not the base mode — a CTF map with \`gameplay "arena"\` metadata previously played arena-with-teams-forced rather than actual captures; now it correctly clamps to team deathmatch.
- **\`ListGameplayModes\`** (cgame side, \`cg_module.h\`): lets a module report which modes it supports for the create-server menu. CTF returns just team deathmatch; the "Gameplay" row hides entirely when only one mode is available. Reviewed and confirmed the hide happens after the hook is installed (\`Cg_Ctf_Init\` runs well before the create-server screen is ever constructed), so there's no stale-hook window.

Both hooks reviewed independently (build + \`verify_projects.py\` re-run, control-flow re-traced given this cvar-coercion logic has had two real bugs found in earlier rounds). One pre-existing gap the review flagged: starting a CTF server with \`+set g_gameplay arena\` still advertises \`"arena"\` in the server-info string at startup, since \`G_Init\` suppresses the \`modified\` flag before \`G_CheckRules\` ever runs — the level plays correctly (team deathmatch) but the browser would show the pre-coercion value until the cvar is touched again. Flagging for a decision on whether that's in scope here or a follow-up.

jdolan:
Pushed a third commit (`16107c9fc`) fixing the startup-coercion gap flagged above: extracted the parse/clamp/coerce sequence into `G_CoerceGameplay()`, called once from `G_Init` (after `G_Ctf_Init` has installed its `ClampGameplay` override, so this genuinely applies per-module) right before the existing block that suppresses first-frame `modified`-flag noise, and from `G_CheckRules`'s runtime path so there's one implementation. Reviewed and confirmed: hook-install ordering is correct, no restart/broadcast fires during init, the `"default"` sentinel and already-canonical strings remain no-ops, and the runtime path is unchanged. Build + `verify_projects.py` clean.

This PR is now fully reviewed through three rounds — ready for your review/merge whenever you're ready.

PR #938, opened 2026-08-12
Tune func_bob curve/timing, add trigger_push start_off/toggle

skies912:
## Summary

**func_bob** (follow-up tuning after #911):
- Scale the arrival-snap threshold to the leg's own travel distance (1% of distance, floored at `.01`, capped at the old flat `.25`) instead of a fixed `.25` units. `G_MoveInfo_Linear_Done` snaps position exactly to the destination once within this threshold, and a flat `.25` was a visible pop on short legs (e.g. ~3% of an 8-unit travel used for a shadow-sync bob).
- Auto-derive `compression` from the leg's duration when the key is unset, instead of always defaulting to a flat `1.0` (pure sine). Uses a continuous falloff in `log(leg_time)` that peaks at `1.0` for a leg around 1 second and tapers toward a `.35` floor for legs much shorter (too few ticks, stutters) or much longer (the sine's slow "wings" become a multi-second crawl before the entity visibly starts moving) than that. Calibrated against two tested legs (~1s and ~7s). Explicit `compression` keys still override.
- `drift` now defaults to `0` (no startup delay) instead of silently defaulting to a full-cycle random delay whenever unset - the old default caused multi-second startup delays on any `func_bob` that wasn't actually part of a desynced array. Negative values are clamped to `0` rather than risking undefined behavior in the float-to-`uint32_t` cast.

**trigger_push**:
- Adds `start_off` and `toggle` spawnflags so `trigger_push` can spawn inactive and be toggled on/off via `Use`, matching the pattern already used elsewhere (e.g. `func_bob`).

## Test plan
- [x] Builds clean (`game.vcxproj`)
- [x] Manually tested in-map: short-leg arrival snap, long-leg startup delay, auto-compression on both a small shadow-sync bob and a 448-unit platform
- [ ] `trigger_push` start_off/toggle behavior

### DIFF
--- a/src/game/common/g_entity_func.c
+++ b/src/game/common/g_entity_func.c
@@ -699,15 +699,22 @@ static void G_func_bob_Bottom(g_entity_t *ent);
 static void G_func_bob_Ease(g_entity_t *ent) {
   g_move_info_t *move = &ent->move_info;
 
+  const float total_distance = Vec3_Distance(move->start_origin, move->end_origin);
+
   vec3_t dir;
   const float remaining = Vec3_DistanceDir(move->dest, ent->s.origin, &dir);
 
-  if (remaining <= .25f || Vec3_Dot(dir, move->dir) < 0.f) {
+  // G_MoveInfo_Linear_Done snaps position exactly to `dest`, so this threshold is the size of
+  // that final pop. A fixed .25 units is invisible on a long mover but reads as a visible jump
+  // on a short leg (e.g. .25 is ~3% of an 8-unit travel) - scale it down for short legs instead,
+  // capped at the old constant so long legs keep the same tolerance as before.
+  const float arrival_epsilon = Clampf(total_distance * .01f, .01f, .25f);
+
+  if (remaining <= arrival_epsilon || Vec3_Dot(dir, move->dir) < 0.f) {
     G_MoveInfo_Linear_Done(ent);
     return;
   }
 
-  const float total_distance = Vec3_Distance(move->start_origin, move->end_origin);
   const float traveled = total_distance - remaining;
 
   const float t = traveled / total_distance;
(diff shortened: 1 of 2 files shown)


src/game/common/g_entity_func.c
```diff
@@ -839,22 +851,39 @@ void G_func_bob(g_entity_t *ent) {
...
+  // than every tick in G_func_bob_Ease. If unset, auto-derive it from the leg's duration instead of
+  // defaulting to a flat 1 (pure sine): a full sine only reads right for a leg lasting roughly a
+  // second. Shorter legs (too few ticks to render the ease smoothly) and longer legs (the sine's
+  // slow "wings" become a disproportionate chunk of real time, e.g. a multi-second crawl before a
+  // platform gets moving) both want a flatter curve; a continuous falloff in log(leg_time) treats
+  // "half as long" and "twice as long" symmetrically, tapering toward a floor the farther the leg
+  // is from that ~1s sweet spot in either direction. Calibrated against two tested legs: ~1s
+  // (wants ~1.0) and ~7s (wants ~0.5) - solving for the width that fits both against a 0.35 floor.
+  // log(0) degrades gracefully to the floor for a zero-velocity leg, no special case needed.
+  const float compression = gi.EntityValue(ent->def, "compression")->value;
```
jdolan, line 879: Heh, okay. These kinds of optimizations are pretty pointless. This will never show up on a profiler, I guarantee it. But it's fine if Claude thought it was being clever here.

jdolan (approved):
Word. Don't forget the TrenchBroom / Radiant configs for the new / modified entity fields :)

src/game/common/g_entity_func.c
skies912: well, the problem that I found was that when using very small distances like 16u (was exp with bobbing lights)the movement stutters before resting and using compression to fix isn't accurate. Slow speeds like <15 are also affected. This fixes it so the movement is alot smoother.

jdolan: Oh, not the calculation. The storing of it on ent->random so that it's "not calculated every tick" was the silly bit 👍 It's a cheap calculation.

PR #953, opened 2026-08-22
Split Linux packaging into quetoo-client and quetoo-server

### LINKED ISSUE #952: RPM and DEB packages for Linux clients
Currently, Linux clients use a binary blob release and the in-game updater to fetch content updates between binary versions. The rationale for the split was twofold:

- Content will be updated more frequently than code (currently the exact opposite is true)
- S3 download costs would make frequent 1GB downloads unsustainable for me financially (we decided to instead use GitHub releases which does not charge for downloads!)

The dedicated server packaging for Linux already produces .deb and .rpm. We could extend this to the client as well. This would require overhauling our release.yml workflow to produce the client packages and to include them in the release page template. The package names might need to be made explicit: quetoo-client and quetoo-server, for example. This would involve modifying the quetoo-update script (and possible other Linux tooling) that the dedicated server installs rely on today.

We would also need to update the Downloads page in https://github.com/jdolan/quetoo-www to link them properly. And, of course, we would keep the current Linux client release as-is, so that we don't abandon / trip up current users who like the packaging they have today.

jdolan:
## Summary
- Splits the single combined `quetoo` Linux `.deb`/`.rpm` (client + dedicated server + master + quemap) into explicit `quetoo-client` and `quetoo-server` packages, per #952.
- `quetoo-client` also includes the `quetoo-dedicated` binary, so a player can run a local dedicated instance without pulling in `quetoo-server`'s systemd/sudoers footprint.
- `quetoo-client` and `quetoo-server` are each fully self-contained (own copy of shared runtime libraries, `game.so` modules, and `quetoo-update`/`quetoo-status`), and declare `Conflicts` against each other — there's no real use case for a single machine to be both a player's desktop and a headless server operator's box.
- `quetoo-update` now detects and updates whichever of `quetoo-client`/`quetoo-server` is actually installed instead of assuming a single `quetoo` package.
- Both packages declare `Replaces`/`Conflicts: quetoo` for a clean upgrade path from the old combined package.
- Updates `release.yml`/`build.yml` artifact names, the release notes template, and `linux/docker/` (`build-pkgs.sh`, `Dockerfile.dedicated`, README) for the new package names.
- The existing `.tar.gz` client archive and the S3-backed in-game installer are unchanged — this is additive, not a replacement.

## Design notes
- Each package is built from its own `linuxdeploy`-bundled AppDir (client executables vs. server executables), so each only bundles the shared libraries its own binaries actually link against.
- An earlier version of this PR introduced a third `quetoo-common` package to de-duplicate the overlap between client and server. That added a 4th package to install (alongside `quetoo-data`) for what's fundamentally two distinct, non-overlapping audiences, so it was dropped in favor of accepting some duplicated bytes and making the two packages mutually exclusive instead.
- `quetoo-www`'s Downloads page is a separate repo and is not updated here; worth a follow-up issue there once these packages are actually published.

## Test plan
- [x] Built all packages end-to-end in a container mirroring `linux/docker/Dockerfile.build` + `build-pkgs.sh` (autotools build, `make archive deb rpm`).
- [x] Installed `quetoo-client` standalone via `apt install` (Debian bookworm) — `quetoo`, `quemap`, `quetoo-dedicated` all resolve their libraries with no missing dependencies.
- [x] Installed `quetoo-server` standalone via `apt install` — `quetoo-dedicated`/`quetoo-master` resolve cleanly, postinstall creates the `quetoo` system user.
- [x] Installing `quetoo-client` + `quetoo-server` together correctly fails with a `Conflicts` error via both `apt` and `dnf`, instead of a confusing raw dpkg file-overwrite error.
- [x] Same standalone-install checks repeated via `dnf install` (Fedora 39).
- [ ] Not run: an actual GitHub Actions release build (requires a tag push) — the workflow YAML was validated for syntax and hand-traced against the new artifact names, but not executed in CI.

Closes #952


jdolan:
Addressed all four points from Copilot's review in 20e3cdbec:

- **Old `quetoo` package transition** (linux/Makefile:259): added `--replaces quetoo --conflicts quetoo` to all three new packages (client/server/common) for both `.deb` and `.rpm`, so upgrading from the old combined package doesn't hit file-ownership conflicts.
- **Missing runtime deps on `quetoo-common`** (linux/Makefile:255): added explicit `bash`/`curl`/`python3` deps, since `quetoo-update`/`quetoo-status` need them.
- **Hardcoded `MODS` list** (linux/Makefile:40): replaced the static `default ctf lithium` list with `$(ls $(STAGING)/lib/quetoo)`, so packaging now reflects whatever `--with-games` actually built instead of assuming the default set.
- **AppStream desktop-id mismatch** (linux/Makefile:196, suppressed comment): the engine itself hardcodes `quetoo.desktop` as the runtime-registered filename (`src/common/sys.c`), so I corrected the metainfo's `launchable` id to `quetoo.desktop` rather than renaming the installed desktop file, which would've created two divergent desktop entries.
- Also fixed the (also suppressed) release-notes comment: `Recommends: quetoo-data` can't actually fetch anything since these are standalone release-asset files, not served from a repo — notes now say to download quetoo-data separately instead of claiming it's automatic.

Re-verified end-to-end in containers (build via `linux/docker/Dockerfile.build`, cross-install `quetoo-common`+`quetoo-client`+`quetoo-server` together via both `apt` and `dnf`) after these changes.

jdolan:
Reworked per offline discussion: dropped \`quetoo-common\` (3987478db). Four packages for one game was too many — \`quetoo-client\` and \`quetoo-server\` now each fully bundle their own copy of the shared libraries and \`game.so\` modules, \`quetoo-client\` also carries the \`quetoo-dedicated\` binary, and the two declare \`Conflicts\` against each other since there's no real scenario for installing both on one machine. Re-verified the standalone-install and now-correctly-rejected combined-install paths via both \`apt\` and \`dnf\`.

jdolan:
Fixed the re-requested review's finding in 3f1c0fbde: \`quetoo-update\`'s dpkg version check was matching the whole \`\${Status}\` string (\`"install ok installed"\`), so a plain \`apt remove\` (no \`--purge\`) of \`quetoo-server\` — which has conffiles and so lands in \`"deinstall ok config-files"\` state — would still report its old version as installed. Fixed to check only the install-state word. While fixing it I found the exact-string match would've also mis-handled a held package (\`"hold ok installed\"\`), so it now checks that word specifically rather than the whole three-word field — verified all four states (installed, held, removed-with-conffiles, unknown) in a container.

Also dropped a leftover \`cp -a $(STAGING)/share\` in the client AppDir build rule that copied the game-data tree with no downstream consumer (a follow-up finding from a self-review pass, not from Copilot).

PR #965, opened 2026-08-27
Repair check_master for challenged delisting and game selection

jdolan:
## Summary

`--with-tests` has not built since `Ms_RemoveServer` grew its `cmd` argument for challenge-authenticated delisting: `check_master.c` still called the one-argument form. Separately, `Fs_Init` no longer selects a game, so the fixture had no write directory and the blacklist test could not open `servers-blacklist`.

- Select `DEFAULT_GAME` after `Fs_Init` in the fixture.
- Assign the registered server a challenge (only a heartbeat does, and `Ms_Challenge` needs the `/dev/urandom` descriptor `Ms_Init` opens) and assert that a shutdown without the challenge, or with the wrong one, leaves the server listed before the authenticated shutdown drops it.

Phase 0 of #963.

## Test plan

- [x] `check_master`: 2/2 pass
- [ ] CI

### DIFF
--- a/src/tests/check_master.c
+++ b/src/tests/check_master.c
@@ -33,6 +33,8 @@ void setup(void) {
   Mem_Init();
 
   Fs_Init(FS_NONE);
+
+  ck_assert(Fs_SetGame(TEST_GAME, NULL));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


jdolan:
Addressed: the fixture now selects `TEST_GAME` (`"tests"`, defined in `tests.h`) instead of `DEFAULT_GAME`, so `servers-blacklist` and everything else a test writes lands in `<UserDir>/tests/` rather than the player's `default/` profile. #966 is stacked on this for the same define.

PR #966, opened 2026-08-27
Give test fixtures the state the code under test now requires

jdolan:
## Summary

Four tests fail on `main` once `--with-tests` builds (see #965). Stacked on #965 for the `TEST_GAME` define in `tests.h`.

- `check_cm_manifest` (28%), `check_cvar` (50%), `check_filesystem` (0%): `Fs_Init` no longer selects a game, so the fixture has no write directory and every write fails. Select `TEST_GAME` after `Fs_Init`, so writes land in `<UserDir>/tests/` rather than the player's `default/` profile.
- `check_filesystem` also read `quetoo.cfg`, which is the archived-cvar file the client writes into the player's profile and never shipped in data — it only ever passed on a developer's machine. It reads `maps.lst` now.
- `check_r_media` (segfault): `R_EndLoading` calls `$(r_context.device, waitForIdle)` and `R_LoadOcclusionQueries`, which reads `r_occlusion.boxes`. Production always runs `R_InitOcclusionQueries` before any media load, so those are invariants, not states to guard; the fixture now supplies a device whose only method is `waitForIdle` and an empty box list.

Together with #965 this takes the suite from 14/19 to 19/19.

Phase 0 of #963.

## Test plan

- [x] all four tests pass locally (100%)
- [x] nothing is written under `<UserDir>/default/`
- [ ] CI


jdolan:
Closed automatically when #965's branch was deleted. Rebased onto main and reopened as #974.

PR #969, opened 2026-08-27
Validate .nav files before trusting their contents

jdolan:
## Summary

`G_Ai_InitNodes` read `num_nodes` and each `num_links` straight into loop bounds and allocations, ignored every `gi.ReadFile` return value, and never checked that a link id named an existing node — later code indexes `g_ai_nodes` with those ids. It also dereferenced a NULL file if `OpenFile` failed after `FileExists` succeeded.

Reading moves into `G_Ai_ReadNodes`, which checks every read, bounds the counts (`num_nodes < AI_NODE_INVALID`, `num_links <= num_nodes`), rejects out-of-range link ids, and returns `false`; the caller then discards the partial state via `G_Ai_ShutdownNodes` and continues without nodes, exactly as it already did for a bad magic or version. The on-disk format is unchanged.

Phase 0 of #963.

## Test plan

- [x] `default`, `ctf`, `lithium` build with no warnings
- [x] `quetoo-dedicated -p <root> +map edge` with the branch `game.so`: `Loaded 346 nodes with 936 total links`, 80 entities, bots join
- [x] same with `edge.nav` truncated to 300 bytes: `Nav file is truncated at node 7`, map loads, no crash
- [x] all 25 `.nav` files in `quetoo-data` pass the new bounds (max 9 links/node, no out-of-range ids)
- [ ] CI

### DIFF
--- a/src/game/common/g_ai_node.c
+++ b/src/game/common/g_ai_node.c
@@ -1061,6 +1061,89 @@ void G_Ai_Node_Render(void) {
 #define AI_NODE_MAGIC ('Q' | '2' << 8 | 'N' << 16 | 'S' << 24)
 #define AI_NODE_VERSION 2
 
+_Static_assert(sizeof(ai_link_t) == 8, "ai_link_t is the on-disk link record; changing it requires a new AI_NODE_VERSION");
+
+/**
+ * @brief Reads the nodes and links from an open .nav file into `g_ai_nodes`.
+ * @return False if the file is malformed, in which case the nodes read so far
+ * must be discarded.
+ */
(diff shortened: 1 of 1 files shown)


jdolan:
Both addressed:
- `_Static_assert(sizeof(ai_link_t) == 8, ...)` beside `AI_NODE_VERSION`, since the struct is the on-disk record.
- Links are now read one record at a time and validated before being added, so a crafted `num_links` no longer pre-allocates anything: work is bounded by the bytes actually present in the file, and a truncated file fails at the first short read. That avoids inventing a per-node cap the writer does not enforce (shipped files max out at 9 links per node).

PR #970, opened 2026-08-27
Publish a validated hook pull speed to the client

jdolan:
## Summary

The game moves hooked players by `g_hook_pull_speed->value`, but published the cvar's raw *string* on `CS_HOOK_PULL_SPEED`, and the client parsed it with an unchecked `strtof(s, NULL)` into the speed prediction uses (`cg_predict.c:81`). `set g_hook_pull_speed abc` left the server pulling at 0 while the client predicted something else, with no warning on either side.

- `G_Hook_PublishPullSpeed` resets a non-positive (or NaN) value to the default, clears `modified` so the reset is not announced as an admin change, and publishes `%.9g` of the value the game actually uses, from both `ConfigureLevel` and `CheckCvars`.
- The client parses with an end pointer and falls back to the same default with a `Cg_Warn` on garbage.
- The default is `PM_SPEED_HOOK_PULL` in `bg_pmove.h`, beside the other movement cvar defaults, so both sides share it.

Phase 0 of #963.

## Test plan

- [x] `default`, `ctf`, `lithium` game and cgame modules build with no warnings
- [x] `quetoo-dedicated +game lithium +set g_hook_pull_speed abc +map edge`: exactly one `Invalid g_hook_pull_speed "abc", resetting to 800`, no spurious "changed" broadcast, 85 entities
- [ ] CI

### DIFF
--- a/src/cgame/common/cg_main.c
+++ b/src/cgame/common/cg_main.c
@@ -279,9 +279,15 @@ static void Cg_UpdateConfigString(int32_t i) {
       cg_state.items = (g_items_t) strtol(s, NULL, 10);
       return;
 #if defined(G_HOOK)
-    case CS_HOOK_PULL_SPEED:
-      cg_state.hook_pull_speed = strtof(s, NULL);
+    case CS_HOOK_PULL_SPEED: {
+      char *end;
+      cg_state.hook_pull_speed = strtof(s, &end);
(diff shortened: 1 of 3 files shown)


jdolan:
Addressed: both sides now require `isfinite(value) && value > 0`, so `inf` (and overflow) resets to the default like NaN and non-positive values. Verified with `+set g_hook_pull_speed inf` under lithium.

PR #976, opened 2026-08-27
Carry the player's height in the movement parameters

jdolan:
## Summary

Decision D1 on #963. `PM_BOUNDS`/`PM_CROUCHED_BOUNDS` were the one movement tunable that never became a runtime parameter, read directly from `bg_pmove.c`, `g_client.c`, `g_ai_node.c`, `g_ai_main.c` and three cgame files — which is why the Race fork had to shadow all of them to change one number.

- `pm_params_t` gains `height` and `height_ducked` (the box top; footprint fixed), serialized last in the `PS_PM_PARAMS` block. `PROTOCOL_MAJOR` 2029 → 2030.
- `g_stand_height` (36) and `g_duck_height` (6) hydrate them in `G_MovementParams`.
- `Pm_Bounds(params, ducked)` builds the live box; `Pm_Init`/`Pm_CheckDuck` use it. `G_PlayerBounds()` and `Cg_PlayerBounds()` serve the other users.
- Left on the constants deliberately: `G_InitPlayerSpawn` (height-independent `PM_SCALE` math), the client model's legs bounds (runs before any frame's params exist), and the cvar defaults.

With `PM_SCALE == 1.f` the result is bit-identical. The ducked box is now scaled like the standing one, which it was not before; only observable if `PM_SCALE` changes.

## Test plan

- [x] full tree builds with no warnings; test suite 19/19 (`check_net_message` included)
- [x] `edge` spawns 80 / 85 / 85 for default / ctf / lithium with bots
- [x] `+set g_stand_height 32` server runs, bots join
- [ ] CI (Windows job covers MSVS)
- [ ] a client connect (protocol bump) needs a GUI session to confirm


jdolan:
Addressed: the three sites that have a client in hand (`g_ai_main.c` ladder checks, the `g_ai_node.c` node-drop trace) now read `Pm_Bounds(&cl->ps.pm_state.params, false)`, the parameters that client actually moved with. `G_PlayerBounds` stays for the sites with no client (spawn-point and nav-node generation traces) and for respawn, where the client's params are not yet hydrated.

PR #979, opened 2026-08-27
Let a module supply the movement kernel

jdolan:
## Summary

Decision D2 on #963. The Race fork's `bg_pmove.c` is ours plus a self-contained Quake II kernel spliced in with 23 `if (Pm_Q2FamilyPhysics())` branches. This gives it a seam instead of a shadow.

- `pm_move_t` gains `Move`, a kernel strategy `Pm_Move` dispatches to after `Pm_Init`/`Pm_ClampAngles`/`Pm_InitLocal` and the frozen / spectator / dead handling; and `Accelerate`, an observer called on every user-intended acceleration (including over-speed frames, which a strafe trainer needs). Both `NULL` by default, so default / ctf / lithium are unchanged.
- `bg_pmove_internal.h` exports the kernel's working state (`pm`, `pm_locals`, `Pm_Debug`) and its 29 steps, previously `static`, so a module's kernel is written exactly as ours is and compiled into both its modules. Modules are `RTLD_LOCAL` and the Windows `.def` exports only the entry point, so nothing new leaks across images.
- `pm_move_t` gets a struct tag so the callbacks can name it.
- The `_DEBUG` pmove playback path NULLs the two new pointers after reading a recorded move.

## Test plan

- [x] all game and cgame modules build with no warnings; `verify_projects.py` 6/6 (header registered in MSVS props/filters and Xcode)
- [x] `edge` 80 / 85 / 85 for default / ctf / lithium
- [ ] CI

### DIFF
--- a/Quetoo.vs15/cgame-ctf.vcxproj.filters
+++ b/Quetoo.vs15/cgame-ctf.vcxproj.filters
@@ -282,6 +282,9 @@
     <ClInclude Include="..\src\game\common\bg_pmove.h">
       <Filter>src\game\common</Filter>
     </ClInclude>
+    <ClInclude Include="..\src\game\common\bg_pmove_internal.h">
+      <Filter>src\game\common</Filter>
+    </ClInclude>
     <ClInclude Include="..\src\game\ctf\bg_item.h">
       <Filter>src\game\ctf</Filter>
     </ClInclude>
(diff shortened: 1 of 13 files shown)


jdolan:
Good catch — that reset ran after `G_PrepareMove` and would have wiped a module's kernel. It now lives only inside the `_DEBUG` playback branch, right after a recorded frame is read, which is the one place the pointers can be garbage.

PR #981, opened 2026-08-27
Add the client command hooks

jdolan:
## Summary

Second D7 batch on #963.

- `HandleClientCommand(cl, cmd, intermission)` — chainable decision run at the very top of `G_ClientCommand`, ahead of chat, the intermission early-return and the built-in table. A feature answers true for the commands it owns; anything unclaimed falls through exactly as today. This is what retires the fork's hard-coded list of its own command names in the shared table.
- `ClientDidChat(cl, text, team)` — notification at the end of `G_Say_f`, after delivery; muted, empty or flood-limited messages are not reported. `text` is the formatted line the other clients saw.
- Also fixes the Inventory group's docblock, which #980 left saying its tails are only in `g_item.c`.

Tails handle nothing / do nothing, so default / ctf / lithium are unchanged.

## Test plan

- [x] all game modules build with no warnings
- [x] `edge` 80 / 85 / 85 under each module
- [ ] CI

### DIFF
--- a/doc/game-module-hooks.md
+++ b/doc/game-module-hooks.md
@@ -287,6 +287,8 @@ have produced a module that did not compile.
 | `ClientWillDisconnect` | `g_client.c` | — |
 | `ClientWillThink` | `g_client.c` | — |
 | `ClientDidMove` | `g_client.c` | — |
+| `HandleClientCommand` | `g_cmd.c` | — |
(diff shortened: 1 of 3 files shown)


jdolan:
Noted — the docblock now says `text` lives only for the call and a feature that keeps it copies it (landing in the level/frame hooks PR, which touches the same header).

PR #982, opened 2026-08-27
Add the client game's message, lifecycle, entity and presentation hooks

jdolan:
## Summary

Decision D11 on #963: the client-game seams that retire the Race fork's shadows of `cg_main.c`, `cg_media.c`, `cg_entity.c`, `cg_discord.c` and `cg_score.c` — its `cg_module_compat.h` lists exactly these.

- **Decisions:** `ParseServerCommand(cmd)` and `ParseConfigString(index)`, run ahead of the built-in handling (`SV_CMD_CGAME` was reserved for modules but never dispatched); `FilterEntity(ent)` consulted in `Cg_Interpolate` and `Cg_AddEntities` (filtered entities still consume their events); `DescribeGameMode()` (tail = the former `Cg_GetGameMode`); `DrawScores(ps)`, not chained (tail = the former `Cg_DrawScores`; parsing stays in common).
- **Notifications:** `StateDidClear`, `MediaDidLoad`, `SceneDidPopulate(frame)`, `ScreenDidUpdate(frame)`.
- `cg_hud_layout_t.draw_time` replaces the fork's `stat_y = h` trick for suppressing the common clock.
- Doc gains the client hooks table; group names unified under `cg-hooks-`.

All tails are the previous behaviour or no-ops, so default / ctf / lithium are unchanged.

## Test plan

- [x] all cgame modules build with no warnings
- [ ] CI
- [ ] HUD/scoreboard screenshots need a GUI session (no behaviour changed by inspection: every tail is the old code path)

### DIFF
--- a/doc/game-module-hooks.md
+++ b/doc/game-module-hooks.md
@@ -339,6 +339,26 @@ brace so that the two branches of the preprocessor close different blocks,
 compiles and then breaks the next person to edit around it. Restate the condition
 and make the block additive instead.
 
+### The client game's hooks
+
+| hook | tail lives in | installed by |
+| --- | --- | --- |
+| `DrawHudElements` | `cg_hud.c` | ctf, techs |
+| `ListGameplayModes` | `cg_main.c` | ctf |
+| `ClipEntity` | `cg_predict.c` | — |
+| `ParseServerCommand` | `cg_main.c` | — |
+| `ParseConfigString` | `cg_main.c` | — |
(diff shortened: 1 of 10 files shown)


jdolan:
Agreed on the event: a filtered entity now clears `ent->current.event` along with the buffered state, so a re-interpolated frame cannot fire it once the filter changes. The sound is looping ambient state rather than a consumed event, so it is left alone.

Issue #865, opened 2026-06-18
Add a "Call Vote" screen


ThiloteE:
## Summary

Add a "Call Vote" screen

## Motivation
1. I played a 1vs1 against a random newb on a large map and we barely saw each other. The newb quit the game, possibly of boredom.
2. Current behaviour of bots is suboptimal: On some maps, they suicide into liquids and on some other maps, they are way beyond my skill-level. Neither is optimal.

## Proposed Implementation
Add a "Call Vote" screen that allows to vote for ...
- [ ] switching to another map (next map, random map, specific map?)
- [ ] Skill level of bot
- [ ] Set number of bots
- [ ] Frag limit (in addition to time limit)
- [ ] Vote force spec

## Alternatives Considered
- Texting random newbies to meet up in the middle.
- Camping at suicide spots.
- Becoming good enough to not get owned 0/10 by bots.

## Additional Context
<img width="1920" height="1074" alt="Image" src="https://github.com/user-attachments/assets/7c801134-b9a7-45d6-9993-7aabf0ee0d08" />
^ Screenshot of "Call Vote" Screen from Quake Champions: Doom Edition

jdolan:
We _do_ support voting on certain things like map and frag limit, but it's not exposed via the UI at all. We absolutely should do that. The client-game module owns the UI for exactly this sort of thing: being able to parse custom packets from the server to present menus. Good feature request. Thanks!

PR #984, opened 2026-08-27
Let clients vote on the map, the bots, the limits and each other

jdolan:
## Summary

Decisions D10 / D16 on #963, and the game half of #865. Voting becomes a common feature of every module, installed over the new hooks (`HandleClientCommand`, `FrameDidEnd`, `ClientWillDisconnect`, `ConfigureLevel`) the way `g_ctf.c` and `g_tech.c` are.

- `vote <type> [argument]` calls a vote; `vote yes` / `vote no` casts. Common types (`bg_vote.h`, shared with the client game): `map <name>|next`, `bots <n>`, `spectate <player>`, `frag_limit <n>`, `time_limit <n>`.
- Published on `CS_VOTE` as `type\arg\yes\no\eligible\deadline\initiator`, empty when inactive; republished when the counts change. New manifest index in all three `g_types.h`.
- Rules: `g_vote` (1), `g_vote_time` (30 s), `g_vote_threshold` (0.5; 1 lets nothing pass), `g_vote_cooldown` (60 s per caller). Eligible = connected humans, spectators included. Passes when yes exceeds the threshold share of eligible voters, fails when it no longer can or at the deadline, and is cancelled when the intermission begins so it never queues a `map` beside `next_map`. With one eligible voter the caller's own yes passes at once — intended for a lone player on a bot server; the cooldown is the brake.
- Two new chainable hooks for modules with votes of their own: `PrepareVote` (validate and canonicalize) and `ApplyVote`.
- Players are matched by exact name: `G_ClientByName` picks the lexicographically nearest name, which for a punitive vote would be the wrong player. Force-spectate tosses inventory and respawns non-voluntarily (score kept, no "likes to watch"). `bots` is refused on maps that fix their own client count.

## Test plan

- [x] default / ctf / lithium build with no warnings; `verify_projects.py` 6/6 (new `g_vote.c`, `g_vote.h`, `bg_vote.h` registered in all three build systems)
- [x] `edge` 80 / 85 / 85 with bots under each module
- [ ] CI
- [ ] Casting a vote needs a connected client; the client half (screen + HUD) follows in the next PR and is where this gets exercised end to end

### DIFF
--- a/Quetoo.vs15/game-ctf.vcxproj.filters
+++ b/Quetoo.vs15/game-ctf.vcxproj.filters
@@ -54,6 +54,9 @@
     <ClInclude Include="..\src\game\common\bg_pmove.h">
       <Filter>src\common</Filter>
     </ClInclude>
+    <ClInclude Include="..\src\game\common\bg_vote.h">
+      <Filter>src\common</Filter>
+    </ClInclude>
     <ClInclude Include="..\src\game\common\bg_pmove_internal.h">
       <Filter>src\common</Filter>
     </ClInclude>
@@ -118,6 +121,9 @@
       <Filter>src\common</Filter>
     </ClInclude>
     <ClInclude Include="..\src\game\common\g_hook.h">
+      <Filter>src\common</Filter>
+    </ClInclude>
+    <ClInclude Include="..\src\game\common\g_vote.h">
       <Filter>src\common</Filter>
     </ClInclude>
     <ClInclude Include="..\src\game\common\bg_hook.h">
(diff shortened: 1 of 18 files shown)


jdolan:
Fixed: `g_vote.h` is now in the unconditional header group of `game_common.props` (it was mirrored from `g_hook.h`, which sits under `QuetooGameHook`), and added to `game.vcxproj.filters`. On the IDs: `D1A5B0000000000000000102` is 24 uppercase hex characters, the same width as the existing ones; the run of zeros just makes it look short.

PR #985, opened 2026-08-27
Add the client game's prediction and input hooks

jdolan:
## Summary

Second D11 batch on #963: the seams that retire the fork's shadows of `cg_predict.c` and `cg_input.c`.

- **Decisions:** `UsePrediction()` (tail = the former `Cg_UsePrediction`; a feature that cannot vouch for its physics answers false) and `Move(cmd)` (tail = the former `Cg_Move`; a feature adding an input reads its key and defers to previous). Both exported through wrappers, as `ClipEntity` is, because `Cg_LoadCgame` fills `cge` before `Cg_Module_Init` installs any chain.
- **Notifications:** `MoveCommandWillRun(pm, cmd)` / `MoveCommandDidRun(pm, cmd)` around `Pm_Move` for each pending command with time, and `PredictionDidComplete(pm)` after the loop. The move is set up once and carried through every command, so a training aid can install `pm->Accelerate` (#979) in `WillRun`.

Tails are the previous code paths or no-ops; default / ctf / lithium are unchanged.

## Test plan

- [x] all cgame modules build with no warnings
- [ ] CI

### DIFF
--- a/doc/game-module-hooks.md
+++ b/doc/game-module-hooks.md
@@ -353,6 +353,11 @@ and make the block additive instead.
 | `DrawHudElements` | `cg_hud.c` | ctf, techs |
 | `ListGameplayModes` | `cg_main.c` | ctf |
 | `ClipEntity` | `cg_predict.c` | — |
+| `UsePrediction` | `cg_predict.c` | — |
+| `Move` | `cg_input.c` | — |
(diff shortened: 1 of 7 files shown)


jdolan:
Docblock now names the missing delta frame among the cases the default refuses.

PR #990, opened 2026-08-27
Select the movement kernel per-player, through the parameters

jdolan:
Groundwork for #963, and it supersedes the `pm_move_t.Move` seam from #979 before race is built on it.

## What changes

`pm_params_t` gains a `kernel`, `Pm_Move` dispatches on it, and Quetoo's own movement moves into `bg_pmove_quetoo.c`. `bg_pmove.c` keeps only what every kernel shares — the trace, the touch list, the prologue, the spectator and frozen cases — and `bg_pmove_internal.h` becomes the contract between the two, down from 23 exported steps to six primitives plus the kernel entry points.

## Why a selection rather than a hook

**A kernel cannot be networked, but a selection can.** `pm_params_t` already travels per-player inside `pm_state_t`, delta-compressed as a unit, so the id arrives alongside the numbers it belongs to and client and server cannot disagree about which physics is running. A function pointer has to be installed on each side separately by two modules that might not agree — and the alternative was a ConfigString, which would have been per-server rather than per-player.

**Movement forks wholesale.** Porting Quake II's kernel against the steps `bg_pmove_internal.h` offered reused five of them and rewrote the rest. A seam made of steps earns very little; the honest seam is the whole kernel.

**A ruleset must be able to stop changing.** A record is comparable only to another record set under the same movement, and shared code that everyone edits can never offer that. A kernel in its own file is finished once it matches what it imitates. Changing what one does is a new id appended to `pm_kernel_t`, never an edit — the exception being `bg_pmove.c`, where a fault is a fault in every ruleset.

Being per-player is what makes this useful beyond race: a class-based mod gives each class its own movement by writing these fields in `G_PrepareMove`, which is what the comment already sitting in `g_client.c` anticipated.

## It costs nothing on the wire

The kernel byte goes in the two dead padding bytes after `gravity`, so `sizeof(pm_params_t)` is **unchanged at 104**, `gravity_water` stays at offset 4, and the 100-byte block the delta compares with `memcmp` stays free of padding. Appending the field instead would have grown the struct to 108 and put three padding bytes inside that block, which would make params look changed on uninitialized padding. It rides its own `PS_PM_KERNEL` delta bit, and `PM_KERNEL_QUETOO` is 0 so a zeroed `pm_move_t` still means Quetoo, as `pm->Move == NULL` used to.

`PROTOCOL_MINOR` goes to 1047 in all four modules.

## Verification

- **The extraction is verbatim, mechanically checked**: all 21 moved functions are byte-identical to their previous text apart from gaining `static`, and `Pm_QuetooMove` is the old tail of `Pm_Move` unchanged. So the 1052-line file needs no line-by-line reading.
- `edge` spawns **80 / 85 / 85** entities for default / ctf / lithium — the established baseline, unchanged — and 80 for race, which matches its item set. Each run loaded its module from the worktree via `-p`.
- Autotools: full build clean, no new warnings; `make check` 19/19.
- Xcode: `quetoo-all` builds through the workspace. `verify_projects.py`: 8 modules agree, each up one source.
- All four `Pm_Move` callers reviewed: the game and the bots both take params from the player state, so a bot moves with its player's kernel.

## Follow-ups this sets up

`PM_KERNEL_QUAKE` (NetQuake) and `PM_KERNEL_QUAKE2` land next as their own files, then race selects Q2 per map or vote. DeFRaG and DP2 are separate rulesets from each other — DP2 is the Quake II one the race fork needs — and both append later.


jdolan:
Thanks — addressed all three, plus a protocol fix from a separate review.

**Xcode object ids** — already fixed in a38a0bd, pushed before this review landed; every new id is now 24 characters and there are no duplicates.

**Net message coverage** — good catch, and worse than it looked: `Fill_TestParams` documents itself as setting *every* `pm_params_t` field, so the kernel was a hole in a contract the test already claimed. It now sets it, the round-trip asserts it explicitly (the existing block `memcmp` starts at `gravity_water`, so it could never have caught the kernel going missing), and a new case proves a kernel change alone is sent without dragging the whole parameter block with it. I mutation-tested it: dropping the read from `Net_ReadDeltaPlayerState` takes the suite to 18/19, so the coverage is load-bearing rather than decorative.

**Explicit layout guarantee** — agreed, and I went with asserting the two invariants that actually matter rather than a bare `sizeof`:

```c
_Static_assert(offsetof(pm_params_t, gravity_water) == sizeof(float), ...);
_Static_assert(sizeof(pm_params_t) - offsetof(pm_params_t, gravity_water) == 25 * sizeof(float), ...);
```

The first fails if `kernel` stops fitting in the padding `gravity` leaves, which would move the delta-compared region; the second fails if that region ever holds anything but the floats the encoder writes, which would make `memcmp` read padding and resend the block at random. Appending a field at the end trips the second; inserting one before `gravity_water` trips the first.

Also in this branch since your review: `PROTOCOL_MAJOR` 2030 to 2031, because `net_message.c` changed and that is what it guards (#976 set the precedent), and `PM_KERNEL_GAME` is gone — it promised a module-defined id range by analogy to `CS_GAME`/`EF_GAME`, but those are ranges the engine forwards without interpreting, whereas a kernel id has to resolve to code both the game and the client game hold. With no resolver, a module setting it would have silently run Quetoo's physics.

PR #991, opened 2026-08-27
Add QuakeWorld's movement, selected by g_movement

jdolan:
Builds on #990. Refs #963.

## QuakeWorld, not NetQuake

Ported from id's `QW/client/pmove.c` — which QuakeWorld already shared between its client and its server for exactly the reason we now do. The two Quakes differ, and it is QuakeWorld that people played and that bunny hopping belongs to; NetQuake would be a separate id.

**The bunny hop is not emulated, it falls out.** `Pm_QuakeAirAccelerate` caps the *wished* speed at 30 units but scales the acceleration by the **uncapped** one. A player already faster than 30 along the direction they're asking for gains nothing; one moving *across* it still has the full 30 available. Turning to hold that perpendicular is the technique, and it's why this is a skill rather than a button. Nothing special-cases it.

`check_pmove` measures **320 → 346.99** over two seconds of ideal strafing, against `sqrt(320² + 20×30²)` = 347.0 predicted. Aim 45° off instead and the dot product is 226, well past the cap, so nothing is gained — which is the mechanism, stated as a test.

**No ducking**, because Quake had none. The movement never raises `PMF_DUCKED` and the standing box is the only box.

## A movement is more than a kernel

QuakeWorld's algorithm run with Quetoo's numbers is not Quake — it needs 320 maxspeed, friction 4, stopspeed 100, a 32-unit box, STEPSIZE 18. Those live in `pm_params_t`, which the server fills. So each kernel exports the parameters that make it itself, and `bg_pmove.c` keys a table by `pm_movement_t`:

```c
[PM_MOVEMENT_QUETOO] = { .name = "quetoo", .label = "Quetoo",     .params = NULL },
[PM_MOVEMENT_QUAKE]  = { .name = "quake",  .label = "QuakeWorld", .params = &pm_quake_params },
```

Quetoo's carries no parameters, which is precisely what makes it the one that follows the server's cvars. Designated initializers keyed by the enum can't drift from the ids.

## Selection

`g_movement` takes the same precedence `gameplay` does: what the admin asked for, else the level's `maps.lst` metadata, else its worldspawn `movement` key, else Quetoo's — and publishes `g_movement_mode` for the server browser, as `g_gameplay_mode` does. It appears in Create Server beside Gameplay. **No config string is needed**: the movement reaches the client inside the per-player movement parameters, already networked and delta-compressed.

Also renames the selection from "kernel" to "movement" so the code and the touchpoints share one word. "Physics" was considered and rejected for over-selling it — this changes how a *player* moves, not how the world behaves. "Kernel" survives only in implementation prose, naming the function that implements a movement.

`PROTOCOL_MINOR` 1048; `PROTOCOL_MAJOR` unchanged, since no net message changed shape.

## Try it

`g_movement quake` on any map, or `{"movement" "quake"}` in a worldspawn, or pick QuakeWorld in Create Server.

## Verification

- **`check_pmove`, 7 cases, all passing**: the 320 cap against Quetoo's 300, the `+= 270` jump that adds rather than replaces, the refusal to duck, the absence of air friction, the bunny hop gain, Quetoo's movement unchanged, and that every movement answers to its own name while none answers to the reserved "default". Autotools-only, as `check_net_message` and `check_editor_map` are.
- Full autotools build clean, suite **20/20**.
- Xcode `quetoo-all` builds through the workspace; `verify_projects.py` 8 modules agree.
- Runtime: `g_movement` of `default`, `quake` and a bogus name all load, the bogus one falling back rather than failing; `edge` still spawns 80 / 85 / 85 for default / ctf / lithium and 80 for race.

Feel is the one thing a test can't judge — that part's yours.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ src/tests/check_http
 src/tests/check_master
 src/tests/check_mem
 src/tests/check_net_message
+src/tests/check_pmove
 src/tests/check_r_media
 src/tests/check_shared
 src/tests/check_thread
--- a/Quetoo.vs15/cgame-ctf.vcxproj
+++ b/Quetoo.vs15/cgame-ctf.vcxproj
@@ -27,6 +27,7 @@
     <ClCompile Include="..\src\game\ctf\bg_item.c" />
     <ClCompile Include="..\src\game\common\bg_pmove.c" />
     <ClCompile Include="..\src\game\common\bg_pmove_quetoo.c" />
+    <ClCompile Include="..\src\game\common\bg_pmove_quake.c" />
     <ClCompile Include="..\src\cgame\ctf\cg_module.c" />
   </ItemGroup>
   <PropertyGroup Label="Globals">
--- a/Quetoo.vs15/cgame-ctf.vcxproj.filters
+++ b/Quetoo.vs15/cgame-ctf.vcxproj.filters
@@ -512,6 +512,9 @@
     <ClCompile Include="..\src\game\common\bg_pmove_quetoo.c">
       <Filter>src\game\common</Filter>
     </ClCompile>
+    <ClCompile Include="..\src\game\common\bg_pmove_quake.c">
+      <Filter>src\game\common</Filter>
+    </ClCompile>
     <ClCompile Include="..\src\game\ctf\bg_item.c">
       <Filter>src\game\ctf</Filter>
     </ClCompile>
--- a/Quetoo.vs15/cgame-lithium.vcxproj
+++ b/Quetoo.vs15/cgame-lithium.vcxproj
@@ -27,6 +27,7 @@
     <ClCompile Include="..\src\game\lithium\bg_item.c" />
     <ClCompile Include="..\src\game\common\bg_pmove.c" />
     <ClCompile Include="..\src\game\common\bg_pmove_quetoo.c" />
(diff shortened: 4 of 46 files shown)


jdolan:
Thanks — all three addressed.

The GPL header typo ("General Above License") was already fixed in 17efc60, which landed after this review; it was in the first commit only. The `Pm_Debug` line and the two `shared.h` comments now say movement rather than kernel, so nothing user-facing or grep-facing uses the older word. "Kernel" survives deliberately in `bg_pmove_internal.h` and `doc/game-module-hooks.md`, where it names the *function* that implements a movement — a distinction no cvar, map key or config needs to know about, and one the plumbing/kernel contract is easier to describe with.

Since your review, this branch also took a round of fidelity corrections against id's actual QuakeWorld source: `Pm_QuakeFlyMove` was re-basing `original_velocity` inside the bump loop (which bled speed the multi-bump slide should give back), the edge-friction probe traced a point where upstream traces the player hull, and `Pm_QuakeNudgePosition` both quantized an origin upstream only appears to quantize and probed with a function that jitters by a whole unit — eight times the precision it was searching for. It also now honours `PMF_TIME_PUSHED` and `PMF_TIME_TELEPORT`, records `pm->step` so stairs do not snap the camera, and declines the grappling hook through `G_AllowHook` rather than handing it to a kernel that would ignore `hook_length`.

PR #992, opened 2026-08-27
Add Quake II's movement, and the race mod's

jdolan:
Builds on #991. Refs #963. Two movements, because the second grows directly from the first.

## Quake II

Ported from id's `qcommon/pmove.c`. Vanilla Quake II, with `pm_airaccelerate` at zero, so the airborne case accelerates plainly at 1.

**I had the air control backwards at first.** My docblock said "there is no strafe jumping here"; the test I wrote to prove it failed, correctly. **Quake II strafe jumping is real** — with no cap on the wished speed the headroom is the whole of it, so a command earns its acceleration over a far *wider* range of angles than QuakeWorld's 30-unit window allows. The test now asserts it from both sides:

| 45° off the velocity | gain |
| --- | --- |
| QuakeWorld | nothing — past its wish cap |
| Quake II | a good deal — its wish is uncapped |

What else separates them: overbounce 1.01 and clipping against the velocity *as it stands*; a quarter-unit ground probe; a jump that adds then insists on at least the jump speed; a lockout after a hard landing; ducking; ladders; and origin *and* velocity quantized to eighths on the way out.

**Two engine differences bit, both the same shape.** `Pm_Init` clears `PMF_ON_GROUND` every move, where Quake II keeps it — so its "were we already grounded" edge was always true, which re-armed the landing lockout every frame instead of detecting a landing, and its ducking (ground-only) never engaged at all. Both now read `pm->ground.ent`, which is seeded from the entity each move and survives `Pm_Init`. Same class as the `PMF_TIME_PUSHED` gap #991's review found. Tests caught both; reading didn't.

## Race

The one movement here that **imitates nothing**, and the naming reflects that. It began from Quake II because the ramp jumping racing is built on reached us through Digital Paint: Paintball 2 — a Quake II engine mod — but Digital Paint is where these mechanics were *found*, not a standard they're held to. This is Quetoo's own racing movement, and it is meant to be tuned.

That makes it the one file here that isn't frozen, and the exception is deliberate: **a movement that imitates a game is finished when it matches; a movement we own is finished when it is fun.** Racing is entirely about movement, so this is where that work happens. The constants are named for what they do rather than where they came from, because they are the tuning surface.

What makes ramps worth racing on:

- Rising off a slope doesn't take the ground away **unless the jump is held**, so letting go keeps contact and keeps vertical speed.
- While that contact holds, the ground accelerates **like air**, so speed carries up and off a slope instead of being scrubbed — and asking for a direction away from the one already travelled gives the slide up, which is ordinary ground again.
- A jump **adds** to the vertical speed present, up to a ceiling of 450, so jumps stack off ramps and lifts.
- Upward speed below 150 is absorbed, above it kept: bumps don't launch, ramps do.
- A step with nowhere to settle retries with a narrower box, so a ramp lip can't swallow a fast move.

No versioning scheme, deliberately — there are no records to protect yet.

## Verification

- `check_pmove`, **15 cases**, all passing. New: Quake II's 300 against QuakeWorld's 320; the jump floor from both the rising and falling side; the air-control angle comparison; ducking to a 4-unit box and a -2 eye and standing back up; the landing lockout refusing a jump; racing keeping the ground at +300 where Quake II loses it; and a racing jump reaching the ceiling exactly and adding nothing past it.
- Full autotools build clean, suite 20/20; Xcode `quetoo-all` through the workspace; `verify_projects.py` 8 modules agree.
- Runtime: `g_movement race` and `quake2` both announce and resolve; a bogus name warns and falls back.

`bg_pmove_internal.h` becomes `bg_pmove_local.h`, matching `g_local.h` and `cg_local.h`. `PROTOCOL_MINOR` 1050.

Quake III is next, as its own PR.

### DIFF
--- a/Quetoo.vs15/cgame-ctf.vcxproj
+++ b/Quetoo.vs15/cgame-ctf.vcxproj
@@ -29,6 +29,7 @@
     <ClCompile Include="..\src\game\common\bg_pmove_quetoo.c" />
     <ClCompile Include="..\src\game\common\bg_pmove_quake.c" />
     <ClCompile Include="..\src\game\common\bg_pmove_quake2.c" />
+    <ClCompile Include="..\src\game\common\bg_pmove_race.c" />
     <ClCompile Include="..\src\cgame\ctf\cg_module.c" />
   </ItemGroup>
   <PropertyGroup Label="Globals">
--- a/Quetoo.vs15/cgame-ctf.vcxproj
+++ b/Quetoo.vs15/cgame-ctf.vcxproj
@@ -28,6 +28,7 @@
     <ClCompile Include="..\src\game\common\bg_pmove.c" />
     <ClCompile Include="..\src\game\common\bg_pmove_quetoo.c" />
     <ClCompile Include="..\src\game\common\bg_pmove_quake.c" />
+    <ClCompile Include="..\src\game\common\bg_pmove_quake2.c" />
     <ClCompile Include="..\src\cgame\ctf\cg_module.c" />
   </ItemGroup>
   <PropertyGroup Label="Globals">
--- a/Quetoo.vs15/cgame-ctf.vcxproj.filters
+++ b/Quetoo.vs15/cgame-ctf.vcxproj.filters
@@ -285,7 +285,7 @@
     <ClInclude Include="..\src\game\common\bg_pmove.h">
       <Filter>src\game\common</Filter>
     </ClInclude>
-    <ClInclude Include="..\src\game\common\bg_pmove_internal.h">
+    <ClInclude Include="..\src\game\common\bg_pmove_local.h">
       <Filter>src\game\common</Filter>
     </ClInclude>
     <ClInclude Include="..\src\game\ctf\bg_item.h">
@@ -513,6 +513,9 @@
       <Filter>src\game\common</Filter>
     </ClCompile>
     <ClCompile Include="..\src\game\common\bg_pmove_quake.c">
+      <Filter>src\game\common</Filter>
+    </ClCompile>
+    <ClCompile Include="..\src\game\common\bg_pmove_quake2.c">
       <Filter>src\game\common</Filter>
     </ClCompile>
     <ClCompile Include="..\src\game\ctf\bg_item.c">
--- a/Quetoo.vs15/cgame-ctf.vcxproj.filters
+++ b/Quetoo.vs15/cgame-ctf.vcxproj.filters
@@ -518,6 +518,9 @@
     <ClCompile Include="..\src\game\common\bg_pmove_quake2.c">
       <Filter>src\game\common</Filter>
     </ClCompile>
+    <ClCompile Include="..\src\game\common\bg_pmove_race.c">
+      <Filter>src\game\common</Filter>
+    </ClCompile>
     <ClCompile Include="..\src\game\ctf\bg_item.c">
       <Filter>src\game\ctf</Filter>
     </ClCompile>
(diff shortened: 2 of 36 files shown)


jdolan:
Both fixed, and the first was worse than flagged.

**The label.** Not a deliberate shortening — `main` has `"QuakeWorld"`, and my Quake II commit regressed it to `"Quake"`. Restored, and the point stands: this movement is QuakeWorld's rather than NetQuake's, and that distinction is exactly what a user-facing label should carry, since NetQuake would be a separate movement if it ever lands.

**The comment.** A leftover from before I corrected myself. I originally believed vanilla Quake II had no strafe jumping, wrote that in both the docblock and this comment, and then a test I wrote to prove it failed — correctly. With no cap on the wished speed the headroom is the whole of it, so Quake II earns acceleration over a *wider* arc than QuakeWorld's 30-unit window allows. I fixed the docblock and added tests asserting the difference from both sides, but missed this one. It now says what the code does.

An independent read-only review of both kernels against id's `qcommon/pmove.c` is running; I will hold the merge for it, since the equivalent review on #991 found a fidelity divergence in the QuakeWorld slide that mattered.

jdolan:
An independent review against id's `qcommon/pmove.c` found **the Quake II port faithful** and the race kernel carrying five faults — three mine, two inherited from the fork it was harvested from. All acted on.

**The abandoned-slide branch was not the ordinary ground path it claimed to be.** It ran friction a second time, gave up the vertical speed differently in *both* signs of gravity, and applied gravity and acceleration in the other order — so giving up a slide behaved unlike never having slid. There is one ground path now; the slide only changes the acceleration and whether the climb is kept. The fork solved this by delegating; I had inlined a copy and let it drift.

**Ground friction now stays off while the slide holds.** It applies to all three axes, so leaving it on scrubbed the very climb the slide exists to carry. The fork excludes it for exactly this reason and I missed that. A test caught it: reverting the exemption takes a stacked jump from 420 to 120.

**A jump above the ceiling took the ground away and raised `PMF_JUMPED` before deciding to do nothing** — costing the player the slope contact it was refusing to add to, and sounding a jump that never happened.

**`PM_RACE_BUMP_SPEED` and the inset step retry are deleted rather than fixed.** The first only bit below ~36 fps: the slide latch above it and the ordinary ground path either side left no window, and what the docblock credited to it actually comes from the latch. The second could not fire in the case it documented — a completed slide never starts embedded, which is what `all_solid` means — and would have placed the player where only the *narrower* box fits if it had. Both were inherited; the docblock now says what the code does.

**Four fit-tests used `Pm_Trace`,** which jitters the start by up to a unit to escape a solid and so cannot answer whether a box fits where it is. Two were step-up guards, two were stand-up-under-a-ceiling tests, where a false negative puts the player inside geometry. `bg_pmove_quake.c` already documents this hazard; these now use `pm->Trace` as it does.

**Quake II's swim-up speed is back to exact equality on the water type.** Bit-testing it was an *improvement* — and vanilla falls through to the slowest speed for a water brush carrying a current bit, which is the movement, quirk and all.

**Two tests passed for the wrong reasons.** The ceiling assertion held even with the ceiling test deleted, because the arithmetic alone still lands on 450; it asserts the value now. The ground-retention test asserted a flag the second classification overwrites, and its premise was wrong twice: a player rising at 300 leaves a flat floor mid-move, and a player asking for nothing gives the slide up by design.

**One honest gap, worth knowing before merge.** That test is gone rather than fixed, because with friction correctly exempted the retention has *no consequence a flat floor can show*. The jump-ceiling case is the only execution-level evidence the slope path works at all — 420 is reachable only if the ground was kept and friction exempted. Whether the mechanic is worth having is a question for a real ramp, not for this harness.

Also: the movement source lists are alphabetical in all eight modules (thanks for starting that), the race kernel's box and eye are now asserted, and the retracted strafe-jumping sentence is gone from the race file too. The pbxproj reorder that had been swept into the previous commit is dropped — it was a pure no-op, verified identical when line-sorted — which needed a force-push with lease.

14 pmove cases green, suite 20/20, both build systems clean, `verify_projects.py` 8/8.

PR #1005, opened 2026-08-28
Keep the player model preview out of the world's voxel lights

jdolan:
Opening the player model view in-game froze the screen: the model drew once and
the rest of the frame turned to noise, in `default` as well as `race`.

**Cause.** The mesh vertex shader's `vertex_lighting` (`light.glsl:247`) walks
the world's voxel light lists for every view, with no view-type gate (the
fragment shader has one). For the preview, `R_DrawMeshEntities` bound the BSP
lights block as a stand-in for the voxel light-data and light-index storage
buffers, and `R_UpdateUniforms` still described the loaded world's voxel grid,
so the shader computed a real voxel index into a small buffer holding something
else and read out of bounds. Metal faulted the command buffer, which then never
presented - hence the freeze - and API validation cannot see OOB storage reads,
so the log was clean. At the main menu there is no world, so nothing was walked.

**Fix.** A view that is not `VIEW_MAIN` gets a 1x1x1 voxel grid in its uniforms,
so every lookup lands on index 0 (a zero-sized grid would hit `clamp()` with an
inverted range, which is undefined), and the two voxel-light slots bind a new
zeroed `voxel_lights_fallback` buffer that reads as "no lights" - the same
pattern as the existing caustics, occlusion and sky fallbacks.

Found by bisecting `R_DrawPlayerModelView` with temporary switches: the pass
without meshes was clean; the pass with meshes but stale uniforms still faulted.
Verified in play: model renders, world stays intact.

### DIFF
--- a/src/client/renderer/r_main.c
+++ b/src/client/renderer/r_main.c
@@ -132,7 +132,13 @@ void R_UpdateUniforms(const r_view_t *view) {
     out->editor = editor->integer;
     out->developer = developer->integer;
 
-    if (r_models.world) {
+    // a view that is not the main view - the player model preview - has no
+    // relation to the world's lighting, and its lookups must all land on the
+    // one voxel of the fallback buffers: clamping to a zero-sized grid would
+    // not, since clamp() with a low bound above its high bound is undefined
+    if (view->type != VIEW_MAIN || !r_models.world) {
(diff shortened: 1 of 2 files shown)


jdolan:
Pushed straight to main as requested.

PR #1009, opened 2026-08-30
Tighten the hook APIs from the #963 sweep, part 1

jdolan:
Items 1-5 of the review of everything that changed in pre-existing `.c`/`.h` files since #975 (#963). One commit per item.

1. **`pm_move_t.Accelerate` removed.** A sampling callback threaded through all five kernels for a training aid that does not exist; nothing assigned it.
2. **`ClipEntity` is opt-in.** The chain has no tail: `G_ClipEntity`/`Cg_ClipEntity` are NULL until a module installs a link, `G_Init`/`Cg_Init` export whatever is installed after module init, and the engine guards the pointer - so Default/CTF/Lithium no longer pay a cross-module call per entity per trace. The export trampolines go away; the server asks after the hull check.
3. **`cgi.Clip`** (`Cl_Clip`, reciprocal of `Sv_Clip`), and prediction traces on behalf of `cgi.client->entity`. The race client's one-way-wall probe becomes the one-liner the server has, the re-entrancy sentinel from #1006 is gone, and the client can mirror the server's "no mover, so solid" rule. `CGAME_API_VERSION` 37.
4. **`ModifyDamage`** is back to `(target, attacker, damage*, knockback*)` returning `bool`. Call ordering in `G_Damage` unchanged.
5. **Level gravity reaches every movement.** `G_MovementParams` branches first (Quetoo from cvars, the rest from their params) and applies the level's gravity over the movement's. `g_level.gravity == 0` now means unset (it was forced to 800), so the movement's own applies unless `g_gravity`, map metadata or worldspawn set one; `G_LevelGravity()` for readers outside pmove. A worldspawn `gravity 0` is warned about, since zero is now "unset".

`ClientDidChat` stays, per discussion - note it is a *Did* hook, so muzzling would need a `ClientWillChat` returning `bool`.

Verified: full build clean; `make check` 20/20 (`--with-tests`); `racetest` under race and `edge` under default load headlessly with the established entity counts (10 / 80). Read-only review pass applied (zero-gravity sentinel guard, NULL-mover recursion guard, orphaned comment). Not lap-tested with a client.

### DIFF
--- a/src/cgame/cgame.h
+++ b/src/cgame/cgame.h
@@ -36,7 +36,7 @@
 #include <Objectively/RESTClient.h>
 #include <Objectively/Vector.h>
 
-#define CGAME_API_VERSION 36
+#define CGAME_API_VERSION 37
 
 /**
  * @brief The client game import struct imports engine functionailty to the client game.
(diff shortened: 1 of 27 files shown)


jdolan:
On the two Copilot notes: `g_level.movement` is only ever assigned from `G_ResolveMovement`/`G_CoerceMovement`, which reject unknown ids, so `Pm_Movement` cannot return NULL there - the same unchecked contract `g_entity.c` already relies on. And the only `Cl_Trace` caller outside prediction is `cl_screen.c`, reachable only under `CL_ACTIVE`, where `cls.cgame` is already dereferenced unguarded (`cl_screen.c:457`, `cl_entity.c:462`).

PR #1015, opened 2026-08-30
Tighten the hook APIs from the #963 sweep, part 4

jdolan:
The deferred decisions from the review, as settled; follows #1011 and #1014 (whose commit this branch carries until it merges). One commit per item.

- **`InitEntity`**: the hook is named for what it does; `G_SpawnEntity` is the allocator again (`G_SpawnEntityDef` gone).
- **`G_ResolveGameplay`** mirrors `G_ResolveMovement`: the level's choice is remembered, `g_gameplay` wins when named, and "default" defers to the level - including at runtime, where it used to mean deathmatch. `g_gameplay_mode` is published from the resolver, so runtime changes reach the browser. An out-of-range metadata id now resolves to deathmatch instead of being stored raw.
- **Movement announced on begin**, with the gameplay, instead of a broadcast during spawn that nobody was present for. Note: the welcome (`Welcome to`, MOTD, gameplay, teams) was built and discarded under `#if 0` all along; it is now sent with `gi.ClientPrint`. That is a visible change for every module - shout if you'd rather keep it silent.
- **`ClientWillChat(cl, text, size, team)`**: false muzzles, or rewrite `text` in place; `ClientDidChat` stays. Per your Will/Did rule, the hooks still missing a counterpart are `ClientDidBegin`, `ClientDidChangeUserInfo`, `ClientWillDisconnect`, `LevelWillSpawn`, `FrameDidEnd` - I have not added those speculatively; say if you want them in one go.
- **One voxel-light fallback buffer** in `r_lights`, bound at all five sites (mesh, BSP x2, decal, sprite).

Verified: all modules incl. race build clean; `make check` 20/20; `racetest` and `edge` load headlessly. Read-only review applied (it caught the dead welcome). Renderer change not run under a renderer.

### DIFF
--- a/src/client/renderer/r_bsp_draw.c
+++ b/src/client/renderer/r_bsp_draw.c
@@ -532,7 +532,7 @@ void R_DrawOpaqueBspEntities(const r_view_t *view, RenderPass *pass) {
     r_lights.bsp_buffer->buffer,
     r_lights.dynamic_buffer->buffer,
     bsp->voxels.light_data_buffer->buffer,
-    bsp->voxels.light_indices_buffer ? bsp->voxels.light_indices_buffer->buffer : r_lights.bsp_buffer->buffer,
+    bsp->voxels.light_indices_buffer ? bsp->voxels.light_indices_buffer->buffer : r_lights.voxel_fallback_buffer->buffer,
   };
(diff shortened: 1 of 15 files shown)


jdolan:
Three of Copilot's notes were against the pre-fix #1014 commit this branch carried; rebased onto main, they are gone. On `%g`: the engine never calls `setlocale`, so `printf`/`sscanf` run in the C locale on both ends and the field is locale-stable; the same holds for every other float the game formats into a string.

PR #1021, opened 2026-08-31
Show the run's milestones on the HUD, against the bests

jdolan:
Step 7d of #963, and the `ClipEntity` typedef collision fix (item 6), one commit each.

- **Milestones**: `G_Race_Checkpoint`/`Split`/`Stage` send `SV_CMD_RACE_MILESTONE [kind][number][label][time][vs best][vs record]` (unicast reliable) instead of centre-printing. Deltas come from the racer's own PB and the course record (`G_Race_BestRecord`, extracted from `G_Race_SubmitRecord`); `RACE_MILESTONE_NO_DELTA` for no reference. The HUD draws the latest for 3 s under the run timer: name and time, then the signed comparisons - green faster, red slower - with the "best" row folded away when it equals the "record" row. A run going idle clears it.
- Behaviour deliberately dropped: the split centre print's `(+delta vs previous split)`; comparisons are now against references, not the run itself.
- **`ClipClientEntity`**: the cgame's clip typedef no longer collides with the game's `ClipEntity`; hooks keep their names, no ABI change.

Verified: full build (race included, now via `make`) clean; `make check` 20/20; `racetest` loads headlessly. Read-only review applied (index math against both record formats verified; zero-delta sign; stale milestone across runs). Wants a lap: run `racetest` twice - the second run should show deltas at each checkpoint once a record exists.

### DIFF
--- a/src/cgame/common/cg_module.h
+++ b/src/cgame/common/cg_module.h
@@ -155,14 +155,15 @@ extern ListGameplayModes Cg_ListGameplayModes;
  * @brief Decides whether `ent` clips a trace made on behalf of `mover`, after the
  * client has applied its own skip rules. There is no tail: the chain is `NULL`
  * until a feature installs a link, and the client does not call an empty one.
- * @details Chainable, and the reciprocal of the game's `ClipEntity`: a feature
+ * @details Chainable, and the reciprocal of the game's `ClipEntity`, under a
+ * typedef of its own so that no one translation unit may ever meet both: a feature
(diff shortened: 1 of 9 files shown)


jdolan:
On Copilot's `PROTOCOL_MINOR` note: wire changes on `main` deliberately don't bump `PROTOCOL_*` until the v1.0.85 tag, per standing policy - the whole 1.0.85 window is one protocol break, and `SV_CMD_RACE_BARRIERS` set the same precedent.

PR #1036, opened 2026-09-11
Add trigger_portal: seamless brush portals with a live view through them

skies912:
## Summary
- New `trigger_portal` brush entity: continuously carries anything crossing it to its paired portal with no teleport sound, effects, or snap. One face textured `common/portal` is authoritative — quemap bakes its centroid and full facing (`portal_origin`/`portal_angles`, including roll taken from the way the texture reads) so the entity's own angles never rotate the brush, and floor/ceiling faces orient correctly, not just walls. Position, velocity, and view/facing are carried via a relative-basis transform between the two faces, so the pair may be oriented any way; one-way pairs work by omitting the destination's `target`. A `strict_displace` spawnflag skips the facing math for a literal hole in space.
- The `common/portal` face is compiled in (`SURF_PORTAL`) and, with `cg_portals 1`, shows the view from the paired portal: cgame registers each portal at map load, carries the main camera through it every frame, renders the far side into a per-portal framebuffer (`cg_portal_scale`), and the BSP shader samples it in place of the face. Frustum culled, clipped to the far face's plane, one level deep (a portal seen through a portal shows its plain material). While the eye is inside a portal's own recess, the main view is clipped to that portal's face too, so the wall the recess sits inside doesn't show past its edges — a void instead. Lights, sprites and beams are shared with portal views, and a portal view looking back at the local player draws them in third person. **Off by default** — cost is one scene render per visible portal.
- A player transits once the leading edge of their body reaches the face — front first into a wall while moving into it, feet or head first into a floor or ceiling (so stepping onto a floor counts, and a ceiling transits before your eyes reach it) — and arrives as far into the far portal as they were into the near one, clamped to its recess, then simply walks on out: nothing skips, and a fresh arrival moving out of the far face can't be taken back through it. Client-side prediction mirrors this exactly (including standing down for spectators/noclip/dead/frozen), and the transit carries the outgoing player state along with the entity, so there's no client/server disagreement to smooth over on the way through.
- Players mid-crossing (within `PORTAL_PLAYER_CLEARANCE` of a portal's volume) pass through each other, so two players crossing at once don't collide; AI keeps its goals through this rather than treating it as death.
- Projectiles are carried by a swept check inside the fly move, so even a hyperblaster bolt crossing the volume in one tick transits at the face; hitscan (bullets, rail, lightning) is traced through portals leg by leg, with the rail drawn per leg and lightning's far leg as its own beam; rail impacts/decals don't land on the portal face itself.

## Known limitation
- Floor/ceiling and wall portal volumes need to be at least 32 units deep along their own normal, as documented in the `trigger_portal` QUAKED comment. Under that, the arrival math has no room to hold a full player, and ordinary client/server timing differences (normally absorbed by the recess) become visible as a position snap. Tested down to 4u to confirm this is really a hard floor and not a fixable-in-place clamp; leaving it documented rather than guessed at further.

## Notes
- Maps with a `trigger_portal` must be recompiled: the face is now emitted, and the facing key is `portal_angles`.
- The TrenchBroom FGD entry (spawnflag + descriptions) was updated in the deployed game config, not in this repo.
- A Portal-style duplicate of the local player straddling the far portal while crossing was tried and backed out for now; the view-side plumbing it needs is still in place.
- History includes two `TEMP:` logging commits from debugging that were reverted by a later commit — happy to squash on merge.

## Test plan
- [x] Two-way pair on opposite walls: walk through both directions, exit position/facing/velocity continuous, no double transit, never stranded behind a face, no snap
- [x] Floor/ceiling pair: step/drop onto a floor, head into a ceiling, momentum preserved across the hop in either texture orientation; infinite-fall loop tested to terminal velocity, including fall damage on exit
- [x] `strict_displace` pair: pure displacement, no re-trigger on arrival
- [x] Two players crossing the same portal simultaneously: no collision, no stall
- [x] `cg_portals 0`: face shows the plain `common/portal` material, zero rendering cost
- [x] `cg_portals 1`: live view with correct parallax while strafing; no hall-of-mirrors when a portal is visible through a portal; void when eye is inside the recess
- [x] Rockets, grenades, hyperblaster (all angles), shotgun/machinegun, railgun (incl. pierce), lightning all carry through; rail/tracers/lightning drawn per leg
- [x] Noclip/spectator through a portal: no prediction fight, flies through cleanly

Maximum sexiness: 
portals


jdolan:
Closing in favor of #1041

Issue #1040, opened 2026-09-14
Render the world through misc_portal surfaces


jdolan:
## Context

A `misc_portal` is a brush surface that shows the world as seen from a paired portal elsewhere in
the map — you can look through it before deciding to walk through it.

The `feature/misc-portal` branch explored this and established that the approach works. Thanks to
@Skies912 for the prototype; it settled several design questions that would otherwise have needed to
be worked out from scratch, and this plan reuses its conclusions. What follows is a
focused, visual-only implementation scoped to the rendering feature alone, with teleport gameplay
left for a later ticket. The prototype's rendering slice is roughly 700 lines, so that is about the
size we should expect here.

### Techniques the prototype established

- **Screen-space sampling of the portal texture.** The portal view is rendered with the identical
  projection from the main camera transformed through the portal, so the destination image lines up
  1:1 in screen space and the fragment shader can sample it at `gl_FragCoord.xy / viewport.zw`.

  This is worth stating explicitly because there is a tempting shortcut that does *not* work: giving
  the portal a material stage and pointing its `r_image_t->texture` at a framebuffer's
  `resolveColorTexture`. That samples with the surface's own UVs and reads as a photo glued to the
  wall, with no parallax. A `SURF_PORTAL` bit, a dedicated sampler, and a shader branch are
  genuinely required.
- **A `clip_plane` uniform**, to cut away geometry behind the destination plane. This is the
  standard companion to the technique above.
- **A dedicated depth-only fragment shader for occlusion queries**, so the clip-plane discard does
  not leak into the query pass.
- **Occlusion queries disabled inside portal views.** Queries belong to the main camera; their
  results are not valid from the portal camera's position.
- **Baking the portal face's frame in quemap.** The frame must not be the entity's own `origin` and
  `angles`, since those would rotate the brush model itself — for rendering and collision alike —
  out from under the mapper. Full pitch and yaw are baked, not yaw alone, so a portal on a floor,
  ceiling or slope orients correctly.

## Design

**`misc_portal` is a purely client-side visual entity.** It is a brush entity with exactly one face
textured `common/portal`, targeting another `misc_portal`. It does not move anyone. A mapper who
wants the portal to be navigable overlays a `trigger_teleporter` *behind* the portal face, so the
player crosses the visual plane before the trigger fires.

This keeps the rendering mechanism and the gameplay decision independent, leaves today's teleporters
untouched, and makes window-only portals — a view into a room you cannot reach — a first-class use.
[...]

PR #1041, opened 2026-09-15
Render the world through misc_portal faces

jdolan:
Implements #1040: a brush entity with a `SURF_PORTAL` face and a `portal` key shows the world as
seen from the point it names, so a player can look through a teleporter before committing to it.

Thanks to @Skies912, whose `feature/misc-portal` prototype established the approach — the
screen-space sampling, the occlusion-query handling and the compiler-side face baking all came
out of that work.

> [!IMPORTANT]
> **Merging this requires rebaking every map.** `BSP_VERSION` goes to 82 for the new portals
> lump, and quetoo-data needs the matching `entities.def`, `Quetoo.fgd` and texture work.

## How it fits together

- **quemap** tags `common/portal` sides with `SURF_PORTAL` and writes a new `BSP_LUMP_PORTALS`
  (`BSP_VERSION` 82). Each portal records its brush side, its draw elements, and the entry and
  exit frames as origin/forward/up pairs — taken from the face and from the entity named by the
  `portal` key, never from the entity's own origin and angles, which must stay free to position
  its brushwork. A `SURF_PORTAL` face is never merged with another, so each portal keeps its own
  draw elements.
- **The viewpoint is named by `portal`, not `target`.** A portal face may be put on any brush
  entity, and `target` is already spoken for: a `func_train` reads it as the first `path_corner`
  of its route and reassigns it while walking, a `func_button` as whatever it fires.
- **`misc_portal`** is a real brush entity with a spawn function, so the server links its model
  and the client draws it like any other inline model.
- **The renderer** resolves portals when it loads the world, the way it resolves BSP lights.
  `R_LoadBspPortals` builds each portal's two frames and attaches it to its draw elements; the
  transform between them is composed per frame, since the model drawing the face may be a mover.
- **cgame supplies only the model matrix** of the entity drawing each portal's face — the one
  part the renderer cannot resolve, since it needs the client entity list. `R_AddPortal` does the
  rest. A portal face's frame is baked in the space of the model that draws it, so a portal may
  ride a `func_bob` or `func_rotating` and its face and camera follow the same transform.
- **The camera is anchored to the portal's plane** rather than carried in full: the player's eye
  is clamped to the face and flattened onto its plane before being carried. At the destination it
  then sits on the exit plane, so nothing behind the exit is ever in front of it and the near
  plane does the work an oblique clip plane would have. A deliberate trade — see *Notes*.
- **Ordering and placement are the renderer's.** cgame offers every portal of the world;
  `R_AddPortal` skips the ones facing away, keeps the rest distance-sorted, evicts the farthest
  when full, and places the view itself. Nothing about a portal view is a caller's to decide.
- **Scene population** is repeated inside the renderer: entities, lights, sprites and beams added
  to the main view are copied into each portal view that survives culling, which culls them for
  itself. Cgame adds, renderer culls.
- **`R_DrawPortals`** draws each surviving portal into its own layer of one array texture before
  the main view, which samples it in screen space — both use the same projection, so the images
  coincide.
- **Material stages** compose the look. In a material declaring `surface portal`, a stage naming
  the material's own diffusemap samples the portal instead of that texture, so `warp`, `color`
  and `blend` dress it up; with `r_portals 0` the same material still draws, sampling the texture
  it names.

## Notes for review

- **This is a magic window, not a hole in space.** Anchoring the camera to the plane costs depth
  parallax: walking up to a portal no longer opens the view. Clamping the flattened eye to the
  face keeps lateral parallax, bounded by the face's own width, which is the part that sells the
  effect. The alternative — a true carry — puts the camera inside the destination's brushwork for
  any free-standing portal, which is how this was found.
- A portal the camera stands behind is skipped before it is offered a view. The face is single
  sided and every BSP pipeline culls back faces, so there is nothing to draw; the test is against
  the camera's position, never the direction it faces.
- Each portal's pass is **scissored to the face's own screen rect**. The face samples at
  `gl_FragCoord`, so those are the only texels that can ever be read, and the saving scales with
  distance without any of the resolution stepping, and its pop, that per-portal sizing would
  bring.
- Portal views render at half the window's resolution. The projection comes from the view's own
  viewport, not the framebuffer, so the image still registers with the face.
- Recursion is blocked in two places: a portal view offers no portals of its own, and
  `R_PushBspPortalLayer` pushes `-1` for any portal view, so a portal seen through a portal falls
  back to its plain material rather than sampling a layer it is not bound to.
- Portal views use frustum culling rather than occlusion queries, whose results belong to the
  main camera. This applies to lights too.
- Decals are deliberately not repeated into portal views: `R_UpdateDecals` clips them into the
  shared, persistent geometry of the blocks they land on, so repeating them clipped each decal
  once per portal and drew it that many times over.
- Sprites draw in portal views but are not softened — softening blends against a double-buffered
  copy of the view's own depth, and one copy cannot serve several portals in a frame.
- Portal-stage `warp` is in surface units, not screen units, so amplitude does not change with
  aspect ratio. Ordinary stage `warp` is unchanged.
- Requires ObjectivelyGPU ≥ the layered-attachment change (`Framebuffer::colorTargetInfoForLayer`).
- `r_portals` (renderer, default on) skips drawing portals. The cvar is deliberately in the
  renderer and on by default: cgame always offers portals, and a map looks wrong without them, so
  this is a performance escape hatch rather than an opt-in.

## Changes beyond portals

- **`CGAME_API_VERSION` 49.** `cgi.AddPortal` returns nothing, and `cgi.CulludeBox` /
  `cgi.CulludeSphere` are gone. The latter pair advertised occlusion testing they could not
  perform: cgame populates the scene alongside `R_DrawViewDepth`, which is what resolves the
  frame's queries, so any answer would have been the previous frame's — and they were wired to
  `R_CullBox` / `R_CullSphere` accordingly.
- **`misc_dust`, `misc_flame` and `misc_weather` now emit regardless of the main view.** Each
  skipped emitting entirely when its volume was off screen, so a torch behind the player emitted
  nothing and a portal looking at it had nothing to draw. Visibility is not an emitter's to
  decide, since it cannot know who is looking.

## Not in scope

Teleport gameplay of any kind, portal recursion, drawing the local player's own model in a portal
view, and any change to `misc_teleporter` / `trigger_teleporter`. A mapper who wants a portal to
be navigable overlays a `trigger_teleporter` behind the face.

Oblique near-plane clipping was implemented and then removed. Anchoring the camera to the exit
plane achieves the same cut for free: anything behind the exit is behind the camera.

## Known gaps

- Portal views sample the previous frame's shadow atlas: `R_DrawShadows` runs inside
  `R_DrawMainView`, after the portal passes. Accepted — the atlas is hash-cached and portal light
  sources are static, so only a caster that moved this frame lags, and only by one frame.
- `scroll`, `scale`, `rotate` and `stretch` do not reach a portal stage — they are applied to
  `vertex.diffusemap` in the vertex shader, and the portal path uses screen coordinates. `warp`
  and `color` do reach it.
- `misc_portal` needs its quetoo-data definitions landed, and TrenchBroom's own copy of
  `Quetoo.fgd` synced separately.
- `MAX_PORTALS` is 8 per frame out of `MAX_BSP_PORTALS` 64 per map; the renderer keeps the
  nearest 8 offered and draws those it can see.
- Portals are culled by frustum and facing only. Occlusion queries are not yet functional in
  SDL_gpu, so a portal facing the camera from across the map still renders even when a wall
  stands between them; `R_CulludeBox` already dispatches to the query path and will benefit as
  soon as queries land.
- A portal on a brush entity the server has not sent — `SVF_NO_CLIENT`, as a `func_wall` spawned
  in the off state is, or one not yet `in_use` — is skipped rather than guessed at, since its
  geometry may be origin-brush relative. The server does no PVS culling, so a visible entity is
  always sent.
- Sprites are not culled per view, so every sprite is assembled and uploaded once per drawn
  portal. Measured at ~1.9ms for 13K sprites across five views on `claustrophobopolis`; a
  frustum test in `R_UpdateSprite` would recover most of it.

## Test plan

- [x] Builds clean on autotools, Xcode and MSVC, zero warnings
- [x] quemap bakes the lump correctly, verified by parsing a compiled BSP
- [x] Renders in-game on a real map
- [x] `r_portals 0` leaves portal faces on their material, sampling its own texture
- [x] Two portals visible at once, each showing its own destination (the layer-index test)
- [x] A portal on a `func_bob` tracks it, face and camera glued at the seam
- [x] Two portals sharing one target
- [x] Double-sided portals, approached from either side
- [ ] A portal seen through another portal falls back to its plain material (fixed in `98a491ed4`,
      not retested in-game since)


src/cgame/default/Makefile.am
```diff
@@ -211,6 +211,7 @@ cgame_la_SOURCES = \
 	cg_intermission.c \
 	cg_inventory.c \
 	cg_light.c \
+	cg_portal.c \
```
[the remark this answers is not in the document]
jdolan: Fixed in `ec7206e24` -- `cg_portal.c` and `cg_portal.h` are now file references, group children, Headers and Sources entries across all four cgame targets, and are in the MSVC `.filters` files too.

src/client/renderer/Makefile.am
```diff
@@ -59,6 +60,7 @@ librenderer_la_SOURCES = \
 	r_mesh.c \
 	r_model.c \
 	r_occlude.c \
+	r_portal.c \
```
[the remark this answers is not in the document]
jdolan: Fixed in `ec7206e24` -- `r_portal.c` and `r_portal.h` are now file references, group children, and Headers/Sources entries in the `client-renderer` target, and are in `librenderer.vcxproj.filters` too.

src/cgame/common/cg_portal.c
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * @brief Places the camera of each portal's view, and adds the portal to the main view.
+ * @details The camera is the player's own, carried into the portal's target frame. Carrying it
+ * rather than pinning it to the target is what gives a portal parallax: leaning to the left of
+ * one shows more of what lies to the right of its exit, as a window does.
+ * @remarks This runs before anything is added to the scene, because the renderer repeats each
+ * addition into the views of the portals the main view holds. A portal's own brushwork is drawn
+ * by whatever entity owns it, as any other brushwork is.
+ */
+void Cg_AddPortals(void) {
```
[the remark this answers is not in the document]
jdolan: Fixed in `ec7206e24`, which landed after this review was requested -- `cg_portal.c` and `cg_portal.h` now have file references, group children, and Headers/Sources entries in all four cgame targets.

src/client/renderer/r_portal.c
```diff
@@ -0,0 +1,261 @@
...
+    },
+    .numColorTargets = 2,
+    .depthAttachment = { .format = SDL_GPU_TEXTUREFORMAT_D32_FLOAT, .clearDepth = 1.f },
+  });
+}
+
+/**
+ * @brief Draws one portal's view into its own layer of the portal framebuffer.
+ */
+static void R_DrawPortal(const r_bsp_portal_t *portal) {
```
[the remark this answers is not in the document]
jdolan: Fixed in `ec7206e24`, which landed after this review was requested -- `r_portal.c` and `r_portal.h` now have file references, group children, and Headers/Sources entries in `client-renderer`. CI `build-windows` is green on that commit too.

src/client/renderer/shaders/bsp_fs.glsl
```diff
@@ -138,13 +163,49 @@ void main(void) {
...
+
+          st += pixels / vec2(viewport.zw);
+        }
+      } else {
+        st += offset;
+      }
+    }
+
+    if (portal) {
+      fragment.diffuse_sample = vec4(texture(texture_portal, vec3(st, portal_layer)).rgb, 1.0);
```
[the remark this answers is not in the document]
jdolan: The `.dxil` blobs are deliberately not regenerated -- `shaders/Makefile.am` documents it: D3D12 is unusable pending an SDL_gpu fix, so the `shaders` target transpiles SPIR-V and MSL only, and the committed `.dxil` files are retained just because ObjectivelyGPU resolves that extension for a D3D12 device. The SPIR-V and Metal artifacts for `bsp_fs`, `bsp_fs_alpha_test` and `sprite_fs` were regenerated and are in this branch.

src/client/renderer/r_main.c
```diff
@@ -416,6 +418,7 @@ static void R_InitLocal(void) {
   r_draw_material_stages = Cvar_Add("r_draw_material_stages", "1", CVAR_DEVELOPER, "Controls the rendering of material stage effects (developer tool).");
   r_depth_pass = Cvar_Add("r_depth_pass", "1", CVAR_DEVELOPER, "Controls the rendering of the depth pass (developer tool).");
   r_occlude = Cvar_Add("r_occlude", "1", CVAR_DEVELOPER, "Controls the rendering of occlusion queries (developer tool).");
+  r_portals = Cvar_Add("r_portals", "1", CVAR_ARCHIVE, "Controls rendering the view through portal surfaces.");
```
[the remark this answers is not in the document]
jdolan: `r_portals` is the intended name and default. The cvar belongs in the renderer because cgame always offers portals and the renderer elects whether to draw them, and it defaults on because a map with portal faces looks wrong without them -- it is a performance escape hatch, not an opt-in. The PR description was stale on both counts and has been corrected.

src/client/renderer/r_portal.c
```diff
@@ -0,0 +1,261 @@
...
+
+  CommandBuffer *commands = r_context.device->commands;
+
+  r_view_t *view = portal->view;
+
+  view->framebuffer = r_portal.framebuffer;
+
+  R_UpdateFrustum(view);
+
+  R_UpdateUniforms(view);
```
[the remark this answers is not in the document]
jdolan: Oblique near-plane clipping was implemented on this branch and then removed by design. A portal targets a point entity, so the exit has no face for the near plane to be skewed onto -- there is nothing to clip against, and the destination camera sits in open space like any other. The PR description still advertised it; that was stale and has been corrected.

src/quemap/writebsp.c
```diff
@@ -614,6 +753,15 @@ int32_t EmitDrawElements(Vector *faces) {
     out->material = FaceMaterial(a);
     out->surface = a_surface & SURF_MASK_DRAW_ELEMENTS_CMP;
 
+    if (a_surface & SURF_PORTAL) {
+      if (num_portal_faces == MAX_BSP_PORTALS - 1) {
```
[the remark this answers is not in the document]
jdolan: Correct, fixed in `675edbd4c`. The check now fires at `MAX_BSP_PORTALS`, so a map may compile all 64.

src/client/renderer/r_portal.c
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  R_UpdatePortalFramebuffer();
+
+  r_view_stats_t *stats = r_stats;
+
+  for (int32_t i = 0; i < view->num_portals; i++) {
+    r_stats = &view->portals[i]->view->stats;
+    R_DrawPortal(view->portals[i]);
+  }
```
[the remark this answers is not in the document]
jdolan: Accepted as a deliberate trade-off rather than fixed. The atlas is hash-cached and portal light sources are static, so the only thing that can lag is a caster that moved this frame, and then by one frame. Hoisting the shadow pass ahead of the portals would mean hoisting the main view's `R_UpdateLights` with it, which conflicts with the portal views overwriting the same global light block -- not worth that restructuring for a shadow a few pixels behind.

src/client/renderer/r_bsp_draw.c
```diff
@@ -56,6 +57,16 @@ typedef struct {
...
 
+/**
+ * @brief Per-draw BSP fragment uniforms.
+ * @remarks The fragment stage has no use for the model matrix, and needs the portal layer
+ * that the vertex stage does not, so the two stages take different structs at the same slot.
+ */
+typedef struct {
+  r_active_dynamic_lights_t active_dynamic_lights;
+  int32_t portal_layer;
+} r_bsp_fragment_locals_t;
```
jdolan: Kind of a weird name for something that is portal-specific, no? Can't we just add portal_layer to the uniform locals above, and push that? I see active_dynamic_lights already exists there.

src/client/renderer/r_bsp_draw.c
```diff
@@ -99,13 +110,67 @@ static struct {
...
 
+/**
+ * @return The layer of the portal texture @p draw samples, or `-1` for none.
+ * @details `R_DrawPortals` clears the layer of every portal and sets it only on those it draws,
+ * so a portal that was not offered, was beyond the cap, or was culled reads `-1` here. Portal
+ * views offer no portals at all, which is what keeps portals from recursing: seen through one,
+ * a portal face falls back to its plain material.
+ */
+static int32_t R_BspPortalLayer(const r_bsp_draw_elements_t *draw) {
+  return draw->portal ? draw->portal->layer : -1;
```
jdolan: This helper is called once. Remove it and inline this instruction.

src/client/renderer/r_bsp_draw.c
```diff
@@ -99,13 +110,67 @@ static struct {
...
+  const int32_t layer = R_BspPortalLayer(draw);
+
+  if (layer != r_bsp_draw.fragment_locals.portal_layer) {
+    r_bsp_draw.fragment_locals.portal_layer = layer;
+
+    $(pass->commands, pushFragmentUniformData, SLOT_UNIFORMS_LOCALS,
+      &r_bsp_draw.fragment_locals, sizeof(r_bsp_draw.fragment_locals));
+  }
+}
+
```
jdolan: Can these be combined if we collapse fragment_locals into bsp_uniform_locals?

src/client/renderer/r_bsp_draw.c
```diff
@@ -206,6 +271,21 @@ static GraphicsPipeline *R_DrawBspMaterialStagePipeline(cm_blend_t src, cm_blend
   return out->pipeline;
 }
 
+/**
+ * @return True if @p block should be skipped when drawing @p view.
+ * @details The main view has hardware occlusion queries, which subsume frustum culling. A
+ * portal view cannot use them at all -- they were resolved for a camera somewhere else
+ * entirely -- so it culls its own frustum and nothing more.
+ */
+static bool R_CullBspBlock(const r_view_t *view, const r_bsp_block_t *block) {
```
jdolan: Probably `static inline bool` this guy

src/client/renderer/r_entity.c
```diff
@@ -92,6 +92,15 @@ r_entity_t *R_AddEntity(r_view_t *view, const r_entity_t *ent) {
   R_SetEntityBounds(e);
 
   view->num_entities++;
+
+  // the view weapon is placed relative to the camera it was added for, so it would appear
+  // adrift in the world of any other view
+  if (!(ent->effects & EF_WEAPON)) {
```
jdolan: The ordering here is kinda brittle, no? It requires that portals have been added to the view before entities. I understand that is the case, but.. Wouldn't a more robust approach be like an `R_UpdatePortalViews` or `R_UpdatePortals`? I think this would be closer to how we handle `r_light_t`, which must also calculate its list of shadow casting entities. See `R_UpdateLights / R_UpdateLightEntities`.

src/client/renderer/r_main.c
```diff
@@ -130,12 +131,12 @@ void R_UpdateUniforms(const r_view_t *view) {
...
-    // one voxel of the fallback buffers: clamping to a zero-sized grid would
-    // not, since clamp() with a low bound above its high bound is undefined,
-    // and a zero-sized box would not either, since voxel_uvw divides by it
-    if (view->type != VIEW_MAIN || !r_models.world) {
+    // a view with no relation to the world's lighting - the player model preview - must land
+    // all of its lookups on the one voxel of the fallback buffers: clamping to a zero-sized
+    // grid would not, since clamp() with a low bound above its high bound is undefined, and a
+    // zero-sized box would not either, since voxel_uvw divides by it. A portal view is a view
+    // of the world like any other, so it keeps the world's voxels
+    if ((view->type != VIEW_MAIN && view->type != VIEW_PORTAL) || !r_models.world) {
```
jdolan: Should we invert this conditional and drop the comment because it all then becomes obvious?

`if (view->type == VIEW_MAIN) ` 

Means that `r_models.world` is guaranteed to be non-null.

src/client/renderer/r_main.c
```diff
@@ -517,7 +520,9 @@ void R_Init(void) {
   R_InitDecals();
   
   R_InitSky();
-  
+
+  R_InitPortal();
```
jdolan: plural

src/client/renderer/r_main.c
```diff
@@ -544,6 +549,8 @@ void R_Shutdown(void) {
 
   R_ShutdownSky();
 
+  R_ShutdownPortal();
```
jdolan: Plural

src/client/renderer/r_material.c
```diff
@@ -188,6 +188,14 @@ static void R_ResolveMaterialStages(r_material_t *material) {
...
     r_stage_t *stage = (r_stage_t *) Mem_LinkMalloc(sizeof(r_stage_t), material);
     stage->cm = cs;
+    stage->flags = cs->flags;
+
+    // a stage naming the material's own diffusemap draws what the face would have drawn, which
+    // for a portal face is the portal. That keeps a portal's material working unchanged when
+    // portals are not being drawn: the stage simply samples the texture it names
+    if (!q_strcmp(cs->asset.name, cm->diffusemap.name)) {
+      stage->flags |= STAGE_PORTAL;
+    }
```
jdolan: Shouldn't this read more like:

```c
if (cm->surface & SURF_PORTAL) {
  if (!q_strcmp(cs->asset.name, cm->diffusemap.name)) {
    stage->flags |= STAGE_PORTAL;
  }
}
```

And with that guard, the comment is not even needed.

src/client/renderer/r_mesh_draw.c
```diff
@@ -586,7 +586,7 @@ void R_DrawMeshEntities(const r_view_t *view, RenderPass *pass) {
   // The player model preview must never bind the current world's voxel/sky
   // data: its view origin has no relation to the loaded map's lighting, so
   // doing so would produce seemingly random lighting on the preview model.
-  const r_bsp_model_t *bsp = (view->type == VIEW_MAIN && r_models.world) ? r_models.world->bsp : NULL;
+  const r_bsp_model_t *bsp = ((view->type == VIEW_MAIN || view->type == VIEW_PORTAL) && r_models.world) ? r_models.world->bsp : NULL;
```
jdolan: Similar to the inversion I recommended above, I think this just becomes:

```c
  const r_bsp_model_t *bsp = view->type == VIEW_PLAYER_MODEL ? NULL : r_models.world->bsp;
```

src/shared/matrix.h
```diff
@@ -454,6 +454,17 @@ static inline vec3_t __attribute__ ((warn_unused_result)) Mat4_Transform(const m
   );
 }
 
+/**
+ * @return The input direction transformed by the specified matrix, ignoring its translation.
+ */
+static inline vec3_t __attribute__ ((warn_unused_result)) Mat4_TransformVector(const mat4_t m, const vec3_t v) {
```
jdolan: So Mat4_RotateVector then?

jdolan (commented):
Comments.

src/client/renderer/r_bsp_draw.c
jdolan: Agreed, collapsed in `58b2932f7`. One block `{ mat4 model; active_dynamic_lights; int portal_layer; }`, declared in both `bsp_vs` and `bsp_fs`, pushed once with `pushUniformData`. `spirv-dis` on the regenerated blobs confirms both stages agree on the offsets (0 / 64 / 128), matching the C struct.

src/client/renderer/r_bsp_draw.c
jdolan: Removed in `58b2932f7`; the expression is inline in `R_PushBspPortalLayer` now.

src/client/renderer/r_bsp_draw.c
jdolan: Yes -- with the structs collapsed these become one `pushUniformData`, which pushes to both stages. Done in `58b2932f7`.

src/client/renderer/r_bsp_draw.c
jdolan: Done in `2eebc37af`.

src/client/renderer/r_entity.c
jdolan: You are right, and it was brittle. Done in `ddd08f7b9`: the fan-out is out of `R_AddEntity`, `R_AddLight`, `R_AddSprite`, `R_AddBeam` and `R_AddDecal`, replaced by a `R_UpdatePortals` that copies the finished scene into each portal view, called from `R_DrawPortals` once the scene is complete -- the `R_UpdateLights` / `R_UpdateLightEntities` shape you pointed at. Nothing depends on the ordering now.

Copying the finished structs is also cheaper than re-adding: `R_AddEntity` was recomputing the matrix, mesh tag, mesh config and bounds once per portal, all of which are view-independent.

It turned up a defect while I was in there. Decals were being repeated too, and `R_UpdateDecals` clips them into the shared, persistent `block->decals` geometry rather than into anything the view owns -- so with four portals up, an impact decal was clipped and drawn five times over, and burned the instance ring five times as fast. Portal views never called `R_DrawDecals` anyway, so they are no longer given a decal list at all.

src/client/renderer/r_main.c
jdolan: Inverted in `2eebc37af`, to `view->type == VIEW_PLAYER_MODEL` for the fallback -- that keeps VIEW_PORTAL on the world's voxels, which the old condition also did. I checked the guarantee before dropping the null check: `Cl_UpdateScene` only runs at `CL_ACTIVE`, so VIEW_MAIN and VIEW_PORTAL both imply a loaded world, and VIEW_UNKNOWN never reaches here. Kept a trimmed comment, since the clamp/divide-by-zero reason for the fallback is not obvious from the code.

src/client/renderer/r_main.c
jdolan: `R_InitPortals` / `R_ShutdownPortals` in `2eebc37af`.

src/client/renderer/r_main.c
jdolan: Same, `2eebc37af`.

src/shared/matrix.h
jdolan: `Mat4_RotateVector` in `2eebc37af`.

src/client/renderer/r_material.c
jdolan: I tried this and then backed it out -- I think the guard would break portals rather than tighten them, so flagging rather than applying.

`cm->surface` is what the `.mat` declares, but `SURF_PORTAL` reaches a face by two routes that never touch it: `quemap/map.c:436` sets it from the side being textured `common/portal`, and a per-face flag set in the editor is preserved by `map.c:415` (a material's flags are only applied when `side->surface == 0`). A face flagged either way, whose material does not itself say `surface portal`, would silently lose its portal stage and draw the named texture instead. `common/portal.jpg` ships with no material today, so anyone writing one is in exactly that position.

It also buys nothing, which is why I did not think the trade was worth it: `bsp_fs` already gates on `(material.flags & STAGE_PORTAL) && portal_layer >= 0`, and the layer is only non-negative for a draw whose elements carry a portal. A non-portal material naming its own diffusemap was already inert.

Left as it was with a comment saying why it cannot be guarded here. Happy to apply it if you would rather require materials to declare the flag -- just say so.

src/client/renderer/r_mesh_draw.c
jdolan: Applied in `2eebc37af`. Worth noting it leaves `r_mesh_draw.c` asymmetric with its neighbours: `R_DrawOpaqueBspEntities`, `R_DrawBlendBspEntities`, `R_DrawDecals`, `R_DrawSprites`, `R_DrawSky` and `R_DrawOcclusionQueries` all still guard `!r_models.world` on the same call path. By the `CL_ACTIVE` argument those guards are dead code, so either they should all go or this one should stay -- your call which way.

jdolan: Agreed, R_DrawPlayerModelView *directly calls R_DrawMeshEntities* so it needs this guard.

Whereas R_DrawOpaqueBspEntities and friends are only called from R_DrawMainView / R_DrawPortalView, which dispatch through R_DrawEntities to hit these, right?

We could replace those early returns with `assert(r_models.world)` so if I ever fuck up the call hierarchy, I'll know immediately / cleanly.

src/client/renderer/r_material.c
jdolan: Yea I think we do want to require that. It's a lot less ambiguous imo. There's zero reason at all for a mapper to use some ordinary material used elsewhere in the map as a portal face.

jdolan: Applied in `e5518b565`. Understood on the intent -- a portal face should be an explicit choice, not an accident of a stage naming its own texture.

Worth knowing the one behaviour this settles: material stages now only draw the portal when the `.mat` says `surface portal`. A face flagged `SURF_PORTAL` by being textured `common/portal` still draws the portal in the base pass, since that reads the face's own flags -- it just will not drive stages unless the material declares it too. That matches what the QUAKED docstring and the new `entities.def` / fgd entries already tell mappers to do, so the documented path is the working one.

src/client/renderer/r_mesh_draw.c
jdolan: Confirmed, and done in `7c20c0a3a`. `R_DrawPlayerModelView` calls `R_DrawMeshEntities` directly (`r_main.c:385`), never `R_DrawEntities`, so that one guard is live and stays. `R_DrawEntities` is only reached from `R_DrawMainView` and `R_DrawPortal`.

Asserted in six places: `R_DrawOpaqueBspEntities`, `R_DrawBlendBspEntities`, `R_DrawDecals`, `R_DrawSprites`, `R_DrawSky` and `R_UpdateOcclusionBounds`. They are dead twice over -- `r_models.world` is assigned in `R_LoadBspModel` and only ever cleared in `R_InitModels`/`R_ShutdownModels`, so it is never NULL again once a map has loaded, and nothing on these paths runs before that.

Left `R_DrawPost` alone: its guard is dead by the same argument, but it is called straight from `Cl_UpdateScene` rather than through this hierarchy, so it did not seem to be part of what you were pointing at. Say the word if you want it too.

Issue #1067, opened 2026-09-20
Issues with the Demo Playback mode


skies912:
### Once a demo has been loaded

- [ ] If I drag the playback slider around too fast, I drop back to console with a `CMD_MAX_STRINGS exceeded` error.
- [ ] Speaking of playback, dragging the slider around doesn't actually set the demo position until I LMB to switch back into first person. I think it's setting me to free mode and that's what breaks it.
- [ ] My player's model doesn't render in Free mode. Intended? I could still hear it playing in first person as well.
- [ ] I think the toolbar doesn't capture the mouse hits correctly so the hit falls through to the demo and toggles modes. I had trouble clicking the playback buttons etc. Mouse capture seems wonky in general.
- [ ] The game time clock doesn't reset back to current playback time until the play button is hit. Frags/deaths DO update.
- [ ] Got a "Corrupt demo file: invalid chunk size" when going to the end of the playback and pressing space.
- [ ] Nitpick but the player anims (like walk/run) don't pause when the demo pauses.

### Demos UI panel

- [ ] `Play` Button appears disabled until I click on demo AND hover over the button. Expected it to become enabled onSelectDemoItem(). It also stays enabled when demos are deleted (and maybe unselected in general?) 
- [ ] Should probably attach a protocol version to the demo itself - tried to play a demo when the feature was first introduced and Quetoo hung with no recovery - had to force kill. Other newer demos worked fine.

jdolan:
How fast are you dragging the slider? 😂 I have not seen that one, but I also didn't intentionally beat on the thing.

I think what is happening on your second point is that you are _in_ 3rd person / free flight, and because that does not draw your own client, it does not _look_ like the scrubber is working. It is.

Yes, EF_SELF is preventing the recorded client from rendering in free-fly / spectate mode. Confirmed bug, easy fix.

Corrupt demo thing is an over-eager warning that can be suppressed.

jdolan:
Thanks for the thorough list, @Skies912 — that was careful testing and it turned up real bugs. Eight of the nine are fixed and merged to main in #1073. Here is where each one landed, in plain terms.

### Fixed

**Dropping to the console when you drag the slider fast.** Every nudge of the slider was sent to the server as its own command, and the server has a flood limit to stop misbehaving clients. Drag quickly and you tripped it, so it kicked you. The slider now sends the position you have dragged to at a steady rate, rather than on every twitch of the mouse.

[...]

PR #1073, opened 2026-09-20
Fix demo playback and the demos panel (#1067)

jdolan:
Addresses eight of the nine items in #1067.

Nothing here is confirmed on screen — it is built and reasoned about, not played. The items worth
exercising are called out at the end.

## Playback

**Dragging the scrubber dropped you to the console.** `demo_seek` is a server user command, so every
one the slider emitted travelled as a `CL_CMD_STRING` and counted against `CMD_MAX_STRINGS`. The
delegate fires on every mouse motion, so a quick drag issued more than eight in one server frame and
the server kicked the client for flooding. The destination is now recorded and sent at the server's
own tick rate, and flushed at once when the drag ends.

**The recorded player was invisible in free flight.** Six sites asked whether the view was inside the
client entity by testing `!thirdPerson`, which is a different question: a detached demo camera is in
neither first nor third person. `Cg_ViewIsSelf` answers the one they meant, so the model, giblet,
beams, muzzle effects and inactive indicator all draw.

**Clicking a transport control also toggled free flight.** A mouse button becomes a key event and then
a bind, so pressing Play fired `+attack` too, which playback reads as "leave the recorded player
behind". While paused the mouse is ungrabbed and drawn as a cursor for those controls, so attack is
dropped for as long as that is true. The cost is that watch/free cannot be toggled while paused.

**"Corrupt demo file: invalid chunk size" at the end.** The keyframe index is appended after the
stream, so the position left by the terminator points at the first index entry, and unpausing read
that entry as a chunk size. The end is latched; a seek clears it. Play at the end is a no-op, because
unpausing there starved the send loop of frames and took the netchan keep-alive with it, timing the
viewer out.

**Animations kept running while paused.** They advance from `cl.unclampedTime`, which climbed whether
or not the demo did. Stopping that clock alone was tried and reverted: it also drives the movement
commands, so the free camera could not be flown. The two are separate concerns wearing one clock, so
there are now three: `cl.inputTime` for what the viewer does, `cl.time`/`cl.unclampedTime` for the
world, and `cl.worldMsec` for the per-frame delta that sprite physics and leg yaw integrate. Outside a
paused demo they are all equal, so nothing changes for anyone else. Stepping a frame also drew the
tick before the one stepped to, because a stopped clock cannot climb into a frame that arrives while
it is stopped.

## Demos panel

**Play looked disabled until you hovered it.** `updatePlay` set the disabled bit directly. The look
lives in the style and only `Control::stateDidChange` asks for it to be recomputed.

**A demo now records the protocol it was made under.** `DemoHeader` gains `protocolMajor` and
`protocolMinor`, and playback refuses a major mismatch before the map loads rather than serving a
stream the client will drop itself over. Every demo on disk today is version 2, whose header has no
such fields, so those are read from the stream instead — the first message of any recording is
`SV_CMD_SERVER_DATA`, which carries both. Version 2 demos still play, and still list: the offset a
demo's stream starts at comes from its own version.

The browser leaves out what it cannot play and says so in one line, with the detail behind
`debug cgame`. It judges the minor only when the recording names the module we are running, because
the minor is per game module and another module's demo plays under that module.

## Not addressed

- **The clock does not reset after a seek** (item 5). `TimeView` reads `CS_TIME`, and a seek jumps to a
  keyframe, which holds only a frame number and an offset. Config strings live in the stream at the
  moment they were broadcast, so the client keeps the last one it saw. Fixing it wants keyframes to
  carry a config-string snapshot.
- **The slider appearing not to seek** (item 2) was the free-flight item seen from outside, per the
  discussion on the issue.

## Worth exercising

- A demo recorded before this branch: it should still list and play.
- Pausing, then flying the free camera, then stepping a frame.
- Pausing over an explosion — the smoke should hold its pose, not keep expanding.


jdolan:
Addressed in `6082a2e3d`.

> Version 3 demo protocol fields must be read from the header during demo discovery.

Right, and for a reason worth recording. The browser was reading the protocol out of the stream whatever the version, because only the stream names the client game that recorded the demo, and the minor is per game module — judging it against the module we happen to be running would hide another module's demo that plays perfectly under its own.

But that left two readers consulting two different places for the same fact: `Sv_LoadDemo` takes a version 3 demo's protocol from the header, the browser took it from the stream. A file whose header and stream disagreed would be listed by one and refused by the other.

The header now wins where it exists, and the stream is still read for the module name. It costs no extra seek — the header read lands exactly where the stream begins.

Version 2 demos, which is every demo on disk today, still come entirely from the stream, since their header has no such fields.

